Issue #27, opened 2016-02-26
Create the stryker 'api' and define interfaces for TestFramework and TestRunner


nicojs:
We should create the Stryker 'api' in a seperate folder. In this folder we should create the new interface definitions for TestFramework and TestRunner.

We should describe API as solution 2 on the image below.
![stryker-api](https://cloud.githubusercontent.com/assets/1828233/13349593/3837d33c-dc7b-11e5-8abc-e72c4438a89c.jpg)

nicojs:
Fixed with issue #29

PR #29, opened 2016-02-26
Created api for TestFramework and TestRunner

nicojs:
Created highlevel api for TestFramework and TestRunner. Should be moved to its own git repo, but leave in here for the time being.
Solve issue #27 

@simondel could you review it


nicojs:
@simondel it now is an optional property, together with all other properties.

simondel:
I have a question about the following interface:

```
interface TestRunResult {
  specNames: string[];
  succeeded?: number;
  failed?: number;
  timedOut?: number;
  errorOccurred?: number;
  timeSpent?: number;
  coverage?: CoverageResult;
}
```

Why is are `timedOut` and `errorOccurred` numbers? Right now these two properties are of type `boolean` in the file `src/TestResult.ts`.

nicojs:
ah yes, that is indeed better, i will update the PR.

nicojs:
I choose to do it with an enum, is this better?

simondel:
LGTM!

Issue #30, opened 2016-03-02
Add a facade to run a test-runner in child process


nicojs:
We need a facade to start a test-runner in a child process. This way, every test runner can stay clear of starting child processes itself. The timeout can also be much easier implemented in this facade, because he can simply kill the child process if it takes too long (and start a new one).

This can be easily implemented using an `adapter` pattern.

Issue #25, opened 2016-02-24
Rewrite karma test runner to keep the karma server open


nicojs:
We should rewrite the karma test runner to keep the karma server open instead for the entire test run, instead of opening up a new node process for each mutation test run. This can be done using the way of working in the [karma-coverage-per-test](https://github.com/nicojs/karma-coverage-per-test/) prototype.

We can maybe also add a new abstraction for starting child processes and handling them, so we don't need to recreate it at a later stage (server side unit tests).

simondel:
See also #27

PR #31, opened 2016-03-02
Added new karma-test-runner

nicojs:
Added the new karma-test-runner which satisfies the api from #27 . 

1. Re-ordered the modular imports. The idea is that we have multiple named exports per sub-module, angular 2 style.
2. Altered the api: the run() method now returns a promise.
3. Altered the api: added additional options object for test runner instances specifics.
4. No support for timeout, see #30


simondel:
I noticed that the Karma test runner is not `silent` when the tests are ran, while it is silent when stryker is being executed.

Is this a bug or are the tests no longer supposed to be `silent`?

nicojs:
> I noticed that the Karma test runner is not silent when the tests are ran, 

That is correct. The karma test runner is not using a child process, and thus the output is visible when running unit tests.

It is not visible when Stryker is being executed, because Stryker isn't using the new test runner yet ;). For that to work, we also need the other components to work is specified by the new api, meaning the JasmineTestFramework #32 and the TestRunnerFasade #30

nicojs:
I added some more commits, because i felled like the plugin model was part of this pull request. So the karma-runner now registers itself to the TestRunnerFactory (new class)

simondel:
We'll have to add comments in the near future so our API is documented but for now we can do without it.

It looks good! Well done :)

nicojs:
solved issue #25

PR #33, opened 2016-03-04
Changed getters and setters into accessors

### LINKED ISSUE #22: Replace getters and setters with attributes.
As we moved to typescript, it would be relatively easy to move to attributes instead of getters-setters. If we need logic in the future, we'll replace the attribute with an ES 5 property.

simondel:
Fixes #22


src/AbstractSyntaxTreeNode.ts
```diff
@@ -8,41 +7,30 @@ import BaseTestRunner from './testrunners/BaseTestRunner';
  * @constructor
  */
 export default class AbstractSyntaxTreeNode {
+  private _node: ESTree.Node;
```
nicojs: I see this pattern a lot. I can agree that you want to isolate the properties (and thus not making them writable), but i would still like to see parameter properties. That would save a lot of boilerplate.

src/Mutant.ts
```diff
@@ -88,27 +125,18 @@ export default class Mutant {
...
-  getFilename() {
-    return this.filename;
-  };
-
-  /**
    * Gets if the Mutant has the status KILLED.
    * @function
    * @returns {Boolean} True if the Mutant has the status KILLED.
    */
   hasStatusKilled() {
```
nicojs: I don't think we the methods `hasStatusKilled`, `setStatusKilled`, `hasStatusUntested`, `setStatusUntested`, `hasStatusSurvived`, `setStatusSurvived`, `setStatusTimedOut` and `hasStatusTimedOut` anymore, as we now have enumerations. This would save a lot code.

simondel: I agree :) I'll fix it :+1:

src/AbstractSyntaxTreeNode.ts
simondel: Sure!

simondel:
@nicojs Your comments have been resolved :+1:

nicojs:
Thanks! Looks fine like this. All classes in `testrunners` will be deleted so you were correct to not touch them.

PR #40, opened 2016-03-05
Removed TypeUtils

### LINKED ISSUE #24: Removal of the TypeUtils
TypeUtils was designed to add runtime checks to validate the types. We don't need it anymore as we moved to compile time checks with typescript.

simondel:
Fixes #24 

Unfortunately, `myVar instanceof Object` misbehaves, just like in JavaScript because Arrays are also Objects. So when checking if a part of an Abstract Syntax Tree is an actual JS Object, I decided to also check if it is NOT an Array.


nicojs:
Looks good! We'll see what we do with the object/array checks before we release v1

Issue #32, opened 2016-03-02
Create the jasmine test framework using the new api


nicojs:
See #29 , we should create the jasmine test framework in order to pick and choose tests.

PR #41, opened 2016-03-07
Jasmine test selector

nicojs:
Added the JasmineTestSelector

* Renamed TestFramework to TestSelector as selecting tests is what it does
* Updated the api to the new format

solve issue #32


simondel:
Looks good!

Issue #43, opened 2016-03-11
New api for Mutation


nicojs:
We want a more clean the mutation api. It should be as easy as possible to create mutations.

simondel:
Fixed wiith #44

PR #44, opened 2016-03-11
Added an api for applying mutations

nicojs:
solve #43

### DIFF
--- a/src/api/mutant.ts
+++ b/src/api/mutant.ts
@@ -1,0 +1,2 @@
+
+export {default as Mutation} from './mutant/Mutation';
--- a/src/api/mutant.ts
+++ b/src/api/mutant.ts
@@ -1,2 +1,2 @@
 
-export {default as Mutation} from './mutant/Mutation';
+export {default as Mutator} from './mutant/Mutator';
--- a/src/api/mutant/Mutation.ts
+++ b/src/api/mutant/Mutation.ts
@@ -1,0 +1,13 @@
+
+interface Mutation {
+  
+  name: string;
+  
+  nodeTypes: string[];
+  
+  canMutate(node: ESTree.Node): boolean;
+  
+  applyMutations(copy: () => ESTree.Node): ESTree.Node[];
+}
+
+export default Mutation;
--- a/src/api/mutant/Mutator.ts
+++ b/src/api/mutant/Mutator.ts
@@ -1,5 +1,5 @@
 
-interface Mutation {
+interface Mutator {
   
   name: string;
   
@@ -10,4 +10,4 @@ interface Mutation {
   applyMutations(copy: () => ESTree.Node): ESTree.Node[];
 }
 
-export default Mutation;
+export default Mutator;
--- a/src/api/mutant/Mutator.ts
+++ b/src/api/mutant/Mutator.ts
@@ -3,11 +3,7 @@ interface Mutator {
   
   name: string;
   
-  nodeTypes: string[];
-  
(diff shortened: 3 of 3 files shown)


simondel:
How will we know which part of the Node has been mutated? 

A node could be quite big, for example, this code: 
```ts
if(a > b){
  //Some code
} else if (false){
  //Lots of code
} else {
  //Some more code
}
```
Is passed as a single `IfStatement` node. If the `RemoveConditionalsMutation` gets this node, it will replace `a > b` with `true` and the receiver of the *mutated* node will most likely generate the entire if/else-if/else batch as *mutated code*.

nicojs:
Hmm, you are correct. We should change it like this:
```javascript
applyMutations(node: ESTree.Node): ESTree.Node[];
```
The node passed trought to applyMutations should be frozen (readonly). We can do that easely with the following piece of code (utilizing the Object.freeze() method).
https://github.com/substack/deep-freeze/blob/master/index.js

Basically: the idea is that you walk the given node and only copy the node you want to mutate. You than mutate the copy and return the mutated node (or nodes). We than can easily fill in the node at the correct place using the location node and escodegen for the code generation.

For example:
```javascript
// Javascript:
if(i%15===0){
	console.log('FizzBuzz')
}
// AST
        {
            "loc": {
                "start": {
                    "line": 1,
                    "column": 0
                },
                "end": {
                    "line": 2,
                    "column": 1
                }
            },
            "type": "IfStatement",
            "test": {
                "loc": {
                    "start": {
                        "line": 1,
                        "column": 3
                    },
                    "end": {
                        "line": 1,
                        "column": 11
                    }
                },
                "type": "BinaryExpression",
                "operator": "===",
                "left": {
                    "loc": {
                        "start": {
                            "line": 1,
                            "column": 3
                        },
                        "end": {
                            "line": 1,
                            "column": 7
                        }
                    },
                    "type": "BinaryExpression",
                    "operator": "%",
                    "left": {
                        "loc": {
                            "start": {
                                "line": 1,
                                "column": 3
                            },
                            "end": {
                                "line": 1,
                                "column": 4
                            }
                        },
                        "type": "Identifier",
                        "name": "i"
                    },
                    "right": {
                        "loc": {
                            "start": {
                                "line": 1,
                                "column": 5
                            },
                            "end": {
                                "line": 1,
                                "column": 7
                            }
                        },
                        "type": "Literal",
                        "value": 15,
                        "raw": "15"
                    }
                },
                "right": {
                    "loc": {
                        "start": {
                            "line": 1,
                            "column": 10
                        },
                        "end": {
                            "line": 1,
                            "column": 11
                        }
                    },
                    "type": "Literal",
                    "value": 0,
                    "raw": "0"
                }
            },
            "consequent": {
                "loc": {
                    "start": {
                        "line": 1,
                        "column": 12
                    },
                    "end": {
                        "line": 2,
                        "column": 1
                    }
                },
                "type": "BlockStatement",
                "body": []
            },
            "alternate": null
        }

```

You than would only retrieve mutation: 

```javascript
 {
                        "loc": {
                            "start": {
                                "line": 1,
                                "column": 5
                            },
                            "end": {
                                "line": 1,
                                "column": 7
                            }
                        },
                        "type": "Literal",
                        "value": true,
                        "raw": "true"
                    }
```

simondel:
Looks like a good approach @nicojs !

simondel:
Shall we add the discussed change about providing a 'frozen' node to the mutation in this PR or a different one?

nicojs:
I think this one. I'll update it

PR #45, opened 2016-03-11
Mutated line fix

### LINKED ISSUE #15: Mutated code is shown incorrectly
escodegen applies formatting to the code. This usually doesn't matter, but in a few cases it does. This causes the mutated line of code to be shown incorrectly because line numbers have changed.

An example can be seen using the [escodegen online demo](https://estools.github.io/escodegen/demo/index.html) by using the following code:

```js
var arr = ['A', 'B'];
var obj = { name: 'Simon', age: 22 };
```

A *possible solution* is to ignore this feature of escodegen and search the AST for the code which should be on the specified line number. By generating this part of the AST and removing all line breaks, the correct line of code could be shown in the report.

simondel:
Fixes #15


src/Mutant.ts
```diff
@@ -89,29 +86,62 @@ export default class Mutant {
...
 
   /**
+   * Inserts the substitude into the mutatedCode based on the mutatedLocation.
+   * This also alters the originalLine and mutatedLine.
+   * @param substitude - The mutated code which will replace a part of the originalCode
+   */
+  private insertSubstitude(substitude: string) {
+    let linesOfCode = this._originalCode.split('\n');
+
+    if (this.mutatedLocation.start.line === this.mutatedLocation.end.line) {
```
nicojs: I think this if can be removed. The multi-line replacing logic will also work for single line mutations

src/mutations/RemoveConditionalsMutation.ts
```diff
@@ -15,23 +15,23 @@ export default class RemoveConditionalsMutation extends BaseMutation {
...
   applyMutation(filename: string, originalCode: string, node: ESTree.IfStatement| ESTree.DoWhileStatement| ESTree.WhileStatement| ESTree.ForStatement, ast: ESTree.Program) {
-    var originalTest = node.test;
-
     var mutants: Mutant[] = [];
-    node.test = <ESTree.Literal>{
-      type: Syntax.Literal,
-      value: false,
-      raw: 'false'
+    
+    var location: ESTree.SourceLocation = {
```
nicojs: I think location can be reused, no need to copy it here i think.

nicojs:
I merged the PR. I think we should later on move the insertSubstitution code to some other class, because now every mutant will have a copy of the entire source file (with its substitution) in-memory. Only the mutation itself i actually needed. We can insert this mutation when we are writing the mutated file to disk and thus saving us from having large strings in the mutant class.

PR #46, opened 2016-03-13
Test runner orchestrator

nicojs:
Added the test runner orchestrator, responsible for managing test runners. This still needs to change because i just added 3 properties to Mutant because i needed them:
```typescript
  public timeSpentScopedTests: number;
  public scopedTestIds: number[];
  public specsRan: string[];
```
We will need to fill these properties or provide other ways of reaching them. That should not be the responsibility of the orchestrator, that's why i didn't include it in this PR


src/TestRunnerOrchestrator.ts
```diff
@@ -0,0 +1,146 @@
...
+        runnerAdapter: IsolatedTestRunnerAdapterFactory.create(this.createTestRunSettings(tempSourceFiles, selector, this.options.port + portOffset, false)),
+        selector
+      };
+    });
+  }
+
+  private copyAllSourceFilesToTempFolder() {
+    return new Promise<FileMap>((resolve, reject) => {
+      let fileMap: FileMap = Object.create(null);
+      var tempFolder = StrykerTempFolder.createRandomFolder('test-runner-source-files');
```
simondel: Who removes this folder and when?

nicojs: Well... i think at the end we can simply remove the entire temp folder. For Milestone 1 (and the beta) i think we do not need to worry about it. The temp folder is created in your temp directory, people are pretty used to cleaning the temp directory from time to time. I see karma is also not cleaning the temp directory.. so maybe don't worry about it? :)

nicojs: Also: I still have my doubts about the `StrykerTempFolder` being in the api folder and about the name. We should discuss this before V1

nicojs:
I just added a commit for the removal of the temp mutation api. It was still in there even though we discussed to not implement it in v1. Sorry :)

src/TestRunnerOrchestrator.ts
simondel: I agree that we should discuss this before V1 :)

simondel:
It looks good, well done!

PR #53, opened 2016-03-18
Integrate new test runner and test selector api's in stryker

### LINKED ISSUE #34: Integrate new test runner and test framework api's in stryker
Integrate the new `TestRunner` and `TestFramework` apis in Stryker. Make sure to support the options we already had, maybe rename them.

nicojs:
Resolve #34 

There is still work to be done, but i hope we can now start releasing a 0.1 release (beta)


simondel:
There are still two issues, as far as I know:
* No specnames are printed if a mutant survives
* Tests are ran even when the mutant is not covered

simondel:
We'll fix the coverage in a later version.

PR #58, opened 2016-03-19
Fixed too many tests being ran

simondel:
The correct tests are now matched with every Mutant. The MutantRunResultMatcher now looks for the smallest statement that covers a mutant. If this statement does NOT cover the mutant: then the mutant is not covered by this test.

There is an issue with Karma, it reports that a run has caused an error if every test failed. This caused mutants which were covered and killed by one test to be reported as 'untested'. I have solved this issue by checking if no test failed when determining if a run caused an error.

### DIFF
--- a/src/MutantRunResultMatcher.ts
+++ b/src/MutantRunResultMatcher.ts
@@ -7,12 +7,16 @@ export default class MutantRunResultMatcher {
 
   matchWithMutants() {
     this.mutants.forEach(mutant => {
+      let smallestStatement: string;
       this.runResultsByTestId.forEach((testResult, id) => {
         let covered = false;
         if (testResult.coverage) {
           let coveredFile = testResult.coverage[mutant.filename];
           if (coveredFile) {
-            covered = this.fileCoversMutant(mutant, coveredFile);
+            if (!smallestStatement) {
+              smallestStatement = this.findSmallestCoveringStatement(mutant, coveredFile)
+            }
+            covered = coveredFile.s[smallestStatement] > 0;
           }
         } else {
           // If there is no coverage result we have to assume the source code is covered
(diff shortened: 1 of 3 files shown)


simondel:
This fixes #58

src/MutantRunResultMatcher.ts
```diff
@@ -1,5 +1,6 @@
 import Mutant from './Mutant';
 import {RunResult, Location} from './api/test_runner';
+import {CoverageResult} from './api/test_runner/Coverage';
```
simondel: @nicojs Can I do this in a cleaner way?

nicojs: Yes, we should pass it through the `test_runner.ts` file. That way, we can later move around the files within the test_runner folder as long as we don't break the api (the api being the `api/test_runner`). The reason it is not yet in the file is probably laziness The import would look like this:

```typescript
import {CoverageResult} from './api/test_runner';
```

src/MutantRunResultMatcher.ts
```diff
@@ -26,12 +26,26 @@ export default class MutantRunResultMatcher {
...
 
-  private mutantCovers(mutant: Mutant, location: Location): boolean {
+  private mutantCoversFile(mutant: Mutant, coveredFile: CoverageResult): boolean {
+    let smallestStatementFound = "";
+    
+    Object.keys(coveredFile.statementMap).forEach(statementId => {
+      let location = coveredFile.statementMap[parseInt(statementId)];
+        
+      if(this.mutantCoversLocation(mutant, location)){
+        smallestStatementFound = statementId;
```
nicojs: This implementation is actually very much dependent on the order in which the statement map is build. I think this is too error prone. What if Istanbul decides to change the order? I think it would be better to compare the location of the statements that cover the mutant.

if nr of lines of the location is smaller OR (the nr of lines are the same AND the number of columns is smaller)... something like that

src/MutantRunResultMatcher.ts
```diff
@@ -12,8 +13,7 @@ export default class MutantRunResultMatcher {
         if (testResult.coverage) {
           let coveredFile = testResult.coverage[mutant.filename];
           if (coveredFile) {
-            covered = Object.keys(coveredFile.statementMap).some(statementId => coveredFile.s[statementId] !== 0
-              && this.mutantCovers(mutant, coveredFile.statementMap[parseInt(statementId)]))
+            covered = this.mutantCoversFile(mutant, coveredFile);
```
nicojs: I think we can do this in 2 steps. Find out the id of the 'smallest' statement which covers the mutant *once*. After that start checking the test results (the `s` collection of each coverage result). The id of the 'smallest' statement shouldn't change when looping over the testresults... or... do you see a scenario in which they do change?

src/MutantRunResultMatcher.ts
simondel: I'll fix this in case we ever come across a coverage reporter which doesn't report coverage as a tree

simondel:
@nicojs Could you re-review this PR?

nicojs:
This already looks al lot better! Can you also add a unit test which tests the different scenarios for 'smallest statement'? In particular:

* Overlapping statements differing in lines
* Overlapping statements differing in columns

PR #61, opened 2016-03-22
Fix package publish

### LINKED ISSUE #59: NPM install: compile TS files
The Typescript files should be compiled to JavaScript files during a NPM install

nicojs:
Resolve #59


nicojs:
I will add the publish task to Travis tomorrow. Looks good. It will automatically release when we push a tag. We'll see how that goes.

simondel:
Looks good so far! Why do we have to include the typings again for the integration test? Can't we just use the typings that Stryker already has?

nicojs:
> Why do we have to include the typings again for the integration test? 

Well... the `tsconfig.json` in `test/integration/install-module/module` cannot have a reference to the files in the folder above. Hover, i did remove the typings, now made sure the typings are installed at npm-install-time.

PR #63, opened 2016-03-25
All Coverage 'map types' are now indexed with a string

### LINKED ISSUE #60: Coverage collections types should not be indexed with a number
All `map` types in `test_runner/Coverage.ts` are typed using a number indexer. For example:

```javascript
export interface StatementMap {
  [ref: number]: Location;
}
```

Although in practice, the ref is always a number, we should change it to be string. This because the usage is somewhat strange:

```javascript
coveredFile.statementMap[parseInt(statementId)]
```

With `parseInt` we convert the string to a number, but for the indexer `[]` it is converted back to a string. This is kind of pointless.

simondel:
Fixes #60

### DIFF
--- a/src/MutantRunResultMatcher.ts
+++ b/src/MutantRunResultMatcher.ts
@@ -41,9 +41,9 @@ export default class MutantRunResultMatcher {
     let smallestStatement: string;
 
     Object.keys(coveredFile.statementMap).forEach(statementId => {
-      let location = coveredFile.statementMap[parseInt(statementId)];
+      let location = coveredFile.statementMap[statementId];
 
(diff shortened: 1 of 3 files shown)


nicojs:
Looks good! Should we do the same for `BranchMap` and `FunctionMap` in this branch?

simondel:
Those two already have strings as their index. What I could do is check if any numbers are still used as indexes in the code. 

I could also re-order the interfaces of Coverage.ts so they have a natural order:
```
CoverageCollection
CoverageResult
CoverageData
BrancheCoverageData
CoverageData
CoverageData
FunctionMap
BranchMap
StatementMap
Location
Position
```

That way they are ordered in the way they contain each other. For example: CoverageCollection contains CoverageResult, which contains CoverageData, BranchCoverageData and so forth.

nicojs:
Ok. The check is done by typescript i think. So no need to do that. Reordering is OK for me

Issue #64, opened 2016-03-25
(bug) No notification when initial test run fails for some reason


nicojs:
When Karma errors in the initial test run it is treated as if there are no tests. When you turn off the `silent` flag you can see the following output:

```
INFO: Running initial test run
using config {"browsers":["PhantomJS"],"frameworks":["jasmine"],"autoWatch":false,"singleRun":false,"plugins":["karma-jasmine","karma-phantomjs-launcher","karma-coverage"],"files":[...],"port":1234,"coverageReporter":{"type":"in-memory"},"preprocessors":{...},"reporters":["coverage"]}
24 03 2016 22:14:08.982:INFO [karma]: Karma v0.13.22 server started at http://localhost:1234/
24 03 2016 22:14:08.986:INFO [launcher]: Starting browser PhantomJS
24 03 2016 22:14:09.494:INFO [PhantomJS 1.9.8 (Windows 8 0.0.0)]: Connected on socket /#tOMWTOX3l_ZzzTYYAAAA with id 67783735
karma run done with  1
```
I've tested this using our internal KarmaDesigner project with the following arguments:

```javascript
	"program": "${workspaceRoot}/node_modules/stryker/dist/src/Stryker.js",
			"stopOnEntry": false,
			"args": ["-s","app/scripts/app.js,app/scripts/httpConfig.js,app/scripts/resources/ConditionRsc.js,app/scripts/resources/PageDtoRsc.js,app/scripts/resources/SurveyResultRsc.js,app/scripts/resources/SurveyRsc.js,app/scripts/resources/SurveySummariesPageRsc.js,app/scripts/routes.js,app/scripts/services/dragDropHelper.js,app/scripts/services/idGenerator.js,app/scripts/services/jsonHelper.js,app/scripts/services/localStorage.js,app/scripts/services/modalHelper.js,app/scripts/services/notificationStorage.js,app/scripts/services/random.js,app/scripts/services/surveyFind.js,app/scripts/services/surveyForEach.js,app/scripts/services/surveyStorage.js,app/scripts/services/tenantHelper.js","-o","test/spec/controllers/editFactRuleSet.test.js"],
```

For clarification: i deliberately not included the angular js sources in the `otherFiles` list. I would expect Stryker to show me this (`cannot call module on undefined` or something)

simondel:
The funny thing is, if you keep it running it will actually log the error due to the `listenToBrowserError` function:
```
  ReferenceError: Can't find variable: angular
[...]

PR #67, opened 2016-03-30
(fix) added a notification for when a test failes in initial run.

nicojs:
* Altered the test run api: added an errorMessages array to collect the error message
* Altered the karma test runner to make sure it defaults to Complete
* Altered Stryker to console log the new errorMessages

Resolve issue #64

New output when running with a test project using angular:

```bash
INFO: Running initial test run
ERROR: test iteration 0 resulted in an error
ERROR: ReferenceError: Can't find variable: angular
at http://localhost:1234/base/app/scripts/app.js?495cda078cd1e18758ce5603c54f15af926b77fb:9
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
             "type": "node",
             "request": "launch",
             "program": "${workspaceRoot}/node_modules/grunt-cli/bin/grunt",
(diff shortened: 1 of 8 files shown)


simondel:
Nice! Now we can finally see what's going wrong :)

Issue #66, opened 2016-03-27
Unable to choose the order of files to load


nicojs:
Right now: we always load the `source` files before `other` files. This approach will not work when you need to load library files.

For example: you need to load angular, after that your own source files, after that your test files. Karma solves this by having one big list of files. Thus you have fine grained control of the order of files.

I'm thinking of introducing the same principle for stryker. Instead of source files and other files, you'll have `files` and `source` files. The `files` list also includes the source files and gives you fine grained control of the order of files. The `source files` are the files we are going to mutate.

If the `source` files contains a file which is not included in the `files` list, we'll either have to give an error, or load the files before or after the 'normal' files.

@simondel what do you think?

nicojs:
I maybe got a better idea. What about we just accept a number of file (patterns). Like this:

```bash
node stryker.js a/b.js :c/d.js g/**/*.js :f/**/*.js test/**/*.js
```
(the file globbing is not yet supported, but will be in the future)

Only the files starting with a colon (`:`) will be targeted for mutations. This way, we do not have to distinguish between `-o` and `-s` anymore and do you have both fine grained control and a simple interface. The idea is that we have to choose a character which is forbidden in windows and linux file systems, but also not used for globbing (for a later feature). 

https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words 

https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

simondel:
I think it would be best two have two parameters:
`files`: The full list of source, test and library files necessary to run the project. This includes the source files you want to mutate.
`filesToMutate`: The source files which should be mutated.

Support for file globbing is of course always nice :) I think that introducing special characters to decide if something should be mutated or not only makes the system more confusing. A system like 'here are all my files, could you please mutate this part?' would be easier to use in my opinion.

nicojs:
We could also do a hybrid solution. I'll sum up the options we have here:

1. Use *nameless parameters* for both files and filesToMutate: `node stryker.js a/b.js >c/d.js g/**/*.js >f/**/*.js test/**/*.js` (in this example, files to mutate start with a '>')
2. Use *named parameters* for both files and filesToMutate: `node stryker.js --files a/b.js c/d.js,g/**/*.js,f/**/*.js,test/**/*.js --filesToMutate c/d.js,f/**/*.js`
[...]

PR #69, opened 2016-03-30
Feature choose order of files

nicojs:
Added a feature to choose the order of files. The stryker API is not 

```bash
Stryker -f <files> -m <filesToMutate> [other options]
```
You can use file globbing in the files and filesToMutate.

resolve issue #66


simondel:
Nice work! I've updated the README to include the new arguments. If you agree with that change, the PR can be merged :)

nicojs:
I changed it a little bit for clarification. Thanks for the reminder!

PR #81, opened 2016-04-07
fix(InputFileResolver) Fix for potentially wrong order of files

nicojs:
Wrong order when globbing promisses resolve in different order.

### DIFF
--- a/src/InputFileResolver.ts
+++ b/src/InputFileResolver.ts
@@ -1,5 +1,6 @@
 import {InputFile} from './api/core';
 import {glob, normalize} from './utils/fileUtils';
+import * as _ from 'lodash';
 
 export default class InputFileResolver {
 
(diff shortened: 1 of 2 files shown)


simondel:
Why did the original code change the order and how does this code fix it?

src/InputFileResolver.ts
```diff
@@ -36,12 +37,12 @@ export default class InputFileResolver {
...
 
-  private static resolveFileGlobs(sourceExpressions: string[], resultFiles: string[]): Promise<void[]> {
-    return Promise.all(sourceExpressions.map((mutateFileExpression: string) => glob(mutateFileExpression).then(files => {
+  private static resolveFileGlobs(sourceExpressions: string[]): Promise<string[]> {
+    return Promise.all(sourceExpressions.map(expression => glob(expression).then(files => {
       if (files.length === 0) {
-        this.reportEmptyGlobbingExpression(mutateFileExpression);
+        this.reportEmptyGlobbingExpression(expression);
       }
-      files.forEach(f => resultFiles.push(f));
```
nicojs: This was the old way of pusing files to the result array. Per `expression` a globbing promise get's resolved. The promisses get created in the correct order, but the resolving can happen in any order. The pushing on the `resultFiles` array will in a random order.

src/InputFileResolver.ts
```diff
@@ -36,12 +37,12 @@ export default class InputFileResolver {
...
-    })));
+      return files;
+    }))).then(files => _.flatten(files));
```
nicojs: This is the new way of returning an array. Per `expression` a globbing promise get's resolved. The promises get created in the correct order, but the resolving can still happen in any order. However the `Promise.all` api will handle this for us and give us a `files: string[][]` in the correct order. All we have to do is flatten it before returning.

It seems to work for a test project i used.

nicojs:
See my remarks in the file changed tab.

nicojs:
Was my response clear? Can this PR be closed?

simondel:
Yes, thanks for the response!

PR #89, opened 2016-04-16
Feature read config

### LINKED ISSUE #86: Read config from a config file
I would like to add a way to read config using a config file instead of command line arguments. Very similar to what karma uses.

_Note: the current command line arguments would NOT be replaced with a config file_

Proposal:

* When you call `stryker` you can supply a config file with `--configFile` (or `-c`?). ~~If you don't supply one it will use `stryker.conf.js` as a default.~~ 
* the config file should look like this:

```javascript
    config.set({
      files: [
           'lib/**/*.js',
           'src/**/*.js',
           'test/**/*.js'
       ],
       mutate: [
            'src/**/*.js'
       ],
    
        // additional optional options with default values: 
         testFramework: 'jasmine',
         testRunner: 'karma',
         timeoutMs: 2000,
         timeoutFactor: 1.5,
         plugins: ['stryker-*'],
         port: 9234
    });
};
```
* By default: any plugin is loaded from the node_modules folder starting with the name `stryker-`.
* A plugin will also have the possibility to register a ConfigWriter to the `ConfigWriterFactory`.
    * Config writers will be applied _after_ the reading of the normal `stryker.conf.js` file. Indeed it has to, because that file defines the plugins to be loaded in the first place.
    * For example: a plugin with the name `stryker-karma-runner` could register an additional `ConfigWriter` with the name `'karma'` to the `ConfigWriterFactory`. That `KarmaConfigWriter` can look for a property with the name `karmaConfigFile` in the existing config. It can read that `karma.conf.js` file and than choose to override some of the settings in the `stryker.conf.js` file. For example override the `files` based on the `karma.conf.js`'s `files` key.
    * This way, plugin implementers can have maximum flexibility in adding config based on config settings that live outside the stryker process.

Does everybody agree with this proposal?

* **Note**: I just changed some parts. This will not replace the command line arguments, just add a new one. I would like to maybe add a `stryker-cli` in the future where you would only accept a config file.*

nicojs:
Adding the config file feature
resolve #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@
 *Professor X: For someone who hates mutants... you certainly keep some strange company.*  
 *William Stryker: Oh, they serve their purpose... as long as they can be controlled.*
 
-
 ## Work in progress
 This repository is a work in progress. We only support Jasmine tests in the browser for now. Please create and vote on [issues](https://github.com/stryker-mutator/stryker/issues) to help us determine the priority on features.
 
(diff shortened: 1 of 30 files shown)


README.md
```diff
@@ -27,25 +27,54 @@ node node_modules/stryker/dist/src/Stryker.js --help
 ```
 
 ## Configuration
+Options can be configured either via the command line or via a config file.
```
simondel: This is still a bit unclear. The documentation below suggests that `files` and `mutate` are required, while they are required either using the *old* options or using a config file.

simondel:
This works with [restangular](https://github.com/mgonto/restangular) and [angular-form-for](https://github.com/bvaughn/angular-form-for/).

If the README is changed, this PR can be merged. Nice work! :+1:

nicojs:
Thanks :v:
I fixed the readme and will merge now

PR #94, opened 2016-04-21
Refactor logging

nicojs:
* Replaced all console.log statements with log4js log statements
* Add some debug logging statements
* Pipe through the test runner output to `log.trace` method.

### DIFF
--- a/src/Stryker.ts
+++ b/src/Stryker.ts
@@ -41,7 +41,7 @@ export default class Stryker {
     this.loadPlugins();
     this.applyConfigWriters();
     this.setGlobalLogLevel(); // loglevel could be changed
-    freezeRecursively(this.config);
+    this.freezeConfig();
     var reporterFactory = new ReporterFactory();
     this.reporter = reporterFactory.getReporter('console');
   }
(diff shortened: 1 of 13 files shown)


src/Stryker.ts
```diff
@@ -102,6 +104,13 @@ export default class Stryker {
     });
   }
 
+  private freezeConfig() {
+    freezeRecursively(this.config);
+    if (log.isDebugEnabled()) {
```
simondel: Is this check added for performance reasons as you don't want to spend time stringifying the config the user does not even see the log message?

nicojs: Yes correct. Its a minor thing but sometimes it might be usefull
Op 22 apr. 2016 08:55 schreef "Simon de Lang" <notifications@github.com>:

In src/Stryker.ts
<https://github.com/stryker-mutator/stryker/pull/94#discussion_r60697760>:

> @@ -102,6 +104,13 @@ export default class Stryker {
>      });
>    }
>
> +  private freezeConfig() {
> +    freezeRecursively(this.config);
> +    if (log.isDebugEnabled()) {

Is this check added for performance reasons as you don't want to spend time
stringifying the config the user does not even see the log message?

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/stryker-mutator/stryker/pull/94/files/c6d9d2b7e42d56de0419cc0b75d55379b38047a5#r60697760>

simondel:
Looks good!

PR #99, opened 2016-05-08
Implemented Mutator API

simondel:
I have finished implementing the Mutator API.

All `Mutation` classes are now `Mutator` classes.
The `Mutator` class is now called `MutatorOrchestrator`.

A Mutator now has to return a list of `ESTree.Node` objects, each of which has an ID (see: *src/api/mutant/StrykerNode.ts*). The `MutatorOrchestrator` now creates the `Mutants` instead of handing this task to the `Mutators`.

Mutators can return a **childnode** of the `Node` they received. This is already done in the `RemoveConditionalsMutator`.

I have also added a Mutator(see: *src/mutators/BlockStatementMutator.ts*) from grunt-mutation-testing which removes the entire content of a block-statement. This includes function definitions.


customTyping/escodegen.d.ts
```diff
@@ -0,0 +1,145 @@
+
+declare module 'escodegen' {
```
nicojs: Can you also make a PR for this in the DefinitelyTyped repo? That way we can remove this in the future.

src/MutatorOrchestrator.ts
```diff
@@ -28,24 +29,22 @@ export default class Mutator {
...
+    let mutants: Mutant[] = [];
 
-    _.forEach(sourceFiles, (sourceFile: string) => {
+    sourceFiles.forEach((sourceFile: string) => {
       try {
-        var fileContent = fileUtils.readFile(sourceFile);
-        var abstractSyntaxTree = parserUtils.parse(fileContent);
-        var nodes = parserUtils.getNodesWithType(abstractSyntaxTree, types);
-        var newMutants = this.findMutants(sourceFile, fileContent, abstractSyntaxTree, nodes);
+        let fileContent = fileUtils.readFile(sourceFile);
```
nicojs: Can't we use an async file io call? Using blocking io can be time consuming when we have hunderds of source files. On the other hand, we can also wait for it to become a problem.

src/MutatorOrchestrator.ts
```diff
@@ -28,24 +29,22 @@ export default class Mutator {
    * @returns {Mutant[]} The generated Mutants.
    */
   mutate(sourceFiles: string[]) {
```
nicojs: we should maybe rename this to `generateMutants`. I think that name more accurately describes what the method is doing.

src/MutatorFactory.ts
```diff
@@ -0,0 +1,24 @@
+import {Mutator} from './api/mutant';
```
nicojs: Cool :+1:

We should move this file in the api folder, that way we can later easily seperate it from the stryker main repo into the stryker-api repo.

src/utils/parserUtils.ts
```diff
@@ -34,30 +33,41 @@ export function parse(code: string): any {
...
 /**
- * Finds all nodes which have one of several types in a syntax tree.
+ * Finds all nodes which a type in a syntax tree.
  * @function
  * @param abstractSyntaxTree - The current part of the abstract syntax tree which will be investigated.
- * @param  types - The list of types which are requested.
- * @returns  All nodes which have one of the requested types.
+ * @returns  All nodes with a type.
  */
-export function getNodesWithType(abstractSyntaxTree: any, types: string[], nodes?: any[], key?: string): any[] {
```
nicojs: This method name is not descriptive. What is it doing? Is the key-parameter ever used? Maybe it should be called `collectNodes` or something, or `collectAndFreezeNodes` (although i'm not a fan of `and` in method names). What do you think?

src/MutatorOrchestrator.ts
```diff
@@ -27,25 +27,23 @@ export default class Mutator {
...
+        let nodes = parserUtils.getNodesWithType(abstractSyntaxTree);
+        let newMutants = this.findMutants(sourceFile, fileContent, abstractSyntaxTree, nodes);
         mutants = mutants.concat(newMutants);
       } catch (err) {
         switch (err.code) {
           case 'ENOENT':
             log.info(`Skipping file ${err.path} because it does not exist`);
             break;
           default:
+            console.log(err);
```
nicojs: We should use log.error for this (using a log4js logger)

src/MutatorOrchestrator.ts
```diff
@@ -64,13 +62,21 @@ export default class Mutator {
...
-        _.forEach(this.mutations, (mutation: BaseMutation) => {
-          if (mutation.canMutate(astnode)) {
-            mutants = mutants.concat(mutation.applyMutation(sourceFile, originalCode, astnode, ast));
+        Object.freeze(astnode);
+        this.mutators.forEach((mutator: Mutator) => {
+          try {
+            mutator.applyMutations(astnode, _.cloneDeep).forEach((mutatedNode: ESTree.Node) => {
+              let mutatedCode = parserUtils.generate(mutatedNode);
+              let originalNode = nodes[(<StrykerNode>mutatedNode).nodeID];
+              mutants.push(new Mutant(mutator, sourceFile, originalCode, mutatedCode, originalNode.loc));
```
nicojs: Maybe add a debug log line here?
`generated 3 mutants on using source line 23 using the RemoveCondition mutator` for example?

src/Stryker.ts
```diff
@@ -92,6 +93,17 @@ export default class Stryker {
     });
   }
 
+  private loadDefaultPlugins() {
```
nicojs: Maybe we should not call this 'default plugins'. Maybe we could better call it `loadDefaultMutators` and add it as a method to the `MutatorOrchestrator`. I don't see a single mutator as being a plugin. So also don't use the `PluginLoader`. That way, a mutators is no longer responsible by itself to add it to the `MutatorFactory` (that should generaly only be done by a plugin)

src/api/mutant/Mutator.ts
```diff
@@ -14,9 +14,10 @@ interface Mutator {
    * If the mutator wants to mutate the node, it should return a clone of the node with mutations,
    * otherwise null. 
    * @param node A FROZEN Node which could be cloned and mutated.
+   * @param deepCopy A function to create a deep copy of an object.
    * @returns An array of mutated Nodes.
    */
-  applyMutations(node: ESTree.Node): ESTree.Node[];
+  applyMutations(node: ESTree.Node, deepCopy: (obj: any) => any): ESTree.Node[];
```
nicojs: As an aside: I think a shallow copy is enough. You generally only want to alter the current node anyway. The children can still point to the frozen nodes. Maybe we can add a copy function with a `boolean` parameter (`deep`).

src/api/mutant/StrykerNode.ts
```diff
@@ -0,0 +1,5 @@
+interface StrykerNode extends ESTree.Node {
+  nodeID: number;
```
nicojs: Ok, i might be nitpicking here, but i think the casing should be `nodeId` in this case. `Id` is short for `Identifier`. Abbreviations should always be written in PascalCase (in contrast to 2 letter acronyms). But i might be alone in this ;) so you can leave it if you want.

src/mutators/BlockStatementMutator.ts
```diff
@@ -0,0 +1,30 @@
+import {Syntax} from 'esprima';
+import {Mutator, MutatorFactory} from '../api/mutant';
+
+/**
+ * Represents a mutator which can remove the content of a BlockStatement.
+ */
+export default class BlockStatementMutator implements Mutator {
+  name = 'BlockStatement';
+  types = [Syntax.BlockStatement];
```
nicojs: by default, attributes are public. I think it should be private in this case

src/mutators/BlockStatementMutator.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    return nodes;
+  }
+
+  private canMutate(node: ESTree.Node) {
+    return !!(node && this.types.indexOf(node.type) >= 0);
+  };
+}
+
+MutatorFactory.instance().register('BlockStatement', BlockStatementMutator);
```
nicojs: Like stated before: I think a mutator should not behave as a plugin. The adding to the `MutatorFactory` might be a job for the MutatorOrchestrator?

simondel: Would you like this to be the case for all *default* Mutators? (So all supported Mutators as of right now)

nicojs: Yes i would. I think we can later support a black list to disable certain mutators.

customTyping/escodegen.d.ts
simondel: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/9326 :+1:

nicojs: Cool! :+1: Usually they are pretty good with accepting new type definitions

src/api/mutant/StrykerNode.ts
simondel: Well I honestly don't care if we use ID or Id, but I think we should be consistent. Shall I change it to `Id`?

src/utils/parserUtils.ts
simondel: I think `collectNodes` is a good name. The comments can specify that nodes are also frozen. The `key` parameter was still left over from an earlier implementation of nodes, nice find!

src/Stryker.ts
simondel: This method has become obsolete due to ca6cfd15603bc151619d4b51526f070adb773dc4

src/utils/parserUtils.ts
nicojs: Maybe `collectFrozenNodes` is an even better name? :)

src/api/mutant/StrykerNode.ts
nicojs: We'll leave it like it is, `nodeID`

simondel:
@nicojs I believe that all feedback has now been implemented :+1:

PR #100, opened 2016-05-10
feat(reporter-api) Add reporter api

nicojs:
* Move Location and Position from test_runner api to core api
* Add reporter api interfaces
* Add BaseReporter helper class so not all methods need to be implemented
* Add source files to the api so an html report can be created

When this PR is accepted i want to start working on the html reporter
@paysdoc @simondel can you two take a look?


src/api/reporter/BaseReporter.ts
```diff
@@ -0,0 +1,22 @@
+import {StrykerOptions} from '../core';
+import SourceFile from './SourceFile';
+import MutantResult from './MutantResult';
+import Reporter from './Reporter';
+
+export default class BaseReporter implements Reporter {
+
+  onConfigReadingComplete(options: StrykerOptions): void {
```
simondel: What do you think of the name `onConfigRead`?

src/api/reporter/MutantResult.ts
```diff
@@ -0,0 +1,13 @@
+import MutantStatus from './MutantStatus';
+import {Location} from '../core';
+
+interface MutantResult {
+  sourceFilePath: string;
+  mutationType: string;
```
simondel: Is this the value of `Mutator.name`? If so, I think we should rename it to `mutatorName`

simondel:
It looks good! I have two questions (apart from my comments):

1. Why is there still a `BaseReporter` class?
2. How do you plan to implement the `MutantResult` interface with the existing `Mutant` class?

src/api/reporter/BaseReporter.ts
nicojs: A lot better :+1:

nicojs:
> Why is there still a BaseReporter class?

If you want to implement just one method in your `Reporter`, you still need to implement all methods. At least: so i thought, i just figured it out:

```typescript
interface Reporter {

  onConfigReadingComplete?(config: StrykerOptions): void;
  // appearently we can use a questionmark ? 
}
```

> How do you plan to implement the MutantResult interface with the existing Mutant class?

Maybe add a result property to keep the result? The idea would be to freeze the result so reporters can't alter it.

nicojs:
I implemented the review comments.

PR #103, opened 2016-05-27
refactor(RunnerOptions) Make the port number optional

simondel:
Mocha doesn't use a port but the runner options force you to specify a port so I made it optional.

### DIFF
--- a/src/api/test_runner/RunnerOptions.ts
+++ b/src/api/test_runner/RunnerOptions.ts
@@ -12,7 +12,7 @@ interface RunnerOptions{
   /**
    * Represents a free port which the test runner can choose to use
    */
-  port: number;
+  port?: number;
    
   /**
    * The underlying strykerOptions


nicojs:
I understand that a port number doesn't make sense in this case, but on the other hand: the TestRunnerOrchestrator doesn't know which test runner it is using. And thus it always needs to specify the port.

Why do you need to make it optional? To make it easier to inject `RunnerOptions` in unit tests?

simondel:
Yeah, I wanted to make injection during testing easier/cleaner for the Mocha test runner. I totally forgot about the fact that the TestRunnerOrchestrator creates a port without knowing if the test runner actually wants a port. With that in mind, I think the best option it to keep it as a mandatory property.

This PR can be closed in my opinion.

nicojs:
Yeah, for testability, you can choose 2 solutions:

1. Use a factory method to create the `RunnerOptions` with a default dummy (and unused) port
2. You can use an explicit `any` typed variable.

PR #104, opened 2016-05-30
Split ConsoleReporter into 2 reporters, implementing the new Reporter api

nicojs:
* Split `ConsoleReporter` into 2 reporters: `ClearTextReporter` and `ProgressReporter`
* Let them implement the new `Reporter` interface
* Created a `BroadcastReporter`, responsible for forwarding report events to configured reporter
* Created a `ReporterOrchestrator` which is responsible for grouping all configured reporters into 1 `BroadcastReporter` 
* Implemented all 4 report events


src/Mutant.ts
```diff
@@ -68,93 +31,43 @@ export default class Mutant {
...
-    this.insertSubstitude(substitude);
+  constructor(public mutatorName: string, public fileName: string, private originalCode: string, public replacement: string, public location: ESTree.SourceLocation) {
+    this.applyMutation();
   }
 
   /**
-   * Inserts the substitude into the mutatedCode based on the mutatedLocation.
-   * This also alters the originalLine and mutatedLine.
+   * Inserts the replacement into the mutatedCode based on the specified location.
    * @param substitude - The mutated code which will replace a part of the originalCode
```
simondel: This parameter no long exists so this line can be removed

src/Mutant.ts
```diff
@@ -68,93 +31,43 @@ export default class Mutant {
...
     
-    for (let lineNum = this.mutatedLocation.start.line - 1; lineNum < this.mutatedLocation.end.line; lineNum++) {
-      this._originalLine += linesOfCode[lineNum];
-      if (lineNum < this.mutatedLocation.end.line - 1) {
-        this._originalLine += '\n';
+    this.originalLines = '';
+    for (let lineNum = this.location.start.line - 1; lineNum < this.location.end.line; lineNum++) {
+      this.originalLines += linesOfCode[lineNum];
+      if (lineNum < this.location.end.line - 1) {
+        this.originalLines += '\n';
```
simondel: You could also build an array of lines and join them using `'\n'`

src/Mutant.ts
```diff
@@ -68,93 +31,43 @@ export default class Mutant {
...
    * @param originalCode - The original content of the file which has not been mutated.
-   * @param substitude - The mutated code which will replace a part of the originalCode.
-   * @param mutatedLocation - The part of the originalCode which has been mutated.
+   * @param replacement - The mutated code which will replace a part of the originalCode.
+   * @param location - The location of the code to be replaced
    */
-  constructor(private _mutator: Mutator, private _filename: string, private originalCode: string, substitude: string, private mutatedLocation: ESTree.SourceLocation) {
-    this.status = MutantStatus.UNTESTED;
-    this.insertSubstitude(substitude);
+  constructor(public mutatorName: string, public fileName: string, private originalCode: string, public replacement: string, public location: ESTree.SourceLocation) {
```
simondel: I believe that we use `filename` in other parts of the code instead of `fileName`

src/MutatorOrchestrator.ts
```diff
@@ -6,26 +6,29 @@ import RemoveConditionalsMutator from './mutators/RemoveConditionalsMutator';
...
 export default class MutatorOrchestrator {
-  private mutators: Mutator[];
+  private mutators: Mutator[] = [];
+  private sourceFiles: SourceFile[];
 
-  public constructor() {
+  /**
+   * @param reporter - The reporter to report read input files to
+   */
+  public constructor(private reporter: Reporter) {
```
simondel: What if there are multiple reporters?

src/Mutant.ts
nicojs: Well.. we could.. but i don't feel like splitting and concatenating all the time.

We also could add the *index based range* to the esprima parsing. See http://esprima.org/demo/parse.html and check *index based range*. It will add a `range` next to the current `loc`:

```json
"range": [
        35,
        54
    ],
```

This will be much easier for finding and replacing code. I think the performance impact will be low. Maybe we could add a story for that? That way we can remove most of the `split('\n\)` from our code.

src/ReporterOrchestrator.ts
```diff
@@ -0,0 +1,48 @@
...
+const log = log4js.getLogger('ReporterOrchestrator');
+
+
+export default class ReporterOrchestrator {
+
+  constructor(private options: StrykerOptions) {
+    this.registerDefaultReporters();
+  }
+
+  public createSingleReporter(): Reporter {
```
simondel: This method does not create a single reporter, it creates a BroadcastReporter with 'n' repoters

nicojs: Yes so in a sense it does create a single reporter. Maybe we should call it `createCombinedReporter()`?

src/MutatorOrchestrator.ts
nicojs: Indeed we would always provide a `BroadcastReporter`. The idea is that the `MutatorOrchestrator` shouldn't care about what kind of reporter we want, as long as it has the methods implemented that it expects to be there.

I've seen this way of working also in Karma, i thought it was a rather elegant solution. What do you think?

src/Mutant.ts
nicojs: Indeed, i will remove it

src/Mutant.ts
nicojs: https://en.wikipedia.org/wiki/Filename

I guess we can choose. Ok, i'll rename all to `filename`

src/Mutant.ts
simondel: Using ranges would make it easier, but you would still have to spit the string to get the data for the `originalLines` and `mutatedLines` properties.

src/MutatorOrchestrator.ts
simondel: It is an elegant solution but this constructor confused me at first because you told me we were using the `BroadcastReporter` and my mind just went like: *hold on.. is this right?*. But it's correct :) :+1:

src/ReporterOrchestrator.ts
simondel: Why not call it by what it actually does? `createBroadcastReporter()`

src/Stryker.ts
```diff
@@ -42,30 +41,27 @@ export default class Stryker {
...
-    var reporterFactory = new ReporterFactory();
-    this.reporter = reporterFactory.getReporter('console');
   }
 
   /**
    * Runs mutation testing. This may take a while.
    * @function
    */
-  runMutationTest(): Promise<void> {
+  runMutationTest(): Promise<MutantResult[]> {
```
simondel: Why are we returning a `MutantResult[]`, just because we can and because the calling process may want to do something with it?

src/api/config/Config.ts
```diff
@@ -15,6 +15,7 @@ export default class Config implements StrykerOptions {
   timeoutFactor = 1.5;
   plugins: string[] = ['stryker-*'];
   port = 9234;
+  reporter = ['progress', 'clear-text'];
```
simondel: When someone specifies some other reporter, will this array be overwritten or will the other reporter be added to this array?

src/Mutant.ts
nicojs: Yes thats true, but for this function it would be OK. Agreed to leave it for now and fix it in another issue?

src/ReporterOrchestrator.ts
nicojs: Yeah.. but than i would expect it to return a `BroadcastReporter` and that one actually doesn't implement the methods statically as far as typescript is concerned. But it would be the best name, agreed. I will change it.

src/Stryker.ts
nicojs: Exactly :smiley: 
As a consumer, you might want to do something with it...

src/api/config/Config.ts
nicojs: Overridden. This is the way Karma works as well.. You want to give the people the opportunity to use something else entirely

simondel: In that case I think we should add a final logging statement to indicate that Stryker has completely finished, but we can do this in a different PR

PR #105, opened 2016-06-02
Remove dist folder

nicojs:
Because it is hard to reference files relative to source files we decided to remove the use of the dist folder

* Update grunt process
* Alter tsc config
* Moved all js and other resources needed for testing to testResources


simondel:
It looks good, great job!
Could you change the **package.json** because it still references the `dist` folder in the `sample` script.

nicojs:
Good catch! Why update grunt-stryker? Doesn't it just `require` stryker? That should still work. I was very happy with our integration test. It really helps to make sure the `require('stryker');` still works

simondel:
Ah, you're right. I thought we couldn't simply `require('stryker')` but we can, so grunt-stryker doesn't have to be updated :)

Feel free to merge this PR.

nicojs:
Thanks! :)

PR #106, opened 2016-06-04
Add range api

nicojs:


src/api/core/Range.ts
```diff
@@ -0,0 +1,7 @@
+
+/**
+ * Represents a location in a file by range [fromInclusive, toExclusive]
+ */ 
+type Range = [number, number];
```
simondel: What's the lowest possible number? `0` or `1`?

nicojs: That would be `0`

simondel:
Looks good to me! Nice job :+1:

PR #109, opened 2016-06-16
Copy test files to the temp directory

simondel:
Mocha requires Stryker to copy not only the source files, but also the test files.

This PR ensures that Stryker copies all files to the temp folder and that a directory structure is created which matches the original directory structure.

### DIFF
--- a/src/TestRunnerOrchestrator.ts
+++ b/src/TestRunnerOrchestrator.ts
@@ -182,7 +182,7 @@ export default class TestRunnerOrchestrator {
       let copyPromises: Promise<any>[] = this.files.map(file => {
         let relativePath = file.path.substr(cwd.length);
         let folderName = StrykerTempFolder.ensureFolderExists(tempFolder + path.dirname(relativePath));
-        let targetFile = folderName + path.sep + path.basename(relativePath);
+        let targetFile = path.join(folderName, path.basename(relativePath));
         fileMap[file.path] = targetFile;
         return StrykerTempFolder.copyFile(file.path, targetFile);
       });
--- a/src/TestRunnerOrchestrator.ts
+++ b/src/TestRunnerOrchestrator.ts
@@ -23,7 +23,7 @@ interface TestRunnerSandbox {
   index: number;
   runnerAdapter: IsolatedTestRunnerAdapter;
   selector: TestSelector;
-  sourceFileMap: FileMap;
+  fileMap: FileMap;
 }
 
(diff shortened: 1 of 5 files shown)


src/TestRunnerOrchestrator.ts
```diff
@@ -158,38 +158,33 @@ export default class TestRunnerOrchestrator {
...
-        if (file.shouldMutate) {
-          let targetFile = tempFolder + path.sep + path.basename(file.path);
-          fileMap[file.path] = targetFile;
-          return StrykerTempFolder.copyFile(file.path, targetFile);
-        } else {
-          return Promise.resolve();
-        }
+        let relativePath = file.path.substr(cwd.length);
+        let folderName = StrykerTempFolder.ensureFolderExists(tempFolder + path.dirname(relativePath));
+        let targetFile = folderName + path.sep + path.basename(relativePath);
```
nicojs: You can use `path.join` to join the paths

test/unit/TestRunnerOrchestratorSpec.ts
```diff
@@ -17,7 +17,12 @@ let expect = chai.expect;
 describe('TestRunnerOrchestrator', () => {
   let sut: TestRunnerOrchestrator;
   let sandbox: sinon.SinonSandbox;
-  let files = [{ path: 'a.js', shouldMutate: true }, { path: 'b.js', shouldMutate: true }, { path: 'aSpec.js', shouldMutate: false }, { path: 'bSpec.js', shouldMutate: false }];
+  let files = [
+    { path: process.cwd() + path.sep + 'a.js', shouldMutate: true },
+    { path: process.cwd() + path.sep + 'b.js', shouldMutate: true },
+    { path: process.cwd() + path.sep + 'aSpec.js', shouldMutate: false },
+    { path: process.cwd() + path.sep + 'bSpec.js', shouldMutate: false }
```
nicojs: Maybe we should add a file which is in a deeper folder. For example `process.cwd() + path.sep + 'folder1' + path.sep + 'folder2' + path.sep + 'bSpec.js'`

nicojs:
Nice job! I have 2 small comments

test/unit/TestRunnerOrchestratorSpec.ts
nicojs: Ignore this comment. As it is a unit test testing an other unit, it is not necessary

PR #110, opened 2016-06-25
Use stryker api

nicojs:
Make use of the new stryker api as an external dependency.

* Replaced all import statements from `import ... from '../api/..'` to `import ... from 'stryker-api/...'`
* Refactored stryker `runMutationTests` to allow for a chain of `then` statements (instead of a 'pyramid hell'). Now, any errors we forget to catch should result in a rejection
* Updated `JasmineTestSelector` to implement the new `TestSelector` interface
* Updated `TestRunnerOrchestrator` to accommodate for the new `TestSelector` interface
* Updated `BroadcastReporter` to implement the new `wrapUp` method
* Updated `Stryker` to call the wrapUp method at the right time.


simondel:
Nice work, it looks good!

PR #111, opened 2016-06-29
Remove karma runner

nicojs:
* Move karma-runner from stryker repo to stryker-karma-runner
* Make the install-module integration test depend on that module
* Fix issues in PluginLoader:
	* Wrong plugin directory
	* Install only the basename of a module directory


simondel:
It looks good! Can we also remove the sampleProject? It expects the Karma runner to be installed.

nicojs:
Hmm true.. We can also add `stryker-karma-runner` as dev-dependency. That way we can still run the sample project. We can also add multiple sample projects per test runner and run them as integration test in the future.

@simondel  what do you think?

simondel:
I think a dev-dependency is the way to go for now , nice suggestion :)

nicojs:
The travis build is erroring, but i'll merge it as it does not seem to have to do with my changes

PR #112, opened 2016-07-04
Optional test selector

nicojs:
Make test selector optional. If no test selector is used, just run every test for every mutation. This feature will help us to integrate the mocha test runner, without having to also implement and support the mocha test selector.


nicojs:
Merged master branch into this branch.
Added an option to pass `'null'` (as string), has the same result as passing `null` (as in `nill`)

simondel:
Looks good! Nice work!

PR #117, opened 2016-07-11
Add bithound config

simondel:
I added a bithound config to make bithound use tslint instead of jslint.

**NOTE:** I may have to edit the config with some more commits.


nicojs:
I see that bithound is currently using jslint:

https://www.bithound.io/github/stryker-mutator/stryker

PR #116, opened 2016-07-11
 feat(event-recorder): Add event recorder reporter

nicojs:
Add an event recorder reporter which can record events by writing there json representation to files, numbering them chronologically.

PR #121, opened 2016-07-12
fix(StrykerTempFolder): Use local tmp folder

nicojs:
When running mocha tests, we need to be able to resolve the local node_modules. The way this works in npm is that it searches for a node_modules in de current working directory. If not found, check the dir above. Etc. By creating the tmp-folder's locally, node_,modules will be resolved as expected.

### DIFF
--- a/src/Stryker.ts
+++ b/src/Stryker.ts
@@ -17,6 +17,7 @@ import InputFileResolver from './InputFileResolver';
 import ConfigReader from './ConfigReader';
 import PluginLoader from './PluginLoader';
 import {freezeRecursively, isPromise} from './utils/objectUtils';
+import StrykerTempFolder from './utils/StrykerTempFolder';
 import * as log4js from 'log4js';
 
 const log = log4js.getLogger('Stryker');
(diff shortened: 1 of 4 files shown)


simondel:
When will this folder be removed form the file system? We can assume that a temp folder will (eventually) be cleared, but this is not the case for other locations such as your project folder.

nicojs:
Hmm that's true. Haven't thought about the cleaning :recycle:

What's your suggestion? I think we can just clear the entire StrykerTempFolder after stryker is done. But we could also make each individual part responsible for cleaning its own stuff.

simondel:
I would use a 'fire and forget' method for saving files. I don't believe that classes such as the TestRunnerOrchestrator should care if they make a mess in the StrykerTempFolder. I would let Stryker remove the temp folder before it exits.

One disadvantage of the StrykerTempFolder is that it will exist if users kill Stryker, but by making Stryker just delete the entire folder after a run, a cleanup will still happen.

nicojs:
Agreed. I'll implement that in this branch.

You always run the risk of having temp-files remain if you kill the process, that's totally acceptable. 

One additional disadvantage is that you cannot run multiple instances of Stryker at one time anymore, because the one that is the first to close will clean all files. But i don't think its a big issue.

nicojs:
@simondel we should first merge [pull request 116](https://github.com/stryker-mutator/stryker/pull/116/files) into master, that one contains fileUtils to easily clean a folder recursively.

nicojs:
I just added the cleaning of the folder :recycle: !

PR #123, opened 2016-07-12
refactor(cli): Add stryker as binary

nicojs:
Refactor Stryker class to not include the cli. Also add stryker as bin so it can be used directly from an npm script (i.e. "stryker": "stryker --configFile stryker.conf.js")

### DIFF
--- a/bin/stryker
+++ b/bin/stryker
@@ -1,0 +1,8 @@
+#!/usr/bin/env node
+
+'use strict';
+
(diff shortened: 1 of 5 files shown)


simondel:
This makes Stryker.ts a lot cleaner :tada:

simondel:
Odd.. My comment on this PR has disappeared and it seems to have not been merged while the master branch says it has been merged

PR #124, opened 2016-07-14
feat(ClearTextReporter) Log every mutant

simondel:
As discussed on Gitter

output

### DIFF
--- a/src/reporters/ClearTextReporter.ts
+++ b/src/reporters/ClearTextReporter.ts
@@ -1,6 +1,9 @@
 import {Reporter, MutantResult, MutantStatus, SourceFile} from 'stryker-api/report';
 import * as chalk from 'chalk';
 import * as _ from 'lodash';
+import * as log4js from 'log4js';
+
+const log = log4js.getLogger('ClearTextReporter');
 
 export default class ClearTextReporter implements Reporter {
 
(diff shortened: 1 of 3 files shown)


src/reporters/ClearTextReporter.ts
```diff
@@ -55,6 +55,38 @@ export default class ClearTextReporter implements Reporter {
...
+    this.writeLine(chalk.green('+   ' + result.mutatedLines));
+    this.writeLine('\n');
+    this.writeLine('Tests ran: ');
+    _.forEach(result.testsRan, (spec: string) => {
+      this.writeLine('    ' + spec);
+    });
+    this.writeLine('\n');
+  }
+
+  private logMutantResultToDebug(result: MutantResult): void {
```
nicojs: This method looks a lot like logMutantResult. Can't we combine the 2? As a second argument we could give the log function for example

src/reporters/ClearTextReporter.ts
```diff
@@ -1,6 +1,9 @@
 import {Reporter, MutantResult, MutantStatus, SourceFile} from 'stryker-api/report';
 import * as chalk from 'chalk';
 import * as _ from 'lodash';
+import * as log4js from 'log4js';
```
nicojs: By logging mutant results only in the `ClearTextReporter`, we will never see the mutants logged when this reporter is not active. Do you think we should maybe specify that somewhere?

src/reporters/ClearTextReporter.ts
```diff
@@ -12,32 +15,34 @@ export default class ClearTextReporter implements Reporter {
...
-    var mutantsTimedOut = 0;
-    var mutantsUntested = 0;
+    let mutantsKilled = 0;
+    let mutantsTimedOut = 0;
+    let mutantsUntested = 0;
+
+    // use these fn's in order to preserve the 'this` pointer
+    let logDebugFn = (input: string) => log.debug(input); 
+    let writeLineFn = (input: string) => this.writeLine(input);
+
```
nicojs: You need to `route` the input through these stupid helper methods, so that the `this` pointer does not get corrupted.

src/reporters/ClearTextReporter.ts
simondel: Will people ever turn this reporter off? I consider it a mandatory reporter.

src/reporters/ClearTextReporter.ts
simondel: Nice! That solves the issues I was having

PR #125, opened 2016-07-15
feat(test-runner): Support lifecycle events

nicojs:
Add support for test runner lifecycle events: init and dispose. These are called once in every test runner life cycle (if implemented).  If they return a promise, we will wait for those promises to be resolved before proceding. We will only wait for max 2 seconds with the dispose before forcing the child process to be killed.


simondel:
Yay Karma is working again :dancer: :tada: :cake:  Nice work :)

PR #127, opened 2016-07-19
test(stryker): Eat own dogfood

nicojs:
It has been a long time comming, but we can now finally run stryker on stryker itself. :dog:. Add stryker-mocha-runner and stryker-html-reporter as dev dependencies. Add the stryker.conf.js used to run stryker.


simondel:
Nice work!!!

PR #130, opened 2016-07-21
fix(TestRunnerOrchestrator): Initial test run

nicojs:
Initialize test runner also in initial test run.


nicojs:
I forgot to call init in initial test run and then forgot make a release of the `stryker-karma-runner` and update the version in stryker. Thats why this bug only came to light now that we're making preparations for the0.4 release

simondel:
0.4 HERE WE COME #hype

Issue #143, opened 2016-08-15
Problems with regular expressions in karma.webpack configuration


knegusen:
First, I really like this project!

Second, i don't know if this should be posted here or in stryker-karma-runner.

Third, my issue is almost the same as this one, [karma-webpack issue](https://github.com/webpack/karma-webpack/issues/42)

The thing is that I can use it when I don't use babel-loader in my webpack configuration but when i use babel-loader I get this message:

```
[2016-08-15 17:01:40.857] [TRACE] IsolatedTestRunnerAdapter - webpack: bundle is now INVALID.

[2016-08-15 17:01:40.891] [TRACE] IsolatedTestRunnerAdapter - 15 08 2016 17:01:40.886:ERROR [karma]: [Error: [object Object] is not a valid test]
Error: [object Object] is not a valid test
    at asMatcher (/Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/node_modules/webpack-core/lib/LoadersList.js:48:9)
    at LoadersList.matchPart (/Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/node_modules/webpack-core/lib/LoadersList.js:65:16)
    at LoadersList.matchObject (/Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/node_modules/webpack-core/lib/LoadersList.js:91:12)
    at LoadersList.<anonymous> (/Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/node_modules/webpack-core/lib/LoadersList.js:78:12)
    at Array.map (native)
    at LoadersList.match (/Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/node_modules/webpack-core/lib/LoadersList.js:71:19)
    at /Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/lib/NormalModuleFactory.js:111:68
    at /Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/node_modules/async/lib/async.js:726:13
    at /Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/node_modules/async/lib/async.js:52:16
    at done (/Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/node_modules/async/lib/async.js:246:17)
    at /Users/axelsson/ProjectsWebstorm/stryker-test/node_modules/webpack/node_modules/async/lib/async.js:44:16
[...]

PR #146, opened 2016-09-01
fix(isolated-test-runner): Support regexes

nicojs:
* Add support for communicating regexes (and functions) from the stryker process to test runners running in child processes.
* Add dependency to 'serialize-javascript'. A minimalistic module for serializing regexes and functions in a json structure.


nicojs:
Should fix issue #143

Issue #151, opened 2016-09-14
False positive mutant in Angular controller


bkimminich:
I have this code in an Angular 1.x controller:

```
    $scope.showDetail = function (id) {
      $uibModal.open({
        templateUrl: 'views/ProductDetail.html',
        controller: 'ProductDetailsController',
        size: 'lg',
        resolve: {
          id: function () {
            return id
          }
        }
      })
    }
```

My original test simply did `expect($uibModal.open).toHaveBeenCalled` which lets this mutant survive: `$scope.showDetail = function (id) {}`.

Performing this mutation manually yields a failing test:

```
	Expected spy open to have been called with [ { templateUrl : 'views/ProductDetail.html', controller : 'ProductDetailsController', size : <jasmine.any(function String() {
	    [native code]
	})>, resolve : { id : <jasmine.any(function Function() {
	    [native code]
	})> } } ] but it was never called.
```

I even extended my test to be more specific about the expected modal dialog: `expect($uibModal.open).toHaveBeenCalledWith({ templateUrl: 'views/ProductDetail.html', controller: 'ProductDetailsController', size: jasmine.any(String), resolve: { id: jasmine.any(Function) } })`

The mutant `$scope.showDetail = function (id) {}` still survives. _How is that possible?_

nicojs:
Hmm this indeed should not happen. Can you confirm that your test did ran for that mutant? Stryker performs analysis on code coverage. A bug there might result in this test not being ran for the mutant.

To see this, please run stryker with the clear test reporter. You can also explicitly disable code coverage analysis by setting `testSelector` to `null`.

bkimminich:
Running `clear-text` shows 15 surviving mutants, but the mentioned one is not among them. So I guess it is one of the `11 mutants untested.`?

```
240 mutants tested.
11 mutants untested.
0 mutants timed out.
225 mutants killed.
Mutation score based on covered code: 93.75%
Mutation score based on all code: 89.64%
```

bkimminich:
Using `testSelector: null` makes the tests go slower and seems to kill 2 extra mutants, but not this one.

simondel:
Could you change your loglevel to debug (`logLevel: 'debug'`)? That will show you the tests that ran for every mutant except for just the killed ones.

bkimminich:
It's among the untested ones:
```
[2016-09-15 15:04:27.114] [DEBUG] ClearTextReporter - Mutant untested!
[...]

PR #155, opened 2016-10-01
fix(MutantRunResultMatcher): False positive fix

nicojs:
When a mutant alters something smaller than an exact statement, it was reported as `untested`, well in fact it was covered.

* Fixed the algorithm
* Add a test to prevent regression

Fix for #151


nicojs:
I just did the test on @bkimminich's juice-shop and with test selector gives the same result as without. Also, the original reported false positive now also gets killed. A thing of absolute beauty :+1:

PR #156, opened 2016-10-02
Fix(TestRunnerOrchestrator) Mention the initial test run

simondel:
Added a logging statement to indicate that Stryker is busy with the initial test run


nicojs:
This is good enough for now. In a later version i would like to add more information about code coverage analysis.

Issue #82, opened 2016-04-07
bug: Stryker continues when there are no mutants


simondel:
If Stryker was unable to mutate the source code, the rest of the mutation testing process still continues (starting multiple test runners, logging the mutation score etc), while we can simply quit if there are no mutants.

PR #161, opened 2016-10-07
fix(stryker): Stop running if there are no mutants

nicojs:
* Refactor stryker to now stop running where there are no mutants
* Make the main `runMutationTest()` method much clearer to understand by moving all details into seperate methods
* Make use of (parameter context matching)[http://es6-features.org/#ParameterContextMatching]

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,7 +19,7 @@
       "env": {
         "NODE_ENV": "development"
       },
-      "externalConsole": false,
+      "console": "internalConsole",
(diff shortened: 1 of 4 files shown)


nicojs:
fix #82

nicojs:
The bithound is still far to strict. The failing files include stuff that we cannot do anything about, like `'58% of commits in this file are bugfixes'`. What shall we do about it?

The warnings about the broken js file we should be able to ignore.

PR #165, opened 2016-10-16
feat(one-pass-coverage): Support one-pass coverage measurement

nicojs:
* Refactor `TestRunnerOrchestrator` into 2 classes: `Sandbox` and `SandboxCoordinator`
* Refactor message protocol to make use of TS 2 [type narrowing](https://www.typescriptlang.org/docs/handbook/advanced-types.html)
* Refactor reporters to support new `MutantResult`s.
* Refactor config module to support new CoverageAnalysis property
* Add coverage instrumentation
* Add class CoverageInstrumenterStream responsible for instrumenting a single file
* Add class CoverageInstrumenter responsible for managing instrumentation of all files.
* Update Stryker, Sandbox and SandboxCoordinator to use the new CoverageInstrumenter
* Reimplement the MutantTestMatcher so it works with new coverageAnalysis
* Add error handling for when TestFramework is missing and coverageAnalysis is `perTest`.
* Add a default coverage reporting implementation in the IsolatedTestRunnerAdapterWortker which whould work for most test runners.


nicojs:
I removed the 'wip' label. I think its ready to be merged. I still need to test with a new mocha test runner version, but it should work

nicojs:
Works fine with the new mocha runner implementation: [stryker-mocha-runner#4]](https://github.com/stryker-mutator/stryker-mocha-runner/pull/4)

Here are my findings:

## Coverage analysis 'off'

```
1074 mutants tested.
0 mutants not covered.
9 mutants timed out.
592 mutants killed.
Mutation score based on covered code: 55.96%
Mutation score based on all code: 55.96%
[2016-11-12 16:32:04.253] [INFO] Stryker - Done in 2 minutes 43 seconds.
```

## Coverage analysis 'all'

```
844 mutants tested.
230 mutants not covered.
9 mutants timed out.
592 mutants killed.
Mutation score based on covered code: 71.21%
Mutation score based on all code: 55.96%
[2016-11-12 16:35:03.617] [INFO] Stryker - Done in 2 minutes 3 seconds.
```

## Coverage analysis 'perTest'

```
844 mutants tested.
230 mutants not covered.
9 mutants timed out.
592 mutants killed.
Mutation score based on covered code: 71.21%
Mutation score based on all code: 55.96%
[2016-11-12 16:38:02.907] [INFO] Stryker - Done in 2 minutes 3 seconds.
```

It's great to see that the mutation score is exactly the same with all codeAnalysis strategies. Also: the mutation score based on covered code is the same.

The time difference for coverageAnalysis `'all'` and `'perTest'` is the same, so it's not what i'd hoped for. I think it will save time with slow test runs (i.e. ui-bootstrap), but not for Stryker. Maybe we need to think of other ways to speed things up.

An other problem i faced was the fact that the progress reporter isn't really helpful to show progress. See #180

src/coverage/CoverageInstrumenter.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  public hooksForTestRun(): string {
+    if (this.coverageAnalysis === 'perTest') {
+      log.debug(`Adding test hooks file for coverageAnalysis "perTest"`);
+      return wrapInClosure(`
+          var id = 0, coverageStateAtStart;
+          window.__coverage__ = globalCoverage = {};
+
+          ${this.testFramework.beforeEach(beforeEachFragmentPerTest)}
```
simondel: What happens if we don't have a testFramework?

src/isolated-runner/IsolatedTestRunnerAdapterWorker.ts
```diff
@@ -1,11 +1,8 @@
...
-import RunMessageBody from './RunMessageBody';
-import ResultMessageBody from './ResultMessageBody';
+import { AdapterMessage, RunMessage, StartMessage, ResultMessage, EmptyWorkerMessage, WorkerMessage } from './MessageProtocol';
+import { RunnerOptions, TestRunner, RunStatus, TestRunnerFactory, RunResult } from 'stryker-api/test_runner';
 import PluginLoader from '../PluginLoader';
 import * as log4js from 'log4js';
-import {isPromise, deserialize} from '../utils/objectUtils';
+import { isPromise, deserialize } from '../utils/objectUtils';
 
 const log = log4js.getLogger('TestRunnerChildProcessAdapterWorker');
```
simondel: Why does this class have a different name than its file?

src/ConfigReader.ts
```diff
@@ -57,4 +60,16 @@ export default class ConfigReader {
     }
     return configModule;
   }
+
+  private validate(options: StrykerOptions) {
+    if (!_.find(VALID_COVERAGE_ANALYSIS_VALUES, v => v === options.coverageAnalysis)) {
```
nicojs: I'll fix this

src/TestFrameworkOrchestrator.ts
```diff
@@ -0,0 +1,48 @@
+import { TestFrameworkFactory, TestFramework } from 'stryker-api/test_framework';
+import { StrykerOptions } from 'stryker-api/core';
+import * as log4js from 'log4js';
+
+const IGNORE_WARNING = 'Set `coverageAnalysis` option explicitly to "off" to ignore this warning.';
```
nicojs: I'll remove the variable. Was used multiple times before, but the new setup is simpler 👍

src/TestFrameworkOrchestrator.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export default class TestFrameworkOrchestrator {
+
+  constructor(private options: StrykerOptions) {
+  }
+
+  determineTestFramework(): TestFramework {
+    if (this.options.coverageAnalysis !== 'perTest') {
+      log.debug('The `coverageAnalysis` setting is "%s", not hooking into the test framework to achieve performance benefits.', this.options.coverageAnalysis);
+      return null;
```
nicojs: I'll make it `TestFramework | null`. Void, or undefined, represents a deeper absence of a value and should never be used for a functional absence of a value in my opinion. By making it `TestFramework | null` we can (at a later time please) enable [`--strictNullChecks`](https://www.typescriptlang.org/docs/release-notes/typescript-2.0.html) and let the compiler help us to not make mistakes.

src/coverage/CoverageInstrumenter.ts
nicojs: At this point, we should always have a TestFramework. When `coverageAnalysis: 'perTest'`, stryker enforces there to be a test framework. If not, it fails fast with a message telling you how to fix it (either choose a test framework, or set coverageAnalysis to 'all'). 

I wanted to not complicate things.

src/isolated-runner/IsolatedTestRunnerAdapterWorker.ts
nicojs: fixed it

Issue #163, opened 2016-10-13
Make amount of parallel test runners configurable


nicojs:
See https://github.com/bkimminich/juice-shop/issues/219

I think we should allow values from 1 to 32

simondel:
This can be closed right?

nicojs:
Fixed with #176

PR #176, opened 2016-11-10
Allow setting config for cpuCount

jeznag:
Was finding that I get `file table overflow` if I use too many CPUs #177


simondel:
Fixes #163

src/TestRunnerOrchestrator.ts
```diff
@@ -168,7 +168,7 @@ export default class TestRunnerOrchestrator {
   }
 
   private createTestRunnerSandboxes(): Promise<TestRunnerSandbox[]> {
-    let cpuCount = os.cpus().length;
+    let cpuCount = this.options.cpuCount || os.cpus().length;
```
simondel: We use the cpuCount to determine how many test runners should be used in parallel. I think it's a good idea to name the option after this purpose. Something like `testRunnerCount`. What do you think?

src/TestRunnerOrchestrator.ts
```diff
@@ -168,7 +168,7 @@ export default class TestRunnerOrchestrator {
...
     let testRunnerSandboxes: TestRunnerSandbox[] = [];
     let allPromises: Promise<any>[] = [];
     log.info(`Creating ${cpuCount} test runners (based on cpu count)`);
```
simondel: We may want to change this logging statement as it's not always based on cpu count. We can either have two logging statements *(based on cpu count)* and *(based on config)* or just have one logging statement and not mention where we get the amount from. What do you think?

CC @nicojs

simondel:
Thanks for taking the time to look through our code and make this change! Glad to hear that it fixes your issue :)

src/TestRunnerOrchestrator.ts
nicojs: I actually would like `maxConcurrency` or `concurrency` better. That way, Stryker can still choose not to spin up more test runners than it needs. For example, why spin up 8 test runners for 6 tests? 

The default would be `infinite`. That way you let Stryker decide what's best.

@jeznag

src/TestRunnerOrchestrator.ts
nicojs: I would suggest just leave out the part about cpu/config. We already log the config on debug.

nicojs:
Thanks for the PR! We will still need to add this to the `stryker-api` before we can accept it.

src/TestRunnerOrchestrator.ts
simondel: I do prefer mentioning the test runner. That way you know what it's about. How does `maxConcurrentTestRunners` sound?

nicojs: @simondel  Agreed, sounds good.

@jeznag do you want to create a PR to add it to the [stryker-api](https://github.com/stryker-mutator/stryker-api)? Or shall we?

jeznag:
I'll try to work on it tomorrow.

jeznag:
@nicojs @simondel do you have a guide for contributors? I don't really know how to get tests and linting etc. set up locally. Probably need to learn type script too!

simondel:
Thanks for your work so far! We have a [Contributing.MD](https://github.com/stryker-mutator/stryker/blob/master/CONTRIBUTING.md) which covers some of your questions. If you are missing anything, feel free to let us know (or create a PR :P).

I would suggest updating to the latest master commit. We recently merged a major change which renamed the TestRunnerOrchestrator to the SandboxCoordinator. This should also fix the bithound checks. 

The test for the SandBoxCoordinator, which tests if Stryker uses the CPU count can be found [here](https://github.com/stryker-mutator/stryker/blob/master/test/unit/SandboxCoordinatorSpec.ts#L76). You could use that as a guide to writing a test to test this new functionality. Are you familiar with the way describe's and it's work?

nicojs:
Version 0.4 of stryker-api released earlier this day with the `maxConcurrentTestRunners` setting in there.

@jeznag the way is clear for you to proceed with this PR :)

jeznag:
@simondel @nicojs have had another go with the rebased version. Took me quite a while cos I was fighting with a sinon issue (https://github.com/sinonjs/sinon/issues/1196).

Noticed a few things:
1. Getting a linting error
image

2. There's no yarn.lock file - should we add one

src/SandboxCoordinator.ts
```diff
@@ -60,12 +60,13 @@ export default class SandboxCoordinator {
   }
 
   private createSandboxes(): Promise<Sandbox[]> {
-    const cpuCount = os.cpus().length;
+    const numConcurrentRunners = this.options.maxConcurrentTestRunners || os.cpus().length;
```
nicojs: The default value for `maxConcurrentTestRunners` is [`Infinity`](https://github.com/stryker-mutator/stryker-api/blob/master/src/config/Config.ts#L20). I think this is also why the travis build is failing. 

I think a better implementation could be:

```typescript
 let numConcurrentRunners = os.cpus().length;
let source = 'CPU count';
if(numConcurrentRunners > this.options.maxConcurrentTestRunners && this.option.maxConcurrentTestRunners > 0){
   numConcurrentRunners = this.options.maxConcurrentTestRunners;
   source = 'maxConcurrentTestRunners';
}
```

or something like that

jeznag: Ok done and added tests for the additional cases.

Let me know if it's all good and I'll squash the commits.

test/unit/SandboxCoordinatorSpec.ts
```diff
@@ -79,10 +79,30 @@ describe('SandboxCoordinator', () => {
...
       const mutants: any[] = [mockMutant(0, []), mockMutant(1, [])];
 
       sut = new SandboxCoordinator({
+        maxConcurrentTestRunners: 1
+      }, expectedInputFiles, expectedTestFramework, reporter);
+
+      firstSandbox.runMutant.returns(Promise.resolve({ status: RunStatus.Complete, tests: [{ name: 'test1', status: TestStatus.Success }] }));
+      secondSandbox.runMutant.returns(Promise.resolve({ status: RunStatus.Complete, tests: [{ name: 'test1', status: TestStatus.Success }] }));
+
+      sut.runMutants(mutants)
```
nicojs: I would rather see a return statement here to return the result of  `runMutants.then()` (also a promise) instead of using the `done` parameter in this test. Using the `done` parameter is more error prone. For example, you forgot to handle the reject scenario of this promise. Now a rejection would result in a timeout eventually, not a useful failure message

test/unit/SandboxCoordinatorSpec.ts
```diff
@@ -73,6 +78,67 @@ describe('SandboxCoordinator', () => {
...
+        .then(() =>  {
+          expect(sandbox.default).to.have.been.calledWithNew;
+          expect(sandbox.default).to.have.callCount(1);
+          expect(sandbox.default).to.have.been.calledWith({ maxConcurrentTestRunners: 1 }, 0, expectedInputFiles, expectedTestFramework, null);
+          done();
+        });
+    });
+
+    it('if config is greater than cpuCount, then runMutants should use the cpuCount', (done) => {
```
nicojs: Right now, this sentence reads *"it if config is greater than cpuCount, then runMutants should use the cpuCount"*. It would read better like *"it should use cpuCount if config is greater than cpuCount"* or something like that.

jeznag: have changed a bit. I think it's fine
image

nicojs:
I'll have a look in a couple of hours, thanks :+1:

nicojs:
Thank you! I squashed and merged it.

nicojs:
I just released v `0.5.3`. @jeznag can you confirm that it works and close the related issues?

nicojs:
> There's no yarn.lock file - should we add one

@jeznag can you please create and issue (or PR?) for this as well as explaining why we would need one? I'm not into yarn yet and don't know exactly what it would mean :grin:

jeznag:
Nice :) @nicojs do you have a suggested sample project I can run it on? I finished up at my last job a week ago and am using tape for all my side projects so don't have a good test bed anymore lol.

I created a PR for yarn.lock

nicojs:
Well... yeah... you can run it on Stryker itself :). After you've compiled the code and installed the mocha test runner `npm i stryker-mocha-runner` you can run `bin/stryker run stryker.conf.js`. I've tried it and it works for me.

jeznag:
Haha of course. Great it worked for me too :)

nicojs:
Great! I've closed the related issues.

>  I finished up at my last job a week ago and am using tape for all my side projects so don't have a good test bed anymore lol.

Well, your in luck my friend! We're looking for someone to implement the `stryker-tape-runner`. Are you interested? It basically involves implementing the [TestRunner](https://github.com/stryker-mutator/stryker-api/blob/master/src/test_runner/TestRunner.ts) interface. You can also look at the [stryker-mocha-runner](https://github.com/stryker-mutator/stryker-mocha-runner) for inspiration, as that is also a node based test runner.

jeznag:
Sure that sounds fun. Really like this project and am keen to contribute more. I'll aim to start work on that over the next month and see if I can get a POC done before the end of the year.

jeznag:
@nicojs @simondel I've put a POC together https://github.com/jeznag/stryker-tape-runner/blob/master/src/TapeTestRunner.ts but am a bit confused by the TestFramework part. What do I have to do there? Can I leave that out?

simondel:
Awesome! Let's discuss this over at #211

PR #179, opened 2016-11-10
fix(stryker.conf.js) Fix bithound status

simondel:
Builds no longer break when we have outdated dependencies. However, the build does break if we have an unused dependency.


package.json
```diff
@@ -46,7 +46,6 @@
     "lodash": "^3.10.1",
     "log4js": "^0.6.33",
     "mkdirp": "^0.5.1",
-    "node-glob": "^1.2.0",
```
simondel: Neither did I! For some reason this check was disabled..

PR #184, opened 2016-11-28
chore: Add yarn.lock file

jeznag:
yarn is an alternate npm client that can install packages
much faster than the npm client in most benchmarks. To achieve that
speed, it needs to keep a record of package dependencies so that
results are consistent between machines and so that caching can be maximised.
The dependency tree data goes in a yarn.lock file.

Having the yarn.lock file will make it easier for contributors to
install required dependencies.


simondel:
I suppose we can generate this file, right?

nicojs:
@simondel good point! Can we add the generation of this file to the build?

jeznag:
Possibly but I haven't heard of people doing that. I think updating it manually will be fine for now.
https://github.com/yarnpkg/yarn/issues/1583

If you want to automate it, the build script would need to run `yarn` and then add the lock file to git each time if there are changes. Might be a bit of work to automate.

nicojs:
@jeznag What happens when this file gets out of date? Will `yarn` not work? Or will it just downgrade the performance of a `yarn install`?

jeznag:
My hope is that I would convince you both to use yarn over npm install so the yarn.lock file would never get out of date :D

nicojs:
I'll give it a shot and come back to you ;)

nicojs:
@jeznag i tried to use yarn for a while. Though i'm not against it, i don't want to rely on this. Instead we've chosen to run our build daily to make sure things don't break.

PR #186, opened 2016-11-30
fix(TestFrameworkOrchestrator) Temporarily ignore bithound checks

simondel:


simondel:
Making Bithound issues worse while trying to fix them, great! 💪

PR #187, opened 2016-12-12
feat(sandbox): Change cwd in `Sandbox`es

nicojs:
Change the current working directory for all isolated sandboxes. This is needed to remove unwanted side effects during test runs. For example RequireJS wasn't able to find files in combination with karma.

* Add new interface `IsolatedRunnerOptions`, which extends `RunnerOptions` with the addition of `SandboxWorkingFolder`
* Change working directory as soon as the IsoltatedTestRunnerAdapterWorker has loaded the plugins as they can be relative to the parent working directory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "sinon-chai": "^2.8.0",
     "stryker-api": "^0.4.0",
     "tslint": "^3.15.1",
-    "typescript": "^2.0.2"
+    "typescript": "~2.0.2"
   },
   "peerDependencies": {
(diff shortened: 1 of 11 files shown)


simondel:
Odd. We have 36 unused dependencies..

nicojs:
Yeah, its strange right? Lets figure it out next week when we have a little more time. In the mean time, what do you think of this implementation?

nicojs:
I tested this PR by running stryker on stryker as well and it didn't break anything.

Issue #172, opened 2016-11-06
Handle duplicate files


nicojs:
Stryker should indeed be able to handle duplicate file paths. We should in that case to the same as Karma, see http://karma-runner.github.io/0.8/config/files.html "ordering".

nicojs:
Fixed since release 0.5.4 #188 
https://github.com/stryker-mutator/stryker/blob/master/CHANGELOG.md#054-2016-12-15

Issue #95, opened 2016-04-28
Exclude files from glob expression


jordyheutinck:
Node glob has an option which enables files to be excluded from a glob expression.

```
var glob = require('glob');

glob("**/*",{"ignore":['index.html', 'js', 'js/app.js', 'js/lib.js']}, function (err, files) {
  console.log(files);
})
```

[Node glob documentation](https://github.com/isaacs/node-glob#options)
Adding this feature to the Stryker config file would make configuring Stryker for larger projects a lot easier and less time consuming.

simondel:
Thanks for your input!

How would you prefer to exlude files? Would you like to use special characters such as `!` to prefix globbing patterns or would you prefer to have separate arrays in the config such as `ignoreFiles` and `ignoreMutate`? Or do you have a different solution?

jordyheutinck:
You're welcome!
I'm not sure what the most viable solution would be, but I like the idea of specifying ignored files through an array inside the files or mutate array.

For example,
 ```
config.set({
        files: [
            'src/lib/jquery/dist/jquery.js',
            'src/lib/angular/angular.js',
            'src/resources/js/**/*.js',
            ignore: [
                '**/*Mock.js'
            ]
        ],
        mutate: [
            'src/resources/js/**/*.js',
            ignore: [
                '**/SomeOtherGlobPattern.*'
            ]
        ]
});
```

For consistencies sake you could also choose to use Karma's solution, which is the separate array you mentioned.

nicojs:
I would also choose to have consistency. Like `exclude` and
`excludeMutate`. The excluded files would also count as excludeMutate files

You're welcome!
I'm not sure what the most viable solution would be, but I like the idea of
specifying ignored files through an array inside the files or mutate array.

For example,

config.set({
        files: [
            'src/lib/jquery/dist/jquery.js',
            'src/lib/angular/angular.js',
            'src/resources/js/**/*.js',
            ignore: [
                '**/*Mock.js'
            ]
        ],
        mutate: [
            'src/resources/js/**/*.js',
            ignore: [
                '**/SomeOtherGlobPattern.*'
            ]
        ]
});

For consistencies sake you could also choose to use Karma's solution, which
is the separate array you mentioned.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly or view it on GitHub
<https://github.com/stryker-mutator/stryker/issues/95#issuecomment-215378098>

nicojs:
[...]

PR #188, opened 2016-12-14
feat(exclude-files): Exclude files with a `!`

nicojs:
Add a way to exclude files. This works respecting the order of the file patterns. Only works when providing patterns as a `string`, not for the `InputFileDescriptor` format.

* Refactor `InputFileResolver`, included a new class `PatternResolver` in order to reuse functionality for both `mutate` and `files`.
* Ignore patterns starting with an `!` as long as they are provided as strings
* Deduplicate files, first occurance wins.
* Document behavior in the readme and add example to cli help text

As discussed in [stryker-api#10](https://github.com/stryker-mutator/stryker-api/pull/10)

### DIFF
--- a/src/InputFileResolver.ts
+++ b/src/InputFileResolver.ts
@@ -7,14 +7,13 @@ const log = log4js.getLogger('InputFileResolver');
 
 const DEFAULT_INPUT_FILE_PROPERTIES = { mutated: false, included: true };
 
-
 export default class InputFileResolver {
 
   private inputFileResolver: PatternResolver;
   private mutateResolver: PatternResolver;
 
   constructor(mutate: string[], allFileExpressions: Array<InputFileDescriptor | string>) {
-    validateFileDescriptor(allFileExpressions);
+    InputFileResolver.validateFileDescriptor(allFileExpressions);
     this.mutateResolver = PatternResolver.parse(mutate || []);
     this.inputFileResolver = PatternResolver.parse(allFileExpressions);
   }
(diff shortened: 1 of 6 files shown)


simondel:
This should fix #172 and #95

src/InputFileResolver.ts
```diff
@@ -4,96 +4,143 @@ import * as _ from 'lodash';
...
+        throw Error(`File descriptor ${JSON.stringify(maybeInputFileDescriptor)} is missing mandatory property 'pattern'.`);
       }
-      normalize(files);
-      return files;
-    });
+    }
+  });
+}
+
+function markAdditionalFilesToMutate(allInputFiles: InputFile[], additionalMutateFiles: string[]) {
```
simondel: Why are many of these functions outside of a class definition?

src/InputFileResolver.ts
```diff
@@ -4,96 +4,143 @@ import * as _ from 'lodash';
...
+
+  private ignore = false;
+  private descriptor: InputFileDescriptor;
 
-  private static markAdditionalFilesToMutate(allInputFiles: InputFile[], additionalMutateFiles: string[]) {
-    let errors: string[] = [];
-    additionalMutateFiles.forEach(mutateFile => {
-      if (!allInputFiles.filter(inputFile => inputFile.path === mutateFile).length) {
-        errors.push(`Could not find mutate file "${mutateFile}" in list of files.`);
+  constructor(descriptor: InputFileDescriptor | string, private previous?: PatternResolver) {
```
simondel: Why did you choose an implementation that gave every PatternResolver a link to the previous one? For me, it would make more sense when the InputFileResolver would combine each of the PatternResolvers, while the PatternResolver now does it himself.

nicojs: This way we can keep the `InputFileResolver` clean by moving part of the logic to `PatternResolver`. It's a more functional approach, right in line with my current experience programming Scala ;).

I can see where you're coming from, do you want me to move it back to the `InputFileResolver`?

src/InputFileResolver.ts
```diff
@@ -23,11 +22,46 @@ export default class InputFileResolver {
...
       const mutateFiles = results[1];
-      markAdditionalFilesToMutate(inputFiles, mutateFiles.map(m => m.path));
-      logFilesToMutate(inputFiles);
+      InputFileResolver.markAdditionalFilesToMutate(inputFiles, mutateFiles.map(m => m.path));
+      InputFileResolver.logFilesToMutate(inputFiles);
       return inputFiles;
     });
   }
+
+  private static validateFileDescriptor(maybeInputFileDescriptors: Array<InputFileDescriptor | string>) {
```
simondel: Do we still have to keep all of these methods static? They only seem to be called from inside non-static functions and they're private.

nicojs: Yes, but they don't use the `this` pointer so they can be static. Rather see them as instance functions?

simondel:
I'll be so glad when it's 2017 and we can drop Node 0.12 support because it keeps failing our build

nicojs:
Released in 0.5.4

nicojs:
We should remove the support next week. You know... as a new years gift to ourselves  :grin:

Issue #90, opened 2016-04-17
Exclude online files from globbing


simondel:
Some config files use files that are not present on the file system. 
For example:

``` js
module.exports = function(config){
  config.set({
    files: [
      'https://ajax.googleapis.com/ajax/libs/angularjs/1.3.12/angular.js',
      'src/**/*.js',
      'test/**/*.js'
    ],
    mutate: ['src/**/*.js']
  });
}
```

This gives the following warning:

```
WARNING: Globbing expression "https://ajax.googleapis.com/ajax/libs/angularjs/1.3.12/angular.js" did not result in any files.
```

simondel:
I would expect Stryker to understand that online files are just files and don't include globbing patterns.

nicojs:
Good point. However, what should happen if you say `mutated: true`?

PR #194, opened 2016-12-27
feat(InputfileResolver): exclude online files from globbing

Kattoor:
This pull request fixed issue #90

### DIFF
--- a/src/InputFileResolver.ts
+++ b/src/InputFileResolver.ts
@@ -1,15 +1,11 @@
 import { InputFile, InputFileDescriptor } from 'stryker-api/core';
-import { glob, normalize } from './utils/fileUtils';
+import { glob, normalize, isOnlineFile } from './utils/fileUtils';
 import * as _ from 'lodash';
 import * as log4js from 'log4js';
 
 const log = log4js.getLogger('InputFileResolver');
 
 const DEFAULT_INPUT_FILE_PROPERTIES = { mutated: false, included: true };
-
-function isWebUrl(pattern: string): boolean {
-  return pattern.indexOf('http://') === 0 || pattern.indexOf('https://') === 0;
-}
 
 export default class InputFileResolver {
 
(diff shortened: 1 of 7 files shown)


src/InputFileResolver.ts
```diff
@@ -7,13 +7,18 @@ const log = log4js.getLogger('InputFileResolver');
 
 const DEFAULT_INPUT_FILE_PROPERTIES = { mutated: false, included: true };
 
+function isWebUrl(pattern: string): boolean {
+  return pattern.indexOf('http://') === 0 || pattern.indexOf('https://') === 0;
```
simondel: You could also use `pattern.startsWith('http://')` 👍

src/InputFileResolver.ts
```diff
@@ -33,11 +38,26 @@ export default class InputFileResolver {
...
         }
       }
     });
   }
 
+  private validateMutationArray(mutationArray: Array<string>) {
+    if (mutationArray) {
+      mutationArray.forEach(mutation => {
+        if (isWebUrl(mutation)) {
+          throw new Error(`Cannot mutate web url "${mutation}".`)
```
simondel: You're missing a semicolon here: https://www.bithound.io/github/stryker-mutator/stryker/blob/869ee89d6fec2bf3ea98d36491e4efc7aaec8658/src/InputFileResolver.ts

src/InputFileResolver.ts
simondel: Turns out this is a part of ES6, thanks for the heads-up @Kattoor

simondel:
This may not work when we copy the files to the sandbox because the file is not on the file system. It may be smart to move `isWebUrl` to a util class so we can use it multiple times.

simondel:
I'll fix this

src/Sandbox.ts
```diff
@@ -64,6 +65,11 @@ export default class Sandbox {
   }
 
   private copyFile(file: InputFile): Promise<void> {
+    if (isOnlineFile(file.path)) {
+      this.fileMap[file.path] = file.path;
+      return Promise.resolve();
+    }
```
nicojs: Could you add an else here? That is a bit cleaner and safer IMHO.

PR #197, opened 2016-12-28
docs(README.md): Various improvements

philippw:
* Improve wording/English
* Fix broken links
* Add a "Mandatory" field to the command line options

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,35 +4,42 @@
 [![Gitter](https://badges.gitter.im/stryker-mutator/stryker.svg)](https://gitter.im/stryker-mutator/stryker?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
 ![Stryker](stryker-80x80.png)
-# Stryker
 
+# Stryker
 *Professor X: For someone who hates mutants... you certainly keep some strange company.*  
 *William Stryker: Oh, they serve their purpose... as long as they can be controlled.*
 
 ## Getting started
-Stryker is a mutation testing framework for JavaScript. It allows you to test your test by temporarily inserting bugs.
+Stryker is a mutation testing framework for JavaScript. It allows you to test your tests by temporarily inserting bugs.
 
-To install stryker, execute the command:
+To install Stryker, execute the command:
+```sh
+$ npm install stryker stryker-api --save-dev
 ```
-npm install stryker stryker-api --save-dev
-```
-***Note****: During the installation you may run into errors caused by [node-gyp](https://github.com/nodejs/node-gyp). It is safe to ignore these errors.*
 
-To test if stryker is working, execute the command:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,87 +52,105 @@ module.exports = function(config){
...
+depends what framework that specific test runner supports. By default, this value is also used for `testFramework`.  
 
-#### Coverage analysis
-**Full notation:** `--coverageAnalysis perTest`    
-**Config file key:** `coverageAnalysis: 'perTest'`    
+#### Type of coverage analysis  
+**Full notation:** `--coverageAnalysis perTest`  
+**Config file key:** `coverageAnalysis: 'perTest'`  
 **Default value:** `perTest`  
+**Mandatory**: yes
```
nicojs: It has a default value, so why should it be mandatory?

README.md
```diff
@@ -45,87 +52,105 @@ module.exports = function(config){
...
+particular mutation are tested for each mutant. 
+This does *not* influence the resulting mutation testing score. It only improves performance.
+
+The possible values are: 
+* **off**: Stryker will not determine the code covered by tests during the initial test run phase. All tests will be executed for each mutant 
+during the mutation testing phase.
+
+* **all**: Stryker will determine the code covered by all tests during the initial test run phase. Only mutants actually covered by your 
+test suite are tested during the mutation testing phase. This setting requires your test runner to be able to report the code coverage back to Stryker. 
+Currently, only the `stryker-mocha-runner` and the `stryker-karma-runner` do this.
```
nicojs: In fact, these are the only 2 runners available a.t.m. So you could also say all runners do this.

README.md
```diff
@@ -45,87 +52,105 @@ module.exports = function(config){
...
+   * The `included` property is optional and determines whether or not this file should be loaded initially by the test-runner (default: true)  
+   * The `mutated` property is optional and determines whether or not this file should be targeted for mutations (default: false)  
 
-*Note*: To include a file/folder which start with an exclamation mark (`!`), use the `InputFileDescriptor` syntax
+*Note*: To include a file/folder which start with an exclamation mark (`!`), use the `InputFileDescriptor` syntax.  
 
-#### Files to mutate
-**Command line:** `-m src/**/*.js,a.js` or `--mutate src/**/*.js,a.js`  
+#### Source code files to mutate  
+**Command line:** `--mutate src/**/*.js,a.js` or `-m src/**/*.js,a.js`
```
simondel: I noticed that with the `files` argument you use a different format: `[--files|-f] node_modules/a-lib/**/*.js,src/**/*.js,a.js,test/**/*.js` . Is there a reason why you didn't use `[--mutate|-m] src/**/*.js,a.js` here?

philippw: No, that's just an oversight. I'll also use the `[--mutate|-m]` here.

PR #206, opened 2016-12-29
docs(stryker-logo): Fix broken stryker logo

nicojs:
Stryker logo was broken after a push. Had to do with line endlings in .gitattributes file. It should be fixed now (marked *.png as binary file)


philippw:
Thanks for fixing this!

PR #210, opened 2016-12-30
fix(InputFileResolver): Don't exclude all files

nicojs:
Stop excluding of all files when a globing expression starts with a `!` and not results in files.


src/InputFileResolver.ts
```diff
@@ -105,14 +105,15 @@ class PatternResolver {
...
+      // If there is a previous globbing expression, resolve that one as well
       if (this.previous) {
-        // If there is a previous globbing expression, resolve that one as well
         return Promise.all([this.previous.resolve(), globbingTask]).then(results => {
           const previousFiles = results[0];
           const currentFiles = results[1];
           // If this expression started with a '!', exclude current files
           if (this.ignore) {
-            return previousFiles.filter(previousFile => currentFiles.some(currentFile => previousFile.path !== currentFile.path));
+            return previousFiles.filter(previousFile => currentFiles.every(currentFile => previousFile.path !== currentFile.path));
```
avassem85: Why does every work better than some. Can you explain that?

nicojs: When an expression starts with a `!` (and needs to ignore all matching files), then every file resulting from the current globing expression should not match to the current expression.

For example:
* `previousFiles: ['file1', 'file2', 'expectedFile']`
* `currentFiles: ['file1', 'file2', 'somefile']`
* Expected result: `'expectedFile'`

For an empty array, `every` returns true. More info: https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Array/every

avassem85: Ok, thanx for the explanation. De rest of the code seems good.

nicojs:
Thanks for reviewing! I'll merge it

PR #219, opened 2017-01-16
Fix channel closed error

nicojs:
Please have a look at the new `Task` api i've implemented. I think it looks a bit cleaner. Also removed the restart functionality from the `IsolatedTestRunnerAdapter` and put it in a `TimeoutDecorator` class. It looks a lot cleaner in my opinion. Its also easier to add debug functionality now, where the TestRunner runs in the same process as Stryker itself, because the timeout logic is not tightly coupled anymore with the spawning of new processes.

New functionality can now be added in `TestRunnerDecorator` classes. The `RetryDecorator` implements retries whenever a child process dies ([EPIPE](https://nodejs.org/api/errors.html) error).


nicojs:
As said during our hackathon, the `Decorator` post fix can be confusing, as people might confuse it with [TypeScript decorators](https://www.typescriptlang.org/docs/handbook/decorators.html). However, i want to keep it like this, as we're implementing a [decorator design pattern](https://sourcemaking.com/design_patterns/decorator). I've looked into ways of utilizing TypeScript decorators. It will work, but it won't be as easy to read.

I've also thought about renaming `IsolatedTestRunnerAdapter` to `IsolatedProcessDecorator`, but the constructor looks different from the `TestRunnerDecorator` base class, so we can't inherit from that (also: inheritance wouldn't add much, as we can't use any of the default implementations in that class).

nicojs:
Ok, all issues fixed. Could you have a look to it again @simondel ?

simondel:
@nicojs Could you look at this comment: https://github.com/stryker-mutator/stryker/pull/219/files/d7641b670767972834b1c0b30924d974ca6ce9a1#diff-391b42e6838a320bbe4b64de4e11e04f

src/isolated-runner/RetryDecorator.ts
```diff
@@ -0,0 +1,60 @@
...
+const BROKEN_PIPE_ERROR_CODE = 'EPIPE';
+const ERROR_MESSAGE = 'Test runner crashed. Tried twice to restart it without any luck. Last time the error message was: ';
+
+/**
+ * Wraps a test runner and implements the retry functionality.
+ */
+export default class RetryDecorator extends TestRunnerDecorator {
+  private currentRunTask: Task<RunResult>;
+
+  run(options: RunOptions): Promise<RunResult> {
```
nicojs: I'm not sure i follow your logic. The `RunOptions` here are provided in the `run` method as specified on the `TestRunner` interface.

simondel: Nevermind, I was confused with the `Config` class.

nicojs: Takes you forever to merge!

PR #222, opened 2017-01-28
chore(lifecycle): Upgrade dependencies

simondel:
Upgrade most dependencies. I left `esprima` and the `@types\estree` alone. The latest estree types break stryker because the `stryker-api` uses custom estree definitions.


nicojs:
Thanks 👍

PR #223, opened 2017-01-28
test(ReporterOrchestrator): Make test less strict

nicojs:
Test would break when adding the html reporter. Use global state for now, instead of mocking the dependency away.

### DIFF
--- a/test/unit/ReporterOrchestratorSpec.ts
+++ b/test/unit/ReporterOrchestratorSpec.ts
@@ -21,7 +21,7 @@ describe('ReporterOrchestrator', () => {
     restoreTTY();
   });
 
-  it('should register default reporters', () => {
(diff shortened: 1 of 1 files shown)


simondel:
Could you fix the build?

test/unit/ReporterOrchestratorSpec.ts
```diff
@@ -22,7 +22,7 @@ describe('ReporterOrchestrator', () => {
   });
 
   it('should register default reporters', () => {
-    expect(ReporterFactory.instance().knownNames()).to.have.lengthOf(5);
+    expect(ReporterFactory.instance().knownNames()).length.to.be.above(4);
```
simondel: Why 4?

simondel:
The build is failing based on the lodash types. This can be ignored as it's solved in the master branch.

PR #226, opened 2017-01-28
fix(bin/stryker): Changed file permissions on stryker.

Archcry:
Changed file permissions on bin/stryker to 755 so it's executable under linux operating systems by default.


nicojs:
Thanks! 👍

Issue #225, opened 2017-01-28
Strict null checks


Floormidabel:

nicojs:
I think this issue is done.

PR #230, opened 2017-01-28
Strict null checks

stienis96:
#225 Adds the strictNullChecks property to the tsconfig and fixes all the problems with enabling this property.


Floormidabel:
10/10 would merge again

src/utils/objectUtils.ts
```diff
@@ -13,7 +13,10 @@ export function freezeRecursively(target: { [customConfig: string]: any }) {
 }
 
 export function isPromise(input: void | Promise<any>): input is Promise<any> {
```
nicojs: This is probably a cleaner implementation now:

```typescript
export function isPromise(input: any): input is Promise<any> {
   return input && typeof input['then'] === 'function'; 
}
```

test/unit/InputFileResolverSpec.ts
```diff
@@ -53,7 +53,7 @@ describe('InputFileResolver', () => {
     let result: Error;
     beforeEach(() => {
       try {
-        sut = new InputFileResolver(undefined, [<any>{ included: false, mutated: true }]);
+        sut = new InputFileResolver([''], [<any>{ included: false, mutated: true }]);
```
nicojs: I think an empty array is cleaner here. That stays a bit more true to the purpose.

test/unit/SandboxCoordinatorSpec.ts
```diff
@@ -28,15 +28,30 @@ describe('SandboxCoordinator', () => {
...
+  let options: Config;
   let reporter: any;
   let coverageInstrumenter: any;
   const expectedTestFramework: any = 'expected test framework';
   const expectedRunResult = { isTheExpectedRunResult: true };
   const expectedInputFiles: any = { isInputFiles: true };
 
   beforeEach(() => {
-    options = {};
+    options = {
```
nicojs: This adds a lot of incidental detail (detail that is not needed for the test). I think its acceptable to add an `any` cast here:

```typescript
    options = <any>{};
```

What you can also do is add a helper method to the `helpers/producers.ts` file where you initialize a full config.

nicojs:
It might be better to fix these issues in #236 as that is already merged together with #234

src/reporters/BroadcastReporter.ts
```diff
@@ -11,25 +13,21 @@ export interface NamedReporter {
...
 
-  private broadcast(methodName: string, eventArgs: any) {
+  private broadcast(methodName: string, eventArgs?: any) {
     let allPromises: Promise<any>[] = [];
     this.reporters.forEach(namedReporter => {
       let reporter = <any>namedReporter.reporter;
       if (reporter[methodName] && typeof reporter[methodName] === 'function') {
         try {
-          let maybePromise: void | Promise<any> = reporter[methodName](eventArgs);
+          let maybePromise: void | Promise<any> = eventArgs ? reporter[methodName](eventArgs) : reporter[methodName]();
```
Archcry: I will look into this sometime soon, I think it's done because there is no certainty whether `eventArgs` is an actual value or `undefined` / `null`.
If there is a possibility that `eventArgs` is `undefined` / `null` the typescript compiler will complain with the newly added flags.

nicojs:
Closing this PR, because we'll merge it in #236

Issue #224, opened 2017-01-28
New mutator for `const`


JellePetersHAN:

simondel:
We won't implement this for now.

PR #231, opened 2017-01-28
feat(ConstantNumberMutator): Add new mutator.

JellePetersHAN:


JellePetersHAN:
This proposal is an implementation of #224

nicojs:
If i run stryker on stryker, the `ConstantNumberMutator` is not used even once. Is this by design? Or am i doing something wrong?

I'm using this command: `bin/stryker run stryker.conf.js --logLevel debug > output`. This writes the full log to the `output` file including killed mutants (because of `--logLevel debug`). 

I would expect the constant at [stryker.ts line 97](https://github.com/JellePetersHAN/stryker/blob/master/src/Stryker.ts#L97) to at least be mutated.

simondel:
`const` is not available in ES5 right?

JellePetersHAN:
First of all let me begin this message with an apology regarding to the lack of communication on my end, I haven't been feeling well and it's been hard for me to be productive (in my studies or otherwise).

---

As for your questions regarding the _constant number mutator_:
- @nicojs The reason the mutator doesn't get any matches on stryker is because you guys don't use any constants (at least not using the `const` keyword).
- @simondel Yes `const` is indeed a ES6 feature. I recall asking at the time of development how one goes about including ES6 dependent mutator, however I had not received an answer that would've led me to believe that it deviated from the past.

---

I've also come across a bug since the last time that I worked on the mutator partaking to the way it mutates the constants.

If someone's code contains:
```javascript
const
    name = "John Doe",
    age = 21;
```
Then the mutator will not mutate the `age` constant, because it only looks at the first declaration. It's quite easy for me to fix this, but it leaves me with a question regarding whether the mutator should regard the multiple declarations as multiple possible mutations or mutate them as a whole. Personally I would go for the former rather than the latter.

simondel:
Due to recent developments surrounding ES6 and TypeScript. I'll close this PR for now. We may implement this in one of the new mutator plugins.

PR #232, opened 2017-01-29
chore(Enhancement): Add commands to package.json

Archcry:
Added various grunt commands to the package.json, this will allow users to run most grunt commands without having grunt-cli installed globally on their system.

All new commands are accessible via `npm run`.


nicojs:
Thanks! Out of curiosity, which tool adds `.tmp.txt` files?

Archcry:
That is what I was wondering too but I could not reproduce it unfortunately.

PR #233, opened 2017-01-29
fix(typo): change not coverage to no coverage

Archcry:
Assuming this should be "no coverage" instead of "not covered" based on the context around this log message


nicojs:
Thanks!

PR #236, opened 2017-01-29
chore(Merge): async-await-typescript-2.1 & strictNullChecks

Archcry:
Combines async-await-typescript-2.1 and strictNullChecks branches into one branch.
All merge conflicts resolved, best of both worlds!


nicojs:
Thanks! I've added some review comments to #230, it might be better to fix them in this PR.

Archcry:
I implemented almost all suggested changes in this branch. Only the `EventRecorderReporter` class needs further revision.

nicojs:
@Archcry @stienis96 @wesselhendriks @Floormidabel  I did a few commits. If you agree i want to merge this branch

.bithoundrc
```diff
@@ -9,11 +9,14 @@
      ],
      "ignore": [
         "src/TestFrameworkOrchestrator.ts",
+        "src/SandboxCoordinator.ts",
+        "src/MutantTestMatcher.ts",
```
Archcry: Should we make an issue so we eventually remove these from the ignore list once the new syntax is supported?

src/Stryker.ts
```diff
@@ -103,7 +101,7 @@ export default class Stryker {
     }
   }
 
-  private async initialTestRun(inputFiles: InputFile[]): Promise<any> {
+  private async initialTestRun(inputFiles: InputFile[]) {
```
Archcry: Is it really a good idea to remove this return type?
Future releases of typescript might require you to set this return type.

.bithoundrc
nicojs: Good idea! I've already contacted the bithound team.

nicojs: I've created #237

src/Stryker.ts
nicojs: What do you mean? The function still has a return type because TypeScript can infer it to be `Promise<{ runResult: RunResult, sandboxCoordinator: SandboxCoordinator}>`, not `Promise<any>`. So, even though we specified it to be `Promise<any>`, TypeScript still knew better than to trust us. We got outsmarted by TypeScript :).

If it it cannot be inferred, we would get an error (result would be implicit `any`). So this is better in my opinion.

Archcry: Ah I did not know about that, should be okay then.

stienis96:
I just had a look at the comments of the old PR: #230. Almost everything is solved in this PR.
But the following stood out.

**[EventRecorderReporter.ts](https://github.com/stryker-mutator/stryker/blob/7d4d3c0c1c21a4d1045847f83411264bbce458c7/src/reporters/BroadcastReporter.ts#L43-L68)** had a comment that said: 

> This change actually result in a lot of code duplication. Isn't there a way to keep the functional implementation?
> 

This isn't changed in this PR.

Should we still tackle this 'problem'?

nicojs:
@stienis96 good point. I've been going back and forth on this point. I don't like code duplication, but on the other hand i do like the object oriented implementation (and the readability that come's with it). If it was pure JavaScript i would probably go with the more functional approach. What do you think?

stienis96:
@nicojs because we made this [StrictReporter](https://github.com/stryker-mutator/stryker/blob/strictNullChecks/src/reporters/StrictReporter.ts) interface we get this code duplication. We need this interface to ensure that the functions aren't null/undefined. See [Reporter.ts](https://github.com/stryker-mutator/stryker-api/blob/master/src/report/Reporter.ts), this was the old interface that was getting implemented. 

I don't know if there is a way to implement this new interface (that is needed for the strictNullChecks) with a functional programming solution in TypeScript. So that you have the following situation: one function to implement all these functions and still being sure that they are all implemented.

If there is a possibility to implement it in one function, without throwing away the StrictReporter interface, I would prefer that. Because it's a lot of code for not that much functionality. Otherwise I would leave it like it's now.

src/ReporterOrchestrator.ts
```diff
@@ -37,7 +38,7 @@ export default class ReporterOrchestrator {
       log.warn(`No reporter configured. Please configure one or more reporters in the (for example: reporter: 'progress')`);
       this.logPossibleReporters();
     }
-    return new BroadcastReporter(reporters);
+    return new BroadcastReporter(reporters) as any;
```
nicojs: This is the magic, an `any` cast basically tells TypeScript that we know best and it in fact *is* a `StrictReporter`.

nicojs:
@stienis96 I tried it in the latest commit: https://github.com/stryker-mutator/stryker/pull/236/commits/be8cad84716f65d30b4913b2513a27f4ca29a56e

Basically, i don't explicitly implement `StrictReporter`. Instead, I've implemented `Reporter`, which doesn't do anything, it might just as well not have been there. [Later in the `ReporterOrchestrator`](https://github.com/stryker-mutator/stryker/pull/236/files#diff-580963edc4d2b0e496dec15f6b63fa46R41) there is an `any` cast to make sure TypeScript ignores the fact that none of the `StrictReporter`'s methods are explicitly defined on the reporter. 

Like i said, it saves some code but complicates a lot as well. Maybe its not worth it. What do you think?

@simondel what do you think?

stienis96:
@nicojs Yeah it looks like the old way. I actually think it's still understandable, so I would leave it like this. But of course only when the build passes.

nicojs:
I'll try a different approach tonight. I actually don't like it :). I see a better implementation using the new [`keyof` type operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html)

Archcry:
Is it possible to make this discussion easily accessible in the future. This might be good information to show people that question the design choice made here.

src/reporters/BroadcastReporter.ts
```diff
@@ -11,25 +10,17 @@ export interface NamedReporter {
...
   constructor(private reporters: NamedReporter[]) {
-    ALL_EVENT_METHOD_NAMES.concat('wrapUp').forEach(method => {
-      (<any>this)[method] = (arg: any) => {
-        return this.broadcast(method, arg);
-      };
-    });
   }
 
-  private broadcast(methodName: string, eventArgs: any) {
+  private broadcast(methodName: keyof Reporter, eventArgs: any = undefined): Promise<any> | void {
```
nicojs: I use the `keyof Reporter` union type to force a string is of type `'onSourceFileRead' | 'onAllSourceFilesRead' | 'onAllMutantsMatchedWithTests' | 'onMutantTested' | 'onAllMutantsTested' | 'wrapUp'`

nicojs:
I've switched it up again. Now using the new `keyof` mapped type operator of TypeScript 2.1. This doesn't reduce the amount of code duplication, but it does introduce some type safety. I cannot just drop a use random string anymore. If you guys agree i'll merge it in tomorrow.

@Archcry 
>Is it possible to make this discussion easily accessible in the future. This might be good information to show people that question the design choice made here.

This merge request will always be kept.

Archcry:
@nicojs, I figured, but won't it be buried underneath lots and lots of other pull-requests?

stienis96:
@nicojs The `keyof` operator is a nice addition. Like you said: the code is a bit longer now, but it's all type safe now. So I'm ok with it. 👍

nicojs:
@Archcry good point! I've linked it in the issue about documenting our design choices.

nicojs:
Merged 🎉  🎊 
Thanks for all the help!

PR #238, opened 2017-02-03
Remove temporary generated TS command file

wmeints:
Found an extra file that you probably don't need.
Should be a quick fix :-)


nicojs:
Yes.... But will this pass our quality gates? ;)

simondel:
Thanks!

nicojs:
What was wrong with better code hub? Bug?

PR #241, opened 2017-02-14
fix(bithoundrc): Don't ignore productioncode in bithound

simondel:


simondel:
Now let's hope that this passes the bithound check..

simondel:
Well, at least the `SandboxCoordinator` is not causing issues anymore...

PR #244, opened 2017-02-25
fix(TypeScript 2.2): Pin version to TypeScript 2.2

simondel:
This should fix our build!


simondel:
This will be fixed in #219

PR #246, opened 2017-02-27
WIP: feat(ConfigReader): Switch cwd to config file dir

### LINKED ISSUE #245: Use stryker.conf.js file location to resolve all files
We should use the directory where the stryker.conf.js is defined as the place to resolve all files from7. Not doing this results in strange behaviors:

```bash
$ ls samples/KarmaAngular1/
app/  bower.json  bower_components/  stryker.conf.js  test/

$integration-tests (master)
$ ls samples/KarmaAngular1/test
karma.conf.js  spec/

$ cat samples/KarmaAngular1/stryker.conf.js
module.exports = function(config){
  config.set({
    karmaConfigFile: './test/karma.conf.js',
    testFramework: 'jasmine',
    testRunner: 'karma',
    plugins: [
      'stryker-karma-runner',
      'stryker-jasmine'
    ]
  });
}
$ ./node_modules/.bin/stryker run samples/KarmaAngular1/stryker.conf.js
[2017-02-27 17:16:02.140] [INFO] KarmaConfigReader - Importing config from "C:\z\github\stryker-mutator\integration-tests\test\karma.conf.js"
[2017-02-27 17:16:02.148] [DEBUG] config - Loading config C:\z\github\stryker-mutator\integration-tests\test\karma.conf.js
[2017-02-27 17:16:02.151] [ERROR] config - File C:\z\github\stryker-mutator\integration-tests\test\karma.conf.js does not exist!
```

**NOTE:** We shouldn't change the current working directory (cwd) as that might result in unexpected behavior.

nicojs:
Switch the current working directory to the directory where the config file (stryker.conf.js) is located. That way, relative paths are always resolved from that directory, which is what you would expect to happen.

Fixes #245

### DIFF
--- a/src/ConfigReader.ts
+++ b/src/ConfigReader.ts
@@ -1,7 +1,8 @@
-import { Config } from 'stryker-api/config';
-import { StrykerOptions } from 'stryker-api/core';
+import * as path from 'path';
 import * as log4js from 'log4js';
 import * as _ from 'lodash';
+import { Config } from 'stryker-api/config';
+import { StrykerOptions } from 'stryker-api/core';
 
 const VALID_COVERAGE_ANALYSIS_VALUES = ['perTest', 'all', 'off'];
 
@@ -27,6 +28,10 @@ export default class ConfigReader {
       process.exit(1);
     }
 
+    if (this.cliOptions.configFile) {
+      this.setCurrentWorkingDirectory(this.cliOptions.configFile);
+    }
+
     // merge the config from config file and cliOptions (precedence)
     config.set(this.cliOptions);
     this.validate(config);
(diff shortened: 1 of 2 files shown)


nicojs:
I liked the BetterCodeHub fail. It pointed out that the code was too complex. Thats because i'm a complex guy. Nevertheless, fixed it.

nicojs:
@simondel do you thing we should change the current working directory? Maybe its better to support a `baseDir` option in `StrykerOptions` and always work with that to resolve paths. Maybe changing the cwd can lead to strange behavior. For example, maybe the `node_modules` cannot be resolved anymore.

What do you think?

simondel:
See #245 for the reason why we do not merge this now

PR #248, opened 2017-03-02
feat(coverage): Compress coverage `perTest` size

### LINKED ISSUE #214: `coverageAnalysis: 'perTest'` does not work for large coverage results
I have not very big project, when i tried to switch on mutation on one folder of this project that contains 273 js files, stryker is stuck after executing initial test run
And throws time out exception
image

If i tried to add only some subfolders (doesn't matter which one) with 100 or 150 it's working perfect and i even can see new progress reporter and report. Usually 150 such files is somewhere about 12 000 mutations

nicojs:
Perform a lossless compression on code coverage result when running in `coverageAnalysis` mode `perTest`.

For more info: see discussion in #214

Fixes #214


simondel:
Do you still remove every line that is covered 0 times from the result?

nicojs:
Yes, because it would be in the baseline. I'll try to fix the better code hub issues tomorrow

nicojs:
I did a re-analysis (pushed a dummy commit) without changing any code and now they succeed. 

I think it compares it to the analysis on the master and our master branch analysis was out-dated (4 months old). I started a new analysis and now the warnings also appear on our main branch. That's good, we should fix that in an other PR.

PR #257, opened 2017-03-23
feat(multi-package): Migrate to multi-package repo

nicojs:
Migrate to a multi-package repo for multiple stryker npm modules. Right now only `stryker-api` and `stryker` are in here, we'll add more in the future.

* Created a lerna project (`lerna init`)
* Add `stryker` and `stryker-api` to the `packages` folder.
* Refactor `stryker-api` to *not* extend all estree nodes, instead created a new type `IdentifiedNode` which incorporates the `estree.Node` and a new interface `Identified` (`{ nodeID: number; }`). This is needed in order for the npm-linked modules te build.

There are still improvements to be made, but this is an important first step.


simondel:
Hm. Bithound doesn't yet understand that we have multiple packages which also have dependencies.

nicojs:
:+1:

Issue #101, opened 2016-05-25
Investigate es6 support


nicojs:
Do we support all language constructs from http://es6-features.org ?

simondel:
Esprima supports ES6 starting from version 2.0 ([source](http://esprima.org/doc/es6.html)). We use version `^2.7.0` by default (on my machine this is `2.7.2` as of right now).

Escodegen however does **not seem to support arrow functions** as the [online demo](https://estools.github.io/escodegen/demo/index.html) states that the following code is invalid:
`let multiply = (number) => { return number * 2 };`. I do not know which version of Escodegen the online demo is using. 

*(I have not tried every ES6 feature)*

nicojs:
How about this:

1. We create a new repository. In that repository we add a file for every es2015 feature out there. 
2. We add a build process to test if each feature is supported. Supported means:
   * Stryker supports running mutation testing on it.
3. In that build process, we test which features we *could support*. Could support means:
   * [esprima](https://github.com/jquery/esprima) supports parsing the code.
   * [escodegen](https://github.com/estools/escodegen) supports regenerating the code.
   * [instanbul](https://github.com/gotwarlost/istanbul) supports instrumenting it.
4. We document exactly which features we currently support, which features we are going to support in the near feature, which features are not implemented yet by one of our dependencies in a handy readme.
5. We create issues per feature we need to support.

CrazyBS:
I'm excited for this!  Any progress updates you are prepared to share?

simondel:
I'm working on a branch which replaces the current mutaton generation process with Babel to mutate code :) I'll share the branch soon!

simondel:
@CrazyBS Do you have any code that does not work with Stryker right now so I can test it?

CrazyBS:
@simondel Unfortunately, I am working on client-side code and can't share it.   However,  I will post the karma.conf.js and a partial package.json.   Hopefully that will give you enough to be helpful, sorry I can't give you the direct source code.

package.json
```json
{
  "private": true,
  "dependencies": {
    "babel-runtime": "^6.23.0",
    "bootstrap": "^3.3.7",
    "deep-freeze": "^0.0.1",
    "intl": "^1.2.5",
    "moment": "^2.18.1",
    "moment-business": "^3.0.1",
    "moment-timezone": "^0.5.13",
    "normalizr": "^3.2.2",
    "promise-polyfill": "^6.0.2",
    "prop-types": "^15.5.8",
[...]

PR #258, opened 2017-03-27
Add source type: module

jeznag:


simondel:
What would this fix for you? Also, we're moving to LernaJS so merging this will be a bit delayed.

SxMShaDoW:
One of my projects is also using `es6 modules`. Are we able to figure how to support es6 modules without extra configuration?

simondel:
Full ES6 support is a bit of an issue unfortunately. The module we use to genrage code is no longer supported and the module that I want to use instead isn't documented well enough to easily implement it

jeznag:
@simondel it seems to work fine for me with ES6 code as long as I either use babel-node or Node v7 + sourceType: module. What are the other ES6 issues that need to be resolved?

SxMShaDoW:
Hmm. I tried to clone your fork and npm link it without any luck. Am I missing something @jeznag ? Id love to use it with node 7. My hypothesis is supporting babel / node 7 would get a lot more visibility in the community.

jeznag:
@SxMShaDoW what errors do you get using it with NodeV7? It works fine for me. To handle import statements, I've had to change the stryker exec to have `#!/usr/bin/env babel-node`

SxMShaDoW:
@jeznag - I got similar error to the travis ci job

``` Running "ts:build" (ts) task
Compiling...
Using tsc v2.2.1
src/utils/parserUtils.ts(29,50): error TS2345: Argument of type '{ comment: boolean; loc: boolean; sourceType: string; range: boolean; tokens: boolean; }' is not assignable to parameter of type 'Options | undefined'.
  Type '{ comment: boolean; loc: boolean; sourceType: string; range: boolean; tokens: boolean; }' is not assignable to type 'Options'.
    Types of property 'sourceType' are incompatible.
      Type 'string' is not assignable to type '"script" | "module" | undefined'. 
```

thus I wasn't able to npm link it.

As far as the changes to stryker exec for imports, you should include the file change as well if people wan tot handle imports :). Still hoping a solution will exist in the future. 

Would love to see this work with [create-react-app](https://github.com/facebookincubator/create-react-app) out of the box.

simondel:
We won't fix this for now. We will fix it when supporting es6. See also https://github.com/stryker-mutator/stryker/issues/101

PR #259, opened 2017-03-29
fix(install-module): Use TypeScript 2.2 in integration tests

simondel:
I'll also fix this on the Lerna branch


nicojs:
Awesome! Thank you

PR #275, opened 2017-04-13
refactor(build): Add ability to run build steps from a child package

nicojs:
Add the ability of running build steps from child packages. For example, run `npm test` from /packages/stryker. Done this by creating a plugin outside this repo and using it as dev dependency.


nicojs:
It took some debug runs, but is fine now. Can be squashed and merged.

Issue #277, opened 2017-04-19
stryker-cli - an error occurred Error: EBUSY: resource busy or locked, rmdir


unlight:
```
[2017-04-19 14:15:24.909] [DEBUG] fileUtils - Cleaning stryker temp folder d:\Dev\angular-webpack-seed\.stryker-tmp
[2017-04-19 14:15:24.923] [ERROR] stryker-cli - an error occurred { Error: EBUSY: resource busy or locked, rmdir 'd:\Dev\angular-webpack-seed\.stryker-tmp\7446073\sandbox4100152'
  errno: -4082,
  code: 'EBUSY',
  syscall: 'rmdir',
  path: 'd:\\Dev\\angular-webpack-seed\\.stryker-tmp\\7446073\\sandbox4100152' }
```
reproduction steps:
```
git clone https://github.com/unlight/angular-webpack-seed
cd angular-webpack-seed
git checkout origin/feature/stryker
npm i
npm run stryker:run
```

expected behaviour:  
No errors

actual behaviour:
```
[2017-04-19 14:24:50.038] [DEBUG] fileUtils - Cleaning stryker temp folder d:\Dev\angular-webpack-seed\.stryker-tmp
[2017-04-19 14:24:50.075] [ERROR] stryker-cli - an error occurred { Error: EBUSY: resource busy or locked, rmdir 'd:\Dev\angular-webpack-seed\.stryker-tmp\3365224\sandbox2226615'
  errno: -4082,
  code: 'EBUSY',
  syscall: 'rmdir',
  path: 'd:\\Dev\\x\\angular-webpack-seed\\.stryker-tmp\\3365224\\sandbox2226615' }
```
windows 10
node v7.9.0
part of stryker config:
```
        testRunner: 'karma',
        testFramework: 'jasmine',
        coverageAnalysis: 'perTest',
        reporter: ['clear-text', 'progress'],
        logLevel: 'debug',
        maxConcurrentTestRunners: 1,
        karmaConfig: {
            preprocessors: {
                '**/*.js': ['webpack']
            },
            browsers: ['Nightmare'],
        },
```

simondel:
Thanks for creating this detailed issue! We think https://github.com/stryker-mutator/stryker/issues/264 will fix this.

nicojs:
Should be fixed in new styker release (`0.6.0`). Please report a new issue if needed.

Issue #264, opened 2017-04-07
Refactor utils


simondel:
The utils dir now contains a lot of code that either shouldn't be there or is already present in other modules. For example: `fileUtils.cleanFolder` is something that we can use from [rimraf](https://github.com/isaacs/rimraf)

nicojs:
Released in `0.6.0` of stryker

PR #286, opened 2017-04-21
refactor(utils): Deduplicate code and use util libraries

nicojs:
Instead of reinventing the wheel we start using modern libraries for common tasks
* [rimraf](https://npmjs.org/packages/rimraf)
* [mkdirp](https://npmjs.org/packages/mkdirp)


nicojs:
Fixes #277, #264

PR #287, opened 2017-04-22
feat($reporters): Add real time clear text reporter

jeznag:
issue-256

### DIFF
--- a/packages/stryker-api/src/core/StrykerOptions.ts
+++ b/packages/stryker-api/src/core/StrykerOptions.ts
@@ -9,11 +9,11 @@ interface StrykerOptions {
    * These include library files, test files and files to mutate, but should NOT include test framework files (for example jasmine).
    * Each element can be either a string or an object with 2 properties
    * * `string`: A globbing expression used for selecting the files needed to run the tests.
-   * * { pattern: 'pattern', included: true } : 
+   * * { pattern: 'pattern', included: true } :
    *    * The `pattern` property is mandatory and contains the globbing expression used for selecting the files
    *    * The `included` property is optional and determines whether or not this file should be loaded initially by the test-runner (default: true)
    *    * The `mutated` property is optional and determines whether or not this file should be targeted for mutations (default: false)
-   * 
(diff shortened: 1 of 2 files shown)


packages/stryker-api/src/core/StrykerOptions.ts
```diff
@@ -37,6 +37,11 @@ interface StrykerOptions {
   configFile?: string;
 
   /**
+   * Options to pass to the reporter
+   */
+  reporterOptions?: Object;
```
nicojs: This is not needed. The idea is that the typed `StrykerOptions` properties are only used for the main stryker options. Any options the reporters use need to go in their own seperate objects. For example `{ clearTextReporter: { realtime: true } }`. On line 4 of the StrykerOptions, you can see that any additional config is allowed: ` [customConfig: string]: any;`. This makes the stryker config essentially a dictionary of config.

The html reporter is doing the same thing. See the [readme](https://github.com/stryker-mutator/stryker/tree/master/packages/stryker-html-reporter#full-config-example) and the [`HtmlReporter.ts`](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker-html-reporter/src/HtmlReporter.ts#L66) for more info. 

Leaving it out of the config will make it more easier for us to spin it of in a separate package later if we want.

packages/stryker/src/reporters/ClearTextReporter.ts
```diff
@@ -7,14 +7,33 @@ const log = log4js.getLogger('ClearTextReporter');
...
+
+  private reporterOptions: any;
 
   private out: NodeJS.WritableStream = process.stdout;
 
   private writeLine(output?: string) {
     this.out.write(`${output || ''}\n`);
   }
 
+  private outputResult(result: MutantResult) {
```
nicojs: I would rather that the logic of `onAllMutantsTested` gets reused by your code, instead of creating a new similar looking method.

simondel:
@jeznag Is it ok if we close this for now?

PR #288, opened 2017-04-22
Spike at sample reduction

jeznag:


jeznag:
@nicojs @simondel I'm confused with the new structure. How do I run stryker on itself now? I get a bunch of unresolved dependencies.

packages/stryker/src/MutatorOrchestrator.ts
```diff
@@ -130,6 +130,27 @@ export default class MutatorOrchestrator {
...
+      } else {
+        mutantsOfThisType.push(mutant);
+      }
+    });
+
+    let mutantsThatMakeTheCut: Mutant[] = [];
+    Object.keys(mutantMap).forEach(mutantType => {
+      const mutantsOfThisType: Mutant[] = mutantMap[mutantType];
+      const numberOfThisTypeToKeep = Math.ceil(proportionToKeep * mutantsOfThisType.length);
+      mutantsThatMakeTheCut = mutantsThatMakeTheCut.concat(mutantsOfThisType.slice(0, numberOfThisTypeToKeep));
```
nicojs: Isn't it better to take every *nth* mutant instead of slicing off the first n mutants? That way, the mutants will be more evenly spread among the source files.

packages/stryker/src/MutatorOrchestrator.ts
```diff
@@ -26,7 +26,7 @@ export default class MutatorOrchestrator {
   /**
    * @param reporter - The reporter to report read input files to
    */
-  public constructor(private reporter: StrictReporter) {
+  public constructor(private reporter: StrictReporter, private options: {}) {
```
nicojs: I would make options be of type `StrykerOptions` here and add the option you need to `StrykerOptions`. `{}` means of type `empty object` and that is normally not what you want.

packages/stryker/src/MutatorOrchestrator.ts
```diff
@@ -130,6 +130,27 @@ export default class MutatorOrchestrator {
       }
     });
 
-    return mutants;
+    return this.reduceMutantSize(mutants, this.options.proportionOfMutantsToUse);
+  }
+
+  private reduceMutantSize(mutants: Mutant[], proportionToKeep: number) {
+    const mutantMap = {};
+    mutants.forEach(mutant => {
```
nicojs: I see what your doing here, but can't we the filtering already on line 121? That would save us the hassle of first generating the mutants and later grouping them.

packages/stryker/stryker.conf.js
```diff
@@ -6,6 +6,9 @@ module.exports = function (config) {
     reporter: ['progress', 'clear-text'],
     coverageAnalysis: 'perTest',
     logLevel: 'info',
-    plugins: ['stryker-mocha-runner']
+    plugins: ['stryker-mocha-runner'],
+    mutationOptions: {
+      proportionOfMutantsToUse: 0.1
```
nicojs: I like this grouping in `mutationOptions`. If we later add the ignoring of certain mutator types we can add that here as well.

Maybe we can call it it:

```
mutantOptions: {
   reduce: 0.9
}
```

What would feel the most natural to you?

simondel:
@jeznag Is it ok if we close this for now?

PR #269, opened 2017-04-07
 feat(stryker-initalizer): Add `init` command to stryker

avassem85:
When running `stryker init`, stryker will prompt for the test runner, test framework and reporters you want to use. It does the following: 

* Install the requested plugins locally
* Configure `stryker.conf.js` to use those plugings. Any initial configuration set by a plugin can be put in the `package.json` of that plugin in the `"initStrykerConfig" key. 
* To find which test runners, reporters and frameworks and there initStrykerConfig, to api-docs.npms.io and registry.npmjs.org respectively

Todo's:
- [x] Test in an environment with corporate proxy
- [x] Implement fallback for when there is no internet connection. Probably: give warning/error and setup a bare bones stryker.conf.js. Any other thoughts?
- [x] Implement scenario when there already is a stryker.conf.js file. Probably give an error and quit will be the best thing to do in that case.

PR #294, opened 2017-04-27
Add Boolean substitution mutators

avassem85:
- Add true -> false
- Add false -> true
- Add !a -> a


packages/stryker/package.json
```diff
@@ -65,9 +65,12 @@
     "progress": "^1.1.8",
     "rimraf": "^2.6.1",
     "serialize-javascript": "^1.3.0",
-    "tslib": "^1.5.0"
+    "tslib": "^1.5.0",
+    "inquirer": "^3.0.6"
```
nicojs: I think inquirer belongs to https://github.com/stryker-mutator/stryker/pull/269 no?

packages/stryker/src/MutatorOrchestrator.ts
```diff
@@ -86,6 +87,7 @@ export default class MutatorOrchestrator {
     mutatorFactory.register('UnaryOperator', UnaryOperatorMutator);
     mutatorFactory.register('UpdateOperator', UpdateOperatorMutator);
     mutatorFactory.register('ArrayDeclarator', ArrayDeclaratorMutator);
+    mutatorFactory.register('BooleanSubstitutionMutator', BooleanSubstitutionMutator);
```
nicojs: `'BooleanSubstitution'` (without `Mutator`)

packages/stryker/src/initializer/StrykerInitializer.ts
```diff
@@ -0,0 +1,154 @@
+'use strict';
```
nicojs: This file should be removed from this PR

packages/stryker/src/initializer/contextChoices.ts
```diff
@@ -0,0 +1,8 @@
+import {TestRunnerDefinition, TestFrameworkDefinition} from './initializer.conf';
```
nicojs: Remove

packages/stryker/src/initializer/contextChoices.ts
```diff
@@ -0,0 +1,8 @@
...
+
```
nicojs: Remove

packages/stryker/src/initializer/initializer.conf.ts
```diff
@@ -0,0 +1,51 @@
+export interface InitializerDefinitions {
```
nicojs: Remove

packages/stryker/src/initializer/stryker.conf.ts
```diff
@@ -0,0 +1,10 @@
+const strykerConfig = {
```
nicojs: Remove

packages/stryker/src/mutators/BooleanSubstitutionMutator.ts
```diff
@@ -0,0 +1,32 @@
...
+      let mutatedNode = copy(node.argument) as IdentifiedNode;
+      mutatedNode.nodeID = node.nodeID;
+      nodes.push(mutatedNode);
+    }
+
+    // true -> false or false -> true
+    if (node.type === this.type && this.booleanValues[node.raw]) {
+      let mutatedNode = copy(node);
+      mutatedNode.raw = this.booleanValues[node.raw];
+      mutatedNode.value = JSON.parse(mutatedNode.raw);
```
nicojs: I think setting the value should be enough. That way you can set it like this: `mutatedNode.value = !mutatedNode.value`

packages/stryker/src/mutators/BooleanSubstitutionMutator.ts
```diff
@@ -0,0 +1,32 @@
+import { Mutator, IdentifiedNode } from 'stryker-api/mutant';
+import { Syntax } from 'esprima';
+
+export default class BooleanSubstitutionMutator implements Mutator {
+  name = 'BooleanSubstitution';
+  private type = Syntax.Literal;
```
nicojs: make this a `readonly` property

packages/stryker/src/stryker-cli.ts
```diff
@@ -40,23 +42,24 @@ Optional location to the stryker.conf.js file as last argument. That file should
...
   }
 }
 
 if (strykerConfig) {
-  program.configFile = strykerConfig;
+  program['configFile'] = strykerConfig;
 }
 
 const commands: { [cmd: string]: () => void } = {
+  init: () => new StrykerInitializer().initialize(),
```
nicojs: Remove this line

packages/stryker/test/unit/initializer/InitializerSpec.ts
```diff
@@ -0,0 +1,338 @@
+import StrykerInitializer from '../../../src/initializer/StrykerInitializer';
```
nicojs: Remove

packages/stryker/test/unit/mutators/BooleanSubstitutionMutatorSpec.ts
```diff
@@ -0,0 +1,56 @@
...
+    // Arrange
+    const program = parse(`!a.a()`);
+    const nodeUnaryExpression = ((program.body[0] as estree.ExpressionStatement).expression as estree.UnaryExpression & Identified);
+
+    // Act
+    const result = sut.applyMutations(nodeUnaryExpression, copy)[0] as estree.Expression & Identified;
+
+    // Assert
+    expect(result).to.be.ok;
+    expect(result.type).to.be.eq(nodeUnaryExpression.argument.type);
```
nicojs: I think its safer to use `Syntax.UnaryExpression` instead of `nodeUnaryExpression.argument.type` here. `nodeUnaryExpression.argument.type` could be something that you don't expect at runtime, since you've casted it before.

packages/stryker/test/unit/mutators/BooleanSubstitutionMutatorSpec.ts
```diff
@@ -0,0 +1,56 @@
...
+    // Act
+    const result = sut.applyMutations(nodeLiteral, copy)[0] as estree.Literal & Identified;
+
+    // Assert
+    expect(result).to.be.ok;
+    expect(result.raw).to.be.eq('true');
+    expect(result.value).to.be.true;
+    expect(result.nodeID).to.be.eq(nodeLiteral.nodeID);
+  });
+});
```
nicojs: I'm missing a unit test for when _not_ to mutate.

packages/stryker/src/mutators/BooleanSubstitutionMutator.ts
avassem85: Not set the raw value?

packages/stryker/test/unit/mutators/BooleanSubstitutionMutatorSpec.ts
avassem85: It will most probably not be a Syntax.UnaryExpression. In the tree I move the Expression(any Expression) in the UnaryExpression (argument) to the position of the UnaryExpression.

packages/stryker/src/mutators/BooleanSubstitutionMutator.ts
```diff
@@ -0,0 +1,32 @@
...
+    
+    // !a -> a
+    if (node.type === Syntax.UnaryExpression && node.operator === '!') {
+      let mutatedNode = copy(node.argument) as IdentifiedNode;
+      mutatedNode.nodeID = node.nodeID;
+      nodes.push(mutatedNode);
+    }
+
+    // true -> false or false -> true
+    if (node.type === this.type && this.booleanValues[node.raw]) {
```
nicojs: I think this.type is kind of strange here in retrospect. Why not add `Syntax.Literal` here directly. It doesn't make sense for BooleanSubstitutionMutator.type to be `Syntax.Literal`

packages/stryker/src/mutators/BooleanSubstitutionMutator.ts
```diff
@@ -0,0 +1,32 @@
...
+      let mutatedNode = copy(node);
+      mutatedNode.raw = this.booleanValues[node.raw];
+      mutatedNode.value = !mutatedNode.value;
```
nicojs: I've just tested it:

```javascript
const esprima = require('esprima');
const escodegen = require('escodegen');
const code = esprima.parse('a = true');
code.body[0].expression.right.value = false;
console.log(escodegen.generate(code));
// 'a = false'
```

You never have to use `raw` value. I would just see if `typeof node.value === 'boolean'` and than set the value to `! value`, this makes it a lot simpler IMHO.

packages/stryker/src/mutators/BooleanSubstitutionMutator.ts
```diff
@@ -0,0 +1,30 @@
...
+    
+    // !a -> a
+    if (node.type === Syntax.UnaryExpression && node.operator === '!') {
+      let mutatedNode = copy(node.argument) as IdentifiedNode;
+      mutatedNode.nodeID = node.nodeID;
+      nodes.push(mutatedNode);
+    }
+
+    // true -> false or false -> true
+    if (node.type === Syntax.Literal && this.booleanValues[node.raw]) {
```
nicojs: I think `this.booleanValues` is not needed anymore. You can use `typeof node.value === 'boolean'` instead.

packages/stryker/src/mutators/BooleanSubstitutionMutator.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export default class BooleanSubstitutionMutator implements Mutator {
+  name = 'BooleanSubstitution';
+  private booleanValues: { [targetedOperator: string]: string } = {
+    'true': 'false',
+    'false': 'true'
+  };
+
+  applyMutations(node: IdentifiedNode, copy: <T>(obj: T, deep?: boolean) => T): IdentifiedNode[] {
+    let nodes: IdentifiedNode[] = [];
```
nicojs: Try to use `const` instead of `let` if you don't need to override the variable.

packages/stryker/src/stryker-cli.ts
```diff
@@ -14,6 +14,7 @@ program
   .usage('<command> [options] [stryker.conf.js]')
   .description(`Possible commands: 
     run: Run mutation testing
+    init: Initalize Stryker for your project
```
nicojs: Remove this line

PR #300, opened 2017-05-05
feat(mocha-framework): Move mocha test framework to seperate package

nicojs:
Move mocha test framework to its seperate package. This should prevent confusion about the state of the mocha test framework plugin. It only works with the stryker-mocha-runner test runner plugin at the moment, so it will downgrade `coverageAnalysis: 'perTest'` to `coverageAnalysis: 'all'` when another test runner is used.

BREAKING CHANGE: Users with `testRunner: "mocha",testFramework: "mocha"` should now also install "stryker-mocha-framework".


nicojs:
I don't know why travis is not cooperating. I'll open a new PR.

PR #302, opened 2017-05-05
chore: Add better code hub batch

nicojs:


simondel:
Of course Better Code Hub approves this change ;)

PR #309, opened 2017-05-18
feat(report-score-result): Report score result as tree

nicojs:
Add score result table to clear-text reporter. Tada:

```
-----------|---------|----------|-----------|------------|----------|---------|
File       | % score | # killed | # timeout | # survived | # no cvg | # error |
-----------|---------|----------|-----------|------------|----------|---------|
All files  |   69.57 |       16 |         0 |          3 |        4 |       0 |
 Add.js    |   82.00 |       14 |         0 |          1 |        2 |       0 |
 Circle.js |   33.00 |        2 |         0 |          2 |        2 |       0 |
-----------|---------|----------|-----------|------------|----------|---------|
```

* Add interface `ScoreResult` to stryker-api. Loosely based on the `stryker-html-reporter` implementation
* Add callback `OnScore` to reporter API
* Add class `ScoreResultCalculator` to stryker, responsible for calculating the score results per directory / file
* Add class `ClearTextScoreTable` to reporters folder, responsible for converting a `ScoreResult` to a clear text table.
* Removed the previous summary: 

```
17 total mutants.
8 mutants survived.
0 mutants timed out.
9 mutants killed.
Ran 3.00 tests per mutant on average.
Mutation score based on covered code: 52.94%
Mutation score based on all code: 52.94%
```


nicojs:
Pretty sure the build is going to be green. Do you might having a look @simondel ?

packages/stryker/src/ScoreResultCalculator.ts
```diff
@@ -0,0 +1,125 @@
...
+      return uniqueFileDirectories
+        .reduce((previousDirectories, currentDirectories) => previousDirectories.filter((token, index) => currentDirectories[index] === token))
+        .join(path.sep);
+    } else {
+      return '/';
+    }
+  }
+
+  private static calculateNumbers(mutantResults: MutantResult[], childScores: ScoreResult[]) {
+    if (childScores.length) {
```
simondel: If we would pass in a `MutationScoreResult` interface for the `MutationScoreResult` that we're working on instead of childScores, we could just check if the current `MutationScoreResult` is a directory. 

I would also suggest moving the check to the `calculate` method. With this method you're just obfuscating why we need both arrays. We could have two methods: `addFolderMutationScore(mutationScoreResult: MutationScoreResult)` and `addFileMutationScore(mutationScoreResult: MutationScoreResult, mutantResults: MutantResult[])`

packages/stryker/src/ScoreResultCalculator.ts
```diff
@@ -0,0 +1,125 @@
...
+      return this.calculateNumbersBasedOnChildScores(childScores);
+    } else {
+      return this.countNumbers(mutantResults);
+    }
+  }
+
+  private static calculateNumbersBasedOnChildScores(childScores: ScoreResult[]) {
+    const sum = (expression: (score: ScoreResult) => number) => _.sum(childScores.map(expression));
```
simondel: Nice 😎

packages/stryker/src/ScoreResultCalculator.ts
```diff
@@ -0,0 +1,125 @@
...
+    };
+  }
+
+  private static countNumbers(mutantResults: MutantResult[]) {
+    let killed = 0;
+    let timedOut = 0;
+    let survived = 0;
+    let noCoverage = 0;
+    let errors = 0;
+    mutantResults.forEach(mutation => {
```
simondel: Can't we use a similar `sum` method here, just like in `calculateNumbersBasedOnChildScores`?

packages/stryker/src/ScoreResultCalculator.ts
```diff
@@ -0,0 +1,125 @@
...
+      killed,
+      survived,
+      noCoverage,
+      errors,
+      timedOut,
+      totalDetected,
+      totalUndetected,
+      totalMutants,
+      totalCovered,
+      mutationScore: Math.floor(totalDetected / totalMutants * 100) || 0,
```
simondel: I don't look at code coverage reports often but do they also always give full percentages or do they use decimals?

packages/stryker/src/reporters/EventRecorderReporter.ts
```diff
@@ -33,9 +33,8 @@ export default class EventRecorderReporter implements StrictReporter {
     return this._baseFolder;
   }
 
-
-  private writeToFile(index: number, methodName: keyof Reporter, data: any) {
-    let filename = path.join(this.baseFolder, `${this.format(index)}-${methodName}.json`);
+  private writeToFile(methodName: keyof Reporter, data: any) {
+    let filename = path.join(this.baseFolder, `${this.format(this.index++)}-${methodName}.json`);
```
simondel: Are you sure you want index++? I know that we did that in the original code but that was called from another method. I'm not sure if the postincrement in `myMethod(this.index++)` is called before or after the call has been made.

packages/stryker/test/unit/reporters/ClearTextReporterSpec.ts
```diff
@@ -33,22 +37,60 @@ describe('ClearTextReporter', function () {
...
+          noCoverage: 4,
+          errors: 5,
+          mutationScore: 80,
+          childResults: [scoreResult({
+            name: 'child1',
+            mutationScore: 60,
+            childResults: [
+              scoreResult({
+                name: 'some/test/for/a/deep/file.js',
+                mutationScore: 59.99
```
simondel: Aren't the scores always floored?

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+import * as _ from 'lodash';
+import * as chalk from 'chalk';
+
+const FILES_ROOT_NAME = 'All files';
+
+interface TableCellValueFactory {
+  (row: ScoreResult, ancestorCount: number): string;
+}
+
+const spaces = (n: number): string => {
```
simondel: Why not use `new Array(n + 1).join(' ');`?

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (!input.endsWith(' ')) {
+      return this.pad(input + ' ', targetLength);
+    }
+    // Start with space
+    else if (!input.startsWith(' ')) {
+      return this.pad(' ' + input, targetLength);
+    }
+    else if (input.length < targetLength) {
+      // Pad rest with spaces in front
+      return spaces(targetLength - input.length) + input;
```
simondel: Can't we combine the last two elseif-statements with the else-statement if we just return `return spaces(targetLength - input.length) + input;`?

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+      return input;
+    }
+  }
+
+  drawLine(length = this.width): string {
+    if (length) {
```
simondel: See my comment on the `spaces` method about using `new Array()`

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+
+const spaces = (n: number): string => {
+  if (n) {
+    return ` ${spaces(n - 1)}`;
+  } else {
+    return '';
+  }
+};
+
+const dots = (n: number): string => {
```
simondel: See my comment on the `spaces` method about using `new Array()`

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+      return '';
+    }
+  }
+
+  drawTableCell(score: ScoreResult, ancestorCount: number) {
+    return this.color(score)(this.pad(this.valueFactory(score, ancestorCount)));
+  }
+
+  drawHeader() {
```
simondel: This doesn't draw the header it just creates it.

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  drawLine(length = this.width): string {
+    if (length) {
+      return '-' + this.drawLine(length - 1);
+    } else {
+      return '';
+    }
+  }
+
+  drawTableCell(score: ScoreResult, ancestorCount: number) {
```
simondel: See my comment on `drawHeader`

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+    return this.color(score)(this.pad(this.valueFactory(score, ancestorCount)));
+  }
+
+  drawHeader() {
+    return this.pad(this.header);
+  }
+
+  protected color(score: ScoreResult) {
+    return (input: string) => input;
```
simondel: What does this do?

packages/stryker/src/reporters/ClearTextScoreTable.ts
simondel: Isn't there a `drawFooter`?

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+
+class FileColumn extends Column {
+  constructor(rows: ScoreResult) {
+    super('File', (row, ancestorCount) => spaces(ancestorCount) + (ancestorCount === 0 ? FILES_ROOT_NAME : row.name), rows);
+  }
+  protected pad(input: string, targetLength: number = this.width): string {
+    if (!input.endsWith(' ')) {
+      return this.pad(`${input} `, targetLength);
+    } else if (input.length < targetLength) {
+      return input + spaces(targetLength - input.length);
```
simondel: Can't we reduce the content of this method to just this one line and still end up with the same result?

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+  private columns: Column[];
+
+  constructor(private score: ScoreResult) {
+    this.columns = [
+      new FileColumn(score),
+      new MutationScoreColumn(score),
+      new Column('# killed', row => row.killed.toString(), score),
+      new Column('# timeout', row => row.timedOut.toString(), score),
+      new Column('# survived', row => row.survived.toString(), score),
+      new Column('# no cvg', row => row.noCoverage.toString(), score),
```
simondel: I think `no cov` is clearer

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+    else if (input.length < targetLength) {
+      // Pad rest with spaces in front
+      return spaces(targetLength - input.length) + input;
+    } else {
+      return input;
+    }
+  }
+
+  drawLine(length = this.width): string {
```
simondel: See my comment on drawHeader

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+      new Column('# error', row => row.errors.toString(), score)
+    ];
+  }
+
+  private drawBorder() {
+    return this.drawPerColumn(column => column.drawLine());
+  }
+
+  private drawHeader() {
+    return this.drawPerColumn(c => c.drawHeader());
```
simondel: Aren't you drawing per row instead of column?

packages/stryker/src/reporters/ClearTextScoreTable.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  private drawPerColumn(toDraw: (col: Column) => string) {
+    return this.columns.map(toDraw).join('|') + '|';
```
simondel: Why not prepend with a `'|'` as well?

packages/stryker-api/src/report/ScoreResult.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * Represents a score result of a file or directory
+ */
+interface ScoreResult {
```
nicojs: As discussed offline, we shouldn't change it. It holds the `mutationScore` among others, so it shouldn't be `MutationScoreResult`. I couldn't find a better name than `ScoreResult`, so leaving it as is.

packages/stryker-api/src/report/Reporter.ts
```diff
@@ -38,9 +39,15 @@ interface Reporter {
   onAllMutantsTested?(results: MutantResult[]): void;
 
   /**
-   * Called when stryker wants to quite
+   * Called when the mutation score is calculated.
+   * @param score The immutable structured score result (tree)
+   */
+  onScore?(score: ScoreResult): void;
```
nicojs: As discussed offline, we should rename it to `onScoreResult`  or `onScoreCalculated`.

packages/stryker-api/src/report/ScoreResult.ts
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * The file or directory name of this score result node
+   */
+  readonly name: string;
+  
+  /**
+   * Any child directory/file score result nodes
+   * If this score result represents a file, the length will be 0
+   */
+  readonly childResults: ReadonlyArray<ScoreResult>;
```
nicojs: >  I would also expect a property `parent`.
Why? A tree doesn't necessarily have a parent. You normally don't need it and would complicate the creating of the thing (cannot use a recursive function).

> I would suggest either having the methods: isFile() and isDirectory()

I understand where you're coming from, but I feel as a consumer of `ScoreResult`, you shouldn't care about whether it is a file or a folder. Take the root for example. That isn't strictly a directory, rather its a group of directories and files (which is most likely a directory, but doesn't have to be).

>  (we could over engineer this and split it into a base interface and two child interfaces for the file and directory. The directory interface could add the children property)
True, that is what i did in the html reporter :sweat:

packages/stryker/src/ScoreResultCalculator.ts
```diff
@@ -0,0 +1,125 @@
...
+    const name = this.determineCommonBasePath(results, basePath);
+    const childResults = this.calculateChildScores(results, name, basePath);
+    const numbers = this.calculateNumbers(results, childResults);
+    return freezeRecursively(_.assign(numbers, { name, childResults }));
+  }
+
+  private static compareScoreResults(a: ScoreResult, b: ScoreResult) {
+    const sortValue = (scoreResult: ScoreResult) => {
+      // Directories first
+      if (scoreResult.childResults.length) {
```
nicojs: True... i'll have another look at it

packages/stryker/src/ScoreResultCalculator.ts
```diff
@@ -0,0 +1,125 @@
+import { MutantResult, MutantStatus, ScoreResult } from 'stryker-api/report';
+import { freezeRecursively } from './utils/objectUtils';
+import * as _ from 'lodash';
+import * as path from 'path';
+
+export default class ScoreResultCalculator {
```
nicojs: As we're not renaming the `ScoreResult` to `MutationScoreResult`, i would suggest leaving the name as it is.

packages/stryker/src/ScoreResultCalculator.ts
```diff
@@ -0,0 +1,125 @@
...
+        return `0${scoreResult.name}`;
+      } else {
+        return `1${scoreResult.name}`;
+      }
+    };
+    return sortValue(a).localeCompare(sortValue(b));
+  }
+
+  private static calculateChildScores(results: MutantResult[], name: string, basePath: string) {
+    const childrenBasePath = path.join(basePath, name) + path.sep;
```
nicojs: Hmm true. I've added a test for this and fixed it.

packages/stryker/src/ScoreResultCalculator.ts
```diff
@@ -0,0 +1,125 @@
...
+    const resultsByFiles = _.groupBy(results, result => result.sourceFilePath.substr(childrenBasePath.length));
+    const uniqueFiles = Object.keys(resultsByFiles);
+    if (uniqueFiles.length > 1) {
+      const filesGroupedByDirectory = _.groupBy(uniqueFiles, file => file.split(path.sep)[0]);
+      return Object.keys(filesGroupedByDirectory)
+        .map(directory => this.calculate(_.flatMap(filesGroupedByDirectory[directory], file => resultsByFiles[file]), childrenBasePath))
```
nicojs: Renamed it to `calculateScoreResult` :+1:

nicojs: Let's say we have mutant results: 

```json
[{
   "sourceFilePath": "a/b",
   ...
},{
   "sourceFilePath": "a/c",
  ...
},{
   "sourceFilePath": "b/c",
  ...
}]
```

Then `filesGroupedByDirectory` will look like:

```json
{ 
  "a": ["a/b" "a/c"],
  "b": ["b/c"]
}
```

Now if we for each file lookup the mutant results here:  `file => resultsGroupedByFiles[file]`, you get an array of arrays: `MutantResult[][]`. We want to flatMap that to a `MutantResult[]`

packages/stryker/src/ScoreResultCalculator.ts
nicojs: > If we would pass in a MutationScoreResult interface for the MutationScoreResult that we're working on instead of childScores, we could just check if the current MutationScoreResult is a directory.

No we cannot. The `ScoreResult` interface is entirely `readonly` (something we should also add to other reporter interfaces). So after instantiating, we cannot alter the properties anymore. We could of course copy the values to a new instance, but i actually like it this way.

> I would also suggest moving the check to the calculate method.

I wouldn't do that. I think the `if` statement represents a new layer of abstraction, which i like to add to it's own function.

Note. We could remove the `calculateNumbersBasedOnChildScores` method all together. It's a performance improvement.

packages/stryker/src/ScoreResultCalculator.ts
nicojs: True. This is a major bug. We don't want to round any numbers here, if a reporter wants to do that, it can, but stryker should only give you the raw value

packages/stryker/src/reporters/ClearTextScoreTable.ts
nicojs: Because i didn't know it existed and I didn't talk to you about it. 👍

packages/stryker/src/reporters/ClearTextScoreTable.ts
nicojs: I don't think so. That won't guarantee that it starts with a space. The code right now really relies on the `pad` function at least adding a space even if it would exceed the target length. ... I hate drawing tables in clear text ;)

packages/stryker/src/reporters/ClearTextScoreTable.ts
nicojs: True. Every `draw...` function just creates a string. Even `draw` itself. I chose to call it `draw` as that is basically what you're doing. Should we change it?

nicojs: No, right now, the footer is a simple border

packages/stryker/src/reporters/ClearTextScoreTable.ts
nicojs: It is protected, so an implementing Column can choose to override it (like the `MutationScoreColumn` does)

packages/stryker/src/reporters/ClearTextScoreTable.ts
nicojs: Apparently yes: 

```typescript
    return `${input} ${spaces(this.width - input.length - 1)}`;
```

Awesome! :+1:

packages/stryker/src/reporters/ClearTextScoreTable.ts
nicojs: I've shamelessly copied the clear text table of [`nyc`](https://github.com/istanbuljs/nyc)

packages/stryker/src/reporters/EventRecorderReporter.ts
nicojs: It doesn't matter if it's before or after. As long as we get unique file names alphabetically ordered, i'm happy

packages/stryker/test/unit/reporters/ClearTextReporterSpec.ts
nicojs: That was the problem. They sometimes were and sometimes were not. Now they're always kept raw.

nicojs:
@simondel thanks for the thorough review. Code is much cleaner now. I think I've fixed or commented on all your remarks. Could you look at it again?

packages/stryker/src/ScoreResultCalculator.ts
nicojs: Yeah, I left this part as is. I think its fine the way it is.

packages/stryker/src/reporters/ClearTextScoreTable.ts
nicojs: No, its per column right? A value in each column is a row.

packages/stryker/src/ScoreResultCalculator.ts
simondel: We won't fix this until we have to know the difference between a file and a directory in a reporter.

packages/stryker-api/src/report/ScoreResult.ts
simondel: We won't add `isDirectory()` or `isFile()` until we have to know the difference between a file and a directory in a reporter.

packages/stryker/src/reporters/ClearTextScoreTable.ts
simondel: I would change it to something like `createHeader` or `generateHeader`

PR #313, opened 2017-06-03
fix(MethodChainMutatorSpec): Fix group name, so it matches the name of the mutator.

JellePetersHAN:


simondel:
Nice find!

nicojs:
Thanks! Next time please scope a commit like this with `test(..)`. That way, it won't end up in our release notes as a bug being fixed. :+1:

PR #321, opened 2017-06-10
fix(npmignore): Align npm ignores

nicojs:
Some .npmignore files were too forgiving, so too much files were installed. Some of those fat ladies could lose some weight.


simondel:
Shouldn't we also pack the package.json?

nicojs:
No, package.json is one of the files you cannot ignore. This is the complete list of files you CANNOT ignore:

* package.json
* README (and its variants)
* CHANGELOG (and its variants)
* LICENSE / LICENCE

https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

nicojs:
@simondel i've updated the .npmignore file for stryker-api, it contained one duplicate line. Can we merge this now?

Issue #312, opened 2017-06-03
Stryker should not do a dry-run when there are no files


simondel:
Repro steps:
* Create a new directory
* Call `npm init` and skip through it
* Install mocha: `npm i -D mocha`
* Call `stryker init` and choose *mocha - mocha*
* Run stryker

Current behavior:
![image](https://cloud.githubusercontent.com/assets/4221997/26750993/92709336-482f-11e7-9bd8-20216d7f9e55.png)

Expected behavior:
No crash. Just a message that no files have been found.

legopiraat:
I'll take this one

simondel:
Awesome! If you need any help, feel free to let us know :)

simondel:
Thanks for fixing this @legopiraat!

PR #324, opened 2017-06-15
Fix: No dry run when there are no files

legopiraat:
Stryker will no longer do a dry run when there are no files.

Related issue: #312


nicojs:
Thanks @legopiraat !

PR #330, opened 2017-06-17
docs(README): Add part about the cli

korthout:
I've updated the README's section on Usage to also discuss using the `init` command and I've added a part about the command-line interface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,12 +61,11 @@ module.exports = function(config){
 }
 ```
 
-As you can see, the config file is *not* a simple JSON file. It should be a common js (a.k.a. npm) module. You might recognize this way of working from the karma test runner.
+As you can see, the config file is *not* a simple JSON file. It should be a common js (a.k.a. node) module. You might recognize this way of working from the karma test runner.
 
 Make sure you *at least* specify the `files` and the `testRunner` options when mixing the config file and/or command line options.
 
 ## Command-line interface
-
 Stryker can also be installed, configured and run using the [Stryker-CLI](https://github.com/stryker-mutator/stryker-cli). If you plan on using Stryker in more projects, the Stryker-CLI is the easiest way to install, configure and run Stryker for your project.
 
 You can install the Stryker-CLI using:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -59,11 +61,20 @@ module.exports = function(config){
...
 
-As you can see, the config file is *not* a simple JSON file, it should be a common js (a.k.a. npm) module.  
-You might recognize this way of working from the karma test runner.
+As you can see, the config file is *not* a simple JSON file. It should be a common js (a.k.a. npm) module. You might recognize this way of working from the karma test runner.
 
 Make sure you *at least* specify the `files` and the `testRunner` options when mixing the config file and/or command line options.
 
+## Command-line interface
+
+Stryker can also be installed, configured and run using the [Stryker-CLI](https://github.com/stryker-mutator/stryker-cli). If you plan on using Stryker in more projects, the Stryker-CLI is the easiest way to install, configure and run Stryker for your project.
```
nicojs: Maybe we can add some information here about the working of `stryker-cli`. Mainly that it is only used as a proxy to the local `stryker` instance.

README.md
```diff
@@ -59,11 +61,20 @@ module.exports = function(config){
 }
 ```
 
-As you can see, the config file is *not* a simple JSON file, it should be a common js (a.k.a. npm) module.  
-You might recognize this way of working from the karma test runner.
+As you can see, the config file is *not* a simple JSON file. It should be a common js (a.k.a. npm) module. You might recognize this way of working from the karma test runner.
```
nicojs: It's actually not called an 'npm' module, but a 'node' module. :) just realize that now.

korthout:
@nicojs I've made the requested changes. Please let me know how you feel about them.

korthout:
I've also updated the README of the `stryker-cli` repo. This last commit is ment to make the descriptions for usage of the CLI more friendly and the same for both repos.

korthout:
@nicojs I think this is ready to be merged. Can you take a final look?

PR #341, opened 2017-07-15
feat(html-reporter): Score result as single source of truth

### LINKED ISSUE #335: Incorrect relative path for html reporter
_This is a repost :pencil: of @michaelpcote [stryker-html-reporter issue #13](https://github.com/stryker-mutator/stryker-html-reporter/issues/13)_
\---------------------------------------------------------------------------------------------------------------------
I've run into an issue that is causing 404 errors that makes the html reports unusable.

The issue occurs whenever the report is able to condense a path to something like:
- parentDir/childDir

This happens because there is only 1 child directory or because only 1 child directory has .js files. For instance, because we are grouping our features together like this:
- feature
  - js (contains directives, services, controllers)
  - css
  - partials

It is formatting incorrectly every time. The issue is that the report on the upper level will show
feature/js. When you click into this link it then shows the js directory and all the files completely unstyled. The index.html for that file then is trying to reference the bootstrap and stryker.js files 1 level up, like "../bootstrap" when it should actually be "../../bootstrap".

nicojs:
Previously we've introduced a reporter event that should function as a single source of truth for all stryker score calculations. Now it is also used in the HTML reporter

* Add properties `representsFile` and `path` to the `ScoreResult` interface, they are used in the html reporter
* Remove all calculations from the html reporter (replaced by the `onScoreResult` reporter callback)
* Refactor: replace dependency on `handlebarsjs` with `typed-html`
* Migrate all handlebarsjs templates to `typed-html` templates
* Refactor the integration tests, much cleaner now.
* Add unit tests for the new implementation of the `HtmlReporter`.
* Small redesign of the html report: remove the double table logic. Every page will now have one table only.
* Fix issue 335, use correct depth to determine the root of the reporter files
* Update ui (protractor) tests

Fixes #335


nicojs:
I'm going on holiday for couple of days. Feel free to make changes and merge this in. I'm very happy to get rid of the old html reporter code.

@simondel watch out when releasing this: it needs the new `ScoreResult` reporter interface, so don't forget to update the peer to `stryker-api`
@korthout would you mind taking a look with a clean-code eye 👍 I'm curious to see if the new code is cleaner.

j-truax:
Any ETA on when this will get committed to master? Anxious to test to see if it fixes relative path issue #335

simondel:
I was planning on testing it some more tomorrow morning. Could you perhaps upload a project setup that had issues with the previous reporter so I can determine if it's fixed?

nicojs:
> Could you perhaps upload a project setup that had issues with the previous reporter so I can determine if it's fixed?

@simondel I added reproduction steps to the issue and than tested it with the bits from this PR. Works fine.

simondel:
Looks good! I removed the package-lock.json from the html reporter

nicojs:
So merge?

simondel:
Yes, you can merge this :)

j-truax:
@nicojs & @simondel Sorry for not getting back to you both sooner with an example. It has been a hectic couple weeks with work & school. I had tried to describe it in #335 , hopefully that helped. I will get the latest version and test it out on our end. Thank you both for the attention on this issue. Very much appreciated!

nicojs:
Its merged in, but not released yet. Will do that later today

PR #351, opened 2017-08-09
feat(IsolatedTestRunner): Handle promise rejections

### LINKED ISSUE #350: Mutation testing in node based environments can generate false warnings
When running stryker using a node-based test runner like mocha, we generally see some warnings:

```
(node:4860) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: fail
(node:4860) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:4860) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 1)
fail
```

If you are new to these messages, read up here: https://stackoverflow.com/questions/40920179/should-i-refrain-from-handling-promise-rejection-asynchronously/40921505

This seems pretty serious to end users, but it isn't really. After all, we mutated the code. These warnings can be expected.

We can add handlers in the sandboxes for these warning like this:

```javascript
process.on('unhandledRejection', function (reason) {
  log.info(`Noticed an unhandled promise rejection, reason: ${reason}`);
});
process.on('rejectionHandled', (reason) => {
  log.info(`Noticed an unhandled promise (reported earlier) was handled after all, but asyncronously. Reason: ${reason}`);
});
```

**Questions:**

1. Do we still want to log something in these handlers? (IMHO: yes, but maybe as debug)
2. Do we want the initial test run to fail when an unhandled (or asyncronous handled) promise occurs? (IMHO: no, it might be the choice of the implementer).

nicojs:
Handle unhandled promise rejection warnings by handling the [`unhandledRejection`](https://nodejs.org/api/process.html#process_event_unhandledrejection) and [`rejectionHandled`](https://nodejs.org/api/process.html#process_event_rejectionhandled) events on `process`. This way, the following messages will not show up in your error log anymore:

* UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: ${unhandledPromiseId}): ${reason}
* PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: ${unhandledPromiseId})

Note: They will still show up when using `logLevel: 'trace'`

Fixes #350


simondel:
Looks good!

PR #355, opened 2017-08-11
feat(ci-integration): Configurable thresholds based on mutation score

### LINKED ISSUE #220: Continuous Integration compatibility
Currently Stryker always returns exit code 0, which makes it hard to integrate with CI servers and similar tooling.

It would be good to have some way of reporting "failure". Ideally this would be via exit code and "failure" would probably be a configurable threshold of mutation score, though it could be a fixed value like "mutation score on covered code was less than 100%".

nicojs:
* Add `thresholds` to the `Config` class and `StrykerOptions`. With this, you can configure the `high`, `low` and `break` values.
* Move validation logic from `Stryker` class to a new `ConfigValidator`.
* Add validaton for configured threshold values
* Add a mechanism to exit with exit code 1 if the mutation score comes below the `thresholds.break` value.
* Update the `ConsoleReporter` and `HtmlReporter` to make use of the threshold values for coloring.
* Refactor stryker unit tests to now mock away functionality more cleanly
* Update test command in package.json to allow for html coverage reporters

Fixes #220


riezebosch:
A rather dumb feature using a very arbitrary metric. Please let the configured reporters decide the desired outcome (so the baseline-reporter will have a vote).

nicojs:
Desired outcome of what? Breaking the build? Or score coloring?

riezebosch:
Breaking the build. Exit code.

nicojs:
Ok. Both can be influenced by a plugin using a `ConfigEditor` plugin: https://github.com/stryker-mutator/stryker/blob/master/packages/stryker-api/src/config/ConfigEditor.ts

The basic setup:

```typescript
import { ConfigEditor, ConfigEditorFactory, Config } from 'stryker-api/config';

class BaselineConfigEditor implements ConfigEditory {
    edit(config: Config){ 
      // TODO: Override config.thresholds
    }
}
ConfigEditorFactory.instance().register('baseline', BaselineConfigEditor );
```

riezebosch:
These classes are not yet available with the [stryker-api 0.7.0 release](https://github.com/stryker-mutator/stryker/tree/stryker-api%400.7.0/packages/stryker-api/src/config/ConfigEditor.ts).

nicojs:
True, it is renamed recently to `ConfigEditor` from `ConfigWriter`. I just realized that we haven't done a release yet. Hang on tight

riezebosch:
It's hard to influence from there because at the time of invoking the ConfigEditor/ConfigEditorFactory I guess not to have any actual numbers yet and the provided [ScoreResult is immutable](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker-api/src/report/ScoreResult.ts#L73) and entirely based on coverage.

Issue #320, opened 2017-06-09
Stryker init creates a temp folder


simondel:
Running stryker init creates a temp folder that is not removed:
![image](https://user-images.githubusercontent.com/4221997/26963037-b0ad51bc-4cea-11e7-94ed-6d28464a6566.png)

sharikovvladislav:
Do you want to not to create this folder or do you want to delete this folder after test run finish?

simondel:
Both are fine by me :) The interesting thing to figure out is why the folder is created in the first place.

sharikovvladislav:
Ah so it is not planned. Ok, ty.

sharikovvladislav:
Can you suggest anything so I can debug some npm package? For example I want to debug this problem. I want to somehow install MY OWN COPY of stryker. What are the ways?

I think I can just: `npm i -g /path/to/stryker`. Will this work? Is this best practise?

Any other ways?

nicojs:
> By the way: I started a conversation in your gitter. I get Stryker is not installed on stryker init right after installing stryker locally.

That shouldn't be happening. Does it still occur? If so, we should add a new issue.

> I think I can just: `npm i /path/to/stryker`. Will this work? Is this best practise?

Install can be indeed be done with `npm i file:/path/to/stryker`. You should first build the ts files, but you know that by now 😉 . Since npm5 `npm i file:...` creates a symlink instead of installing it locally like you might expect. To get the local installation behavior i suggest to use (install-local)[https://npmjs.org/package/install-local].

> Running stryker init creates a temp folder that is not removed:

The `init` command should not create a .stryker-tmp folder at all. The reason that it does is because the implementation of [StrykerTempFolder](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker/src/utils/StrykerTempFolder.ts). Whenever that file is loaded, it immediately creates the folder. This is a very bad practice! Importing the file has a side effect (creation of the folder). This is counter intuitive. 

The preferred solution here is to convert all functions in that file to a class. The constructor should than except a string containing the name of the folder to be created (with `'.stryker-tmp'` as the default). In the constructor, it can create the temp folder on disk. At the beginning of the mutation test method in `Stryker`, it should create a new instance of the class. All classes using `StrykerTempFolder` should now use the instance created by `Stryker`.

sharikovvladislav:
> ....Does it still occur?... 

I hope I will be able to check this at the evening.

[...]

Issue #336, opened 2017-07-10
Configurable tmp folder location


viamin:
Currently Stryker creates a new dot folder in the current working directory which is not in a default temp location [https://github.com/stryker-mutator/stryker/blob/8e0e6b016b513a5fbd1f24fc7a08499ef3151dba/packages/stryker/src/utils/StrykerTempFolder.ts#L9](https://github.com/stryker-mutator/stryker/blob/8e0e6b016b513a5fbd1f24fc7a08499ef3151dba/packages/stryker/src/utils/StrykerTempFolder.ts#L9), which means every user of stryker needs to edit their .gitignore file.

It would be nice if this location could be configured in the stryker.conf.js file, and by default live in ./tmp/stryker instead of .stryker-tmp

nicojs:
Thank you @viamin  for reporting this issue.

The position of your .stryker-tmp folder is indeed not ideal. We actually used your system's temp folder earlier, but we ran into a problem: `require` searches for node_modules relative to your current file. For example: `const chai = require('chai');` won't work if you copy your files to your systems temp folder. This is why we ended up with this solution.

In other words: your systems temp folder will only work for browser-based test runners like [karma](http://karma-runner.github.io/).

If we would add a configuration item for this it will be one more way users can make mistakes in configuring Stryker. It's already kind of hard to do. Would you still want this option?

viamin:
I was not referring to system /tmp, but the project root ./tmp, which would be in the same directory as node_modules. I'm proposing a project with these files:
/path/to/project_root/tmp/stryker/ (replacing /path/to/project_root/.stryker-tmp/)
/path/to/project_root/node_modules
/path/to/project_root/package.json
etc.

nicojs:
Ah i see what you mean. That should work indeed. Do you want to take a shot at implementing this? Would you only allow relative path locations? If we allow for the systems tmp folder for example, we should warn people about the node_modules.

sharikovvladislav:
I think this one can be implemented after https://github.com/stryker-mutator/stryker/pull/361 or with it.

avassem85:
Solution created in https://github.com/stryker-mutator/stryker/pull/475

simondel:
We're closing this issue for now. If someone wants to fix this, let us know!

PR #361, opened 2017-08-28
Bugfix: Stryker init creates a temp folder

sharikovvladislav:
Impl of https://github.com/stryker-mutator/stryker/issues/320

TODO: 
- [x] change name of class to TempFolder
- [x] refactor codebase to the new TempFolder class
- [x] refactor unit tests
- [x] perhaps: units


sharikovvladislav:
Is TODO is ok, should I do something else?

nicojs:
Todo looks fine indeed. Thanks for picking this up!

> is it ok that I ask you this questions? :) should I do everything alone maybe? I think it is much faster if I ask. It is much worse when I do something and do it again after you said I did bad.

That's fine, if your ok with waiting up to a day on answers. Gitter can be faster at times.

>do you have this problem? http://i.imgur.com/gowExH0.png. I can't see @types/jasmine. If I add it, everything becomes ok.

We're not using jasmine as a test framework, instead we're using mocha (test framework) and chai (assertion library). It looks like jasmine, but it is not. 

> If I understand correctly there are no units for current StrykerTempFolder module. Should I add tests? What kind of tests do we need?

True and yes i would like unit tests, but not strictly necessary. We can do that some other time.

sharikovvladislav:
Ok, lets continue with questions in gitter.

sharikovvladislav:
Any ideas ? https://gitter.im/stryker-mutator/stryker?at=59a40add66c1c7c477ecba5c 

I put back old name and tests became green. I am not sure this is correct?

nicojs:
indeed, the integration tests of the stryker-html-reporter contain a snapshot of the events stryker reporter callbacks, in order to test the integration of stryker-html-reporter without the need to run stryker each time. So it's loosely coupled.

I see you changed the expect back to it's original name, i think that's fine. There is no technical reason to keep the events in the stryker-html-reporter in sync with current stryker file names. That would be a waste of time

nicojs:
For the unit tests, be sure to mock away all the interactions with the file system, as we want our unit tests to be able to run in complete isolation, without any side effects. That is the difference with integration tests, as there we usually test with side effects. Side effects here is creating files stuff.

In this case, you should stub `mkdirp.sync`, `fs.writeFile` and `deleteDir`.

sharikovvladislav:
Okay! So one more item to do is to create new unit tests for refactored class, right? 

I did all other work? Am I right? Is everything correct? I mean: refactoring of the class to fix the bug.

nicojs:
> I did all other work? Am I right? Is everything correct? I mean: refactoring of the class to fix the bug.

Yes looks good! indeed the only thing remaining is add unit tests.

sharikovvladislav:
Ok, cool. I will research how to do this.

packages/stryker/src/utils/TempFolder.ts
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Creates a random integer number.
+   * @returns A random integer.
+   */
+  random(): number {
+    return Math.ceil(Math.random() * 10000000);
+  }
+
+  private static _instance: TempFolder;
+  static instance() {
```
sharikovvladislav: Now we have constructor with default argument. The problem is we never use it. Current implementation do not allow it. What do you think @nicojs @simondel ?

nicojs:
It might be better to not create the temp folder in the constructor. Instead, we should have an `initialize(folderName: string)` method. This creates the temp folder. This will make it better testable and we are able to implement #336 easier in a later stage. The idea is that the `initialize` function is called once right before the initial test run, from the `Stryker` class. Right now it would always provide `.stryker-tmp` as a default, later we can change it to a config setting. If `initialize` is **not** called, we should not do anything different, just let the code break. 

Making methods dependent on each other like this is a bad practise, but the only way around that would be to not use a Singleton, instead create an instance of a TempFolder and pass that along. You can do that as well if you choose.

The TempFolder is only used during mutation testing. It is the place were we store the sandboxes on disk. A sandbox is a copy of all your files with (possibly) a mutation in one of them. It is where the actual mutation testing occurs. I started a discussion on #366 to see if we can do in-memory testing. If that happens we can delete the `TempFolder` all together as its not needed.

So the `TempFolder` is only used for sandboxes. If a class wants to log something it should create an instance of the `log4js` `Logger` and start using that. Later we can add logging to a log file using `log4js` functionality. For now, don't mind about log files.

sharikovvladislav:
Okay, I updated the code. What do you think? I added initialize method. I will update tests if you think this is ok.

sharikovvladislav:
Any thoughts?

packages/stryker/test/unit/MutatorOrchestratorSpec.ts
```diff
@@ -5,20 +5,23 @@ import MutatorOrchestrator from '../../src/MutatorOrchestrator';
...
   let sut: MutatorOrchestrator;
   let fileUtilsStub: sinon.SinonStub;
   let sandbox: sinon.SinonSandbox;
   let reporter: StrictReporter;
+  let tempFolderMock: Mock<TempFolder>;
 
   beforeEach(() => {
     sandbox = sinon.sandbox.create();
-    sandbox.stub(StrykerTempFolder, 'writeFile');
+    tempFolderMock = mock(TempFolder);
```
nicojs: Looks good! Awesome this works, even with a class with a private constructor.

packages/stryker/src/utils/TempFolder.ts
```diff
@@ -0,0 +1,87 @@
+import * as fs from 'mz/fs';
+import * as path from 'path';
+import * as mkdirp from 'mkdirp';
+import * as log4js from 'log4js';
+import { deleteDir } from './fileUtils';
+
+const log = log4js.getLogger('TempFolder');
+
+export class TempFolder {
```
nicojs: We should add a `private constructor() { }` in order to prevent anyone from creating an instance them selves.

packages/stryker/test/unit/SandboxSpec.ts
```diff
@@ -37,9 +39,14 @@ describe('Sandbox', () => {
       notMutatedFile,
       { path: onlineFile, mutated: false, included: true }
     ];
-    sandbox.stub(StrykerTempFolder, 'createRandomFolder').returns(workingFolder);
-    sandbox.stub(StrykerTempFolder, 'copyFile').returns(Promise.resolve({}));
-    sandbox.stub(StrykerTempFolder, 'writeFile').returns(Promise.resolve({}));
+    tempFolderMock = mock(TempFolder);
+    sandbox.stub(TempFolder, 'instance').returns(tempFolderMock);
+    tempFolderMock.createRandomFolder.returns(workingFolder);
+    tempFolderMock.copyFile.resolves({});
```
nicojs: Looks great! It is indeed better to use `stub.resolves` instead of `stub.returns(Promise.resolve())`

packages/stryker/test/unit/TempFolderSpec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+describe.only('TempFolder', () => {
+  let sandbox: sinon.SinonSandbox;
+
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create();
+  });
+  afterEach(() => sandbox.restore());
+
+  it('TempFolder is presented', () => {
```
nicojs: What exactly are you testing here? This test can never fail, right? If one of the methods is missing, we would get a compile error in our production code.  The benefit of using typescript, is that we don't need tests like these. You can remove it :+1:

In this test, your basically verifying that TypeScript is transpiled correctly.

packages/stryker/test/unit/TempFolderSpec.ts
```diff
@@ -0,0 +1,84 @@
+import { TempFolder } from '../../src/utils/TempFolder';
+
+import * as sinon from 'sinon';
+import { expect } from 'chai';
+import * as mkdirp from 'mkdirp';
+import * as fs from 'mz/fs';
+
+describe.only('TempFolder', () => {
```
nicojs: Remove the `only` here 😺

packages/stryker/test/unit/TempFolderSpec.ts
```diff
@@ -0,0 +1,84 @@
...
+    expect(tempFolderInstance.clean).to.exist;
+    expect(tempFolderInstance.random).to.exist;
+  });
+
+  describe('createRandomFolder', () => {
+    it('should create dir with correct path', () => {
+      const mockCwd = process.cwd() + '/some/dir';
+      sandbox.stub(mkdirp, 'sync');
+      sandbox.stub(process, 'cwd').returns(mockCwd);
+      sandbox.stub(TempFolder.instance(), 'random').returns('rand');
```
nicojs: I think it's better to move this stubbing to the top before each. Putting it right here clutters the tests a bit. You can still do the `cwdStub.returns(mockCwd)` and `randomStub.returns('rand')` here.

packages/stryker/test/unit/TempFolderSpec.ts
```diff
@@ -0,0 +1,84 @@
...
+    });
+    it('should call fs.writeFile', () => {
+      TempFolder.instance().writeFile('filename', 'data');
+      expect(fs.writeFile).to.have.been.calledWith('filename', 'data', {
+        encoding: 'utf8'
+      });
+    });
+  });
+
+  xdescribe('constructor logic', () => {
```
nicojs: Should this still be removed? Or changed into the testing of `initialize()`?

sharikovvladislav:
Uhm. I was asking about is current implementation is ok (not tests). Test are not ready at all. Actually, I am ready to put them into a trash.

packages/stryker/test/unit/TempFolderSpec.ts
sharikovvladislav: This test will fail if somebody accidently remove/rename this class. This test will force him think about it again.

packages/stryker/test/unit/TempFolderSpec.ts
sharikovvladislav: Yes it should be changed. Tests are not ready. I will update them since current code is ok (as I am getting from your review). I just wanted to know if TempFolder.ts code is correct.

packages/stryker/test/unit/TempFolderSpec.ts
sharikovvladislav: Yep, I will. Tests are not ready atm :(

sharikovvladislav:
@nicojs Do you think TempFolder should generate an exception if some method was called, but `initialize()` were not called?

packages/stryker/test/unit/TempFolderSpec.ts
sharikovvladislav: Fixed

packages/stryker/test/unit/TempFolderSpec.ts
sharikovvladislav: I removed it.

packages/stryker/src/utils/TempFolder.ts
sharikovvladislav: Fixed

packages/stryker/test/unit/MutatorOrchestratorSpec.ts
sharikovvladislav: ç

packages/stryker/src/utils/TempFolder.ts
```diff
@@ -0,0 +1,89 @@
...
+      readStream.pipe(writeStream);
+      readStream.on('end', () => resolve());
+    });
+  }
+
+  /**
+   * Deletes the Stryker-temp folder
+   */
+  clean() {
+    log.debug(`Cleaning stryker temp folder ${this.baseTempFolder}`);
```
sharikovvladislav: Is word "clean" in that context"? As I understand "clean" means "clean contents of the directory". But purpose of that function is to remove directory completely.

Are you sure this is correct? Perhaps "remove" will be better since it will say what really happens.

sharikovvladislav:
How do you see unit test for `copyFile`?:

```
  copyFile(fromFilename: string, toFilename: string, instrumenter: NodeJS.ReadWriteStream | null): Promise<void> {
    return new Promise<void>((resolve, reject) => {
      let readStream: NodeJS.ReadableStream = fs.createReadStream(fromFilename, { encoding: 'utf8' });
      let writeStream = fs.createWriteStream(toFilename);
      readStream.on('error', reject);
      writeStream.on('error', reject);
      if (instrumenter) {
        readStream = readStream.pipe(instrumenter);
      }
      readStream.pipe(writeStream);
      readStream.on('end', () => resolve());
    });
  }
```

Do you have any examples of units for same things?

packages/stryker/src/utils/TempFolder.ts
nicojs: Yeah, `remove` would be better. Yes the purpose is to remove the dir completely.

nicojs:
> ❓ @nicojs Do you think TempFolder should generate an exception if some method was called, but initialize() were not called?

Well... maybe that's smart. It communicates clear intent (but you will need extra unit tests).

> How do you see unit test for copyFile?:

You can use the 'test/helpers/streamHelpers.ts'. It has helpers to convert a stream to a string and back. But you can also skip this unit test, as we're not going to use it anymore in the future when we have transpiler support (coming with `stryker-typescript`).

packages/stryker/test/unit/MutatorOrchestratorSpec.ts
nicojs: Ah true. Damn, i think this is a limitation of typescript. It will work if you you change the `mock`:

```typescript

export function mock<T>(constructorFn: { new(...args: any[]): T; }): Mock<T>;
export function mock<T>(constructorFn: any): Mock<T>;
export function mock<T>(constructorFn: any): Mock<T> {
  return sinon.createStubInstance(constructorFn) as Mock<T>;
}
```

And than use it like this: `mock<TempFolder>(TempFolder);`

sharikovvladislav:
> but you will need extra unit tests

How this unit test will look like if one and first initialize call will influence all tests. Perhaps I can reset state to uninitialised manually after each test. I will not initialize it before each test but only for specific.

packages/stryker/src/utils/TempFolder.ts
sharikovvladislav: > Is word "clean" in that context"?

Sorry I wanted to say "is it correct in that context"?

packages/stryker/test/unit/MutatorOrchestratorSpec.ts
sharikovvladislav: Hm. Now I am getting this: https://travis-ci.org/stryker-mutator/stryker/jobs/270886353#L1682

packages/stryker/src/Stryker.ts
```diff
@@ -81,7 +81,8 @@ export default class Stryker {
       this.reporter.onScoreCalculated(score);
       ScoreResultCalculator.determineExitCode(score, this.config.thresholds);
       await this.wrapUpReporter();
-      await StrykerTempFolder.clean();
+      TempFolder.instance().initialize();
```
nicojs: I think this one should be moved up right? Right before the `initialTestRun`?

nicojs:
> Perhaps I can reset state to uninitialized manually after each test. 

> I will not initialize it before each test but only for specific.

That sounds good, but i would do it in a `beforeEach` just to be sure. So one `beforeEach` in a `describe('when temp folder is initialized')` and one in a `describe('when temp folder is not initialized')`

The private constructor seems to work now. I think just moving the call to the initialization

packages/stryker/src/Stryker.ts
sharikovvladislav: Fixed

sharikovvladislav:
I added tests. Here is test run output:

```
    createRandomFolder
      when temp folder is initialized
        ✓ should create dir with correct path
      when temp folder is not initialized
        ✓ should throw error
    clean
      when temp folder is initialized
        ✓ should call deleteDir fileApi
      when temp folder is not initialized
        ✓ should throw error
    writeFile
      ✓ should call fs.writeFile
  6 passing (17ms)
```

- What do you think?
- I did not add initialize call check to writeFile & copyFile since this methods does not work with baseTempFolder. Why they are in TempFolder?
- Did you try prettier? (:

sharikovvladislav:
Also final check is passed: tmp dir is not created on `stryker init` but it is created on `stryker run`.

packages/stryker/test/unit/TempFolderSpec.ts
nicojs: No. That's why we use typescript. Having the test won't hurt of course.

nicojs:
Thanks!

Issue #362, opened 2017-08-28
Clearing require cache breaks grunt-stryker


jmpresso:
**Issue:**
In advance of running grunt-stryker, clear Node's require cache. When grunt-stryker subsequently runs, it will fail. It will be unable to load any of the reporters, even though the node modules are installed.

_Warning: Could not find a reporter with name html, did you install it correctly (for example: npm install --save-dev stryker-html-reporter)? Use --force to continue._

For my particular use case, my build is set up to run grunt-mocha-test twice (with different options for various reasons), with the "clearRequireCache" flag, followed by Stryker. However, it can be reproduced by directly clearing the cache without grunt-mocha-test.

**Reproduction Steps**
Clear the require cache before running Stryker. Ex:

```
    grunt.registerTask('clearRequireCache',
        function() {
            Object.keys(require.cache).forEach(
                function(key) {
                    console.log("deleting " + key);
                    delete require.cache[key]
                }
            )
        }
    );
    
    grunt.registerTask('strykerCacheTest', ['clearRequireCache', 'stryker:mutationTest']);
```

**Issue Analysis**
I tried to debug this a bit, but wasn't able to fully understand why the reporters are not registered.

However, I did notice that I could resolve this in grunt-stryker's stryker.js. If I move the Stryker require from outside the module to inside of grunt.registerMultiTask, it resolves the issue. I suspect clearing the require cache is blowing some initialization away that should take place when Stryker is initially required.

Does this seem to be an acceptable fix, or is my issue just considered more of an edge case that I need to handle on my own (i.e. forking stryker-grunt)?

nicojs:
Is it correct to assume that you `require` grunt stryker somewhere in the file? And have a task for it? 

I think i know what's wrong. Could you please test it by moving `var Stryker = require('stryker').default;` in 'node_modules/grunt-stryker/tasks/stryker.js'  from line 3 to inside the `grunt.registerMultiTask`? I think that should fix your issue.

jmpresso:
Sorry if I wasn't clear- I'm using grunt-stryker. I don't have a require in my own code. I just load the task in my gruntfile-  `grunt.loadNpmTasks('grunt-stryker')`

Yes, if I move that require inside of the grunt.registerMultiTask, it resolves my issue. I was moreso questioning whether this would be an acceptable change to grunt-stryker.

nicojs:
[...]

PR #369, opened 2017-09-02
refactor(stryker task): Move stryker require inside stryker task (#362)

jmpresso:
Only running the Stryker grunt task should cause 'stryker' to be required. Also guards against upstream tasks clearing require cache.


simondel:
Nice find! Thanks for fixing this :)

Issue #371, opened 2017-09-03
Wrong indents in stryker.conf.js file after stryker init command


sharikovvladislav:
Hi!

Right after `stryker init` command executed successfully I get this code in `stryker.conf.js`:

```
svlad at sharikovvlad in ~/sandbox/test-stryker
$ cat stryker.conf.js

  module.exports = function(config){
    config.set(
      {
  "files": [
    {
      "pattern": "src/**/*.js",
      "mutated": true,
      "included": false
    },
    "test/**/*.js"
  ],
  "testRunner": "jest",
  "reporter": [
    "clear-text",
    "progress"
  ],
  "coverageAnalysis": "all"
}
    );
  }
svlad at sharikovvlad in ~/sandbox/test-stryker
$
```

All indents are saved. 

Why don't you use some format tool when creating the config file?

sharikovvladislav:
Is it bug or feature?

If it is a bug can I try to fix this? I think that it will be nice idea to use some format tool like `prettier`. What do you think?

Example of usage is here: https://github.com/prettier/prettier#prettierformatsource--options

sharikovvladislav:
I think return value of `wrapInModule` method in `StrykerConfigWriter` class should be passed through `prettier.format`. What do you think @nicojs @simondel ?

sharikovvladislav:
Here is the result with my fix:

```
svlad at sharikovvlad in ~/sandbox/test-stryker
$ cat stryker.conf.js
module.exports = function(config) {
  config.set({
    files: [
      {
        pattern: "src/**/*.js",
        mutated: true,
        included: false
      },
      "test/**/*.js"
    ],
    testRunner: "jest",
    reporter: ["clear-text", "progress"],
    coverageAnalysis: "all"
  });
};

svlad at sharikovvlad in ~/sandbox/test-stryker
$
```

What do you think?

nicojs:
Looks great. Prettifier doesn't have additional deps. Go for it

sharikovvladislav:
Ye I already have diff... I have problems with my git... 

Just woooooot:

```
svlad at sharikovvlad in ~/dev/stryker on format-stryker-conf-on-init [+!?$]
$ git commit -m 'fix: #371 indents in "stryker.conf.js" file after "stryker init" command'
error: gpg failed to sign the data
fatal: failed to write commit object

svlad at sharikovvlad in ~/dev/stryker on format-stryker-conf-on-init [+!?$]
$
```

[...]

PR #372, opened 2017-09-03
fix: #371 indents in "stryker.conf.js" file after "stryker init" command

sharikovvladislav:
Fix for https://github.com/stryker-mutator/stryker/issues/371.


sharikovvladislav:
Any thoughts why build is red?

For example, one of the tests is failing: 

```
  1) StrykerInitializer initialize() should configure coverageAnalysis: "all" when the user did not select a testFramework:
     AssertionError: expected writeFile to have been called with arguments stryker.conf.js, match(""coverageAnalysis": "all"")
stryker.conf.js
  config.set({
    files: [
      {
        pattern: "src/**/*.js",
        mutated: true,
        included: false
      },
      "test/**/*.js"
    ],
    testRunner: "awesome",
    reporter: ["dimension", "mars"],
    coverageAnalysis: "all"
  });
};
```

Why? Actual calls contains `coverageAnalysis: "all"`. Why error?

sharikovvladislav:
Must be in `"`

sharikovvladislav:
I need only 1 fix. Here is test:

```
    it.only('should configure testFramework, testRunner and reporters', async () => {
      inquirerPrompt.resolves({ testFramework: 'awesome', testRunner: 'awesome', reporters: ['dimension', 'mars', 'progress'] });
      await sut.initialize();
      let expectedReporters = JSON.stringify({
        reporter: [
          'dimension',
          'mars',
          'progress'
        ]
      }, null, 2);
      expectedReporters = expectedReporters.substr(1, expectedReporters.lastIndexOf(']') - 1);
      expect(fs.writeFile).to.have.been.calledWith('stryker.conf.js', sinon.match('testRunner: "awesome"')
        .and(sinon.match('testFramework: "awesome"'))
        .and(sinon.match('coverageAnalysis: "perTest"'))
        .and(sinon.match(expectedReporters)));
    });
```

Here is actuall call argument: 
```
  config.set({
    files: [
      {
        pattern: "src/**/*.js",
        mutated: true,
        included: false
      },
      "test/**/*.js"
    ],
    testRunner: "awesome",
    reporter: ["dimension", "mars", "progress"],
    testFramework: "awesome",
    coverageAnalysis: "perTest"
  });
};
```

How should I update test? I tried replace `"reporter"` with `reporter` but the test did not become red.

simondel:
The dependency that you added seems to be about 5mb in size. Is there any other dependency available?

sharikovvladislav:
Hi, @simondel ! I know only prettier. Probably there maybe another format tools exists, but successfull is only one as I know (probably because of PR by facebook :smirk_cat:).

nicojs:
We could also just use [`objectHelpers.serialize`](https://github.com/stryker-mutator/stryker/blob/93f28ccf952da851830acb0d5ed9d52e659f54e1/packages/stryker/src/utils/objectUtils.ts#L4) to serialize the javascript and then replace newlines with newlines + spaces using [string.replace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace). 

It's a primitive approach, but i think 5MB is a bit steep for the functionality we need. How do projects like `karma` solve this issue?

sharikovvladislav:
Does `karma` have ability to generate config file?

sharikovvladislav:
Oh yeah, I see: https://karma-runner.github.io/1.0/intro/configuration.html

So prettier is bad solution I have to find another, am I right?

nicojs:
> So prettier is bad solution I have to find another, am I right?

Prettier is a good solution if we were to do heavy code generation, but i think 50MB is too steep of a price to pay for some code formatting when decide to generate a single file once every project lifecycle. Do you agree with this?

> We could also just use objectHelpers.serialize to serialize the javascript and then replace newlines with newlines + spaces using string.replace.

`objectHelpers.serialize` uses [`serialize-javascript`](https://github.com/yahoo/serialize-javascript), which underneath uses plain [`JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) with a `replacer` function. Using `JSON.stringify` you can already specify a number of spaces (3rd parameter of the function).

Have fun! 👍

sharikovvladislav:
> 50MB

woot? Yesterday it was:

> 5mb in size

What size is ok?

The current solution is using JSON.stringify and it is not working properly.

sharikovvladislav:
```
svlad at sharikovvlad in ~/sandbox/test-stryker
$ cat stryker.conf.js
"\n      module.exports = function(config){\n        config.set(\n          {\n  \"files\": [\n    {\n      \"pattern\": \"src\u002F**\u002F*.js\",\n      \"mutated\": true,\n      \"included\": false\n    },\n    \"test\u002F**\u002F*.js\"\n  ],\n  \"testRunner\": \"jest\",\n  \"reporter\": [\n    \"clear-text\",\n    \"progress\"\n  ],\n  \"coverageAnalysis\": \"all\"\n}\n        );\n      }"
```

It is result of serializing by objectHeleprs

sharikovvladislav:
Is it a really big problem since we are not in a browser but in a local environment.
Current stryker dir is 27 mb.

simondel:
If I do an npm install of stryker, it's about 17 mb. Perhaps you could ook at [js-beautify](https://www.npmjs.com/package/js-beautify). It's around 2.5mb

sharikovvladislav:
I think I checked the size of node_modules with something else, not only with the stryker. Since that, I got 27 mb. I checked again and now I have 21 mb :) How did you got 17 mb :) ?

nicojs:
Prettier is 1.2 MB when zipped. See this link:

sharikovvladislav:
Hm. All packages included into that archive. Actually, I remembered some issue in prettier repo about it where guys explained why they will do not separate prettier package. But I can't find it :) 

Package could be lesser 2x times :)

simondel:
Good Point about the cache @nicojs! Let's keep using prettier

nicojs:
Thanks @sharikovvladislav !

Issue #367, opened 2017-08-29
Breaking thresholds don't work with grunt-stryker


jmpresso:
**Issue:**
I'm using grunt-stryker 0.7.0, and stryker 0.8.0, with the thresholds configuration. Ex:
```
thresholds: {
            high: 100,
            low: 100,
            break: 100
        }
```

When my score drops below the threshold, it reports the failure, but does not fail the grunt task.

> [2017-08-29 12:04:25.590] [ERROR] ScoreResultCalculator - Final mutation score 99.53 under breaking threshold 100, setting exit code to 1 (failure).
> [2017-08-29 12:04:25.590] [INFO] ScoreResultCalculator - (improve mutation score or set `thresholds.break = null` to prevent this error in the future)
> [2017-08-29 12:04:25.971] [INFO] HtmlReporter - Your report can be found at: file:///C:/Users/up50/projects/vg-ccc-client.npm/reports/mutation/html/index.html
> [2017-08-29 12:04:26.070] [INFO] Stryker - Done in 11 seconds.
> 
> Done, without errors.
> 
> Process finished with exit code 0

You can see the process exit code is 0 (successful), when it should've failed.

**Analysis**
When the mutation score is under 'break', it does invoke [setExitCode ](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker/src/utils/objectUtils.ts#L72)in objectUtils, setting process.exitCode to 1.

In grunt-stryker, I verified it receives the process.exitCode as 1. However, the process still exits with an exitCode of 0. I suspect this has something to do with the [this.async](https://gruntjs.com/api/inside-tasks) call in [stryker.js](https://github.com/stryker-mutator/stryker/blob/master/packages/grunt-stryker/tasks/stryker.js#L32). . If I remove this 'done' call, it appears to work properly. When a break occurs the grunt process will exit with a code of 1, otherwise it will exit with a normal code of 0.

this.async must not respect an exitCode. Perhaps the error code needs to be checked, and a false passed to done if its >0? That's assuming this.async is really needed here with a promise. When I removed the line, it still appeared to wait and end appropriately.

simondel:
This has been fixed with `grunt-stryker@0.8.2` Thanks for fixing it @jmpresso !

PR #374, opened 2017-09-03
fix(stryker task): Fail grunt task if exit code from runMutationTest greater than zero (#367) 

jmpresso:


integrationTest/test/grunt-stryker-test/stryker.conf.js
```diff
@@ -1,8 +1,8 @@
 module.exports = function (config) {
   config.set({
     files: [
-      {pattern: 'testResources/sampleProject/src/**/!(InfiniteAdd).js', mutated: true, included: true },
```
nicojs: Did the grunt integration test just never work? And we never saw it because grunt didn't care about exit code?

jmpresso: There are 2 grunt integration tests, one with the config file and one without. The one that uses this config file never ran any tests, which I don't believe was intentional. It didn't fail, just ended without running any tests since the path was wrong. I fixed the path since I was using a version of this to test my change- it was really unrelated to my change.

I was initially going to add an integration test for failing the task with a break threshold, but couldn't figure out a simple way to add it and pass the build.

PR #380, opened 2017-09-19
Refactor(SampleProject): Remove sampleProject

Archcry:
Removing sampleProject as it is no longer used


simondel:
Better Code Hub complains that we now have pretty much all code in the packages folder. Instead of some code in the packages folder and some in the sampleProject folder.

nicojs:
Thanks @Archcry !

PR #393, opened 2017-09-22
feat(typescript): Force no code quality compiler options

### LINKED ISSUE #391: TypeScript: Override some quality options
When transpiling typescript code, we should override *some* of the compiler options designed for code quality. For example: if we now mutate an if statement to `if(false)` and the user has the code quality option `allowUnreachableCode=false` the mutant will not be tested and is treated as a false positive (basically). 

I don't think this is what we want. If you think about it: indeed, nobody in their right mind would mutate an if statement like that and typescript is right to report a warning when done so, but at the same time, that is *not* what we want to test when we mutate the if-statement. We want tot see if your tests can detect the side effect of *never* going in the if statement.

See https://www.typescriptlang.org/docs/handbook/compiler-options.html for a list of compiler options. I think we should override the following:

* `allowUnreachableCode` should be true 
* `noUnusedLocals` should be false 
* `noUnusedParameters` should be false

Another way to look at it: we're mutating code, of course the code quality will be less. That doesn't mean that the mutant is invalid.

nicojs:
Fixes #391

### DIFF
--- a/packages/stryker-typescript/README.md
+++ b/packages/stryker-typescript/README.md
@@ -81,6 +81,14 @@ Enable the config editor by pointing the `tsconfigFile` property to your tsconfi
 }
 ``` 
 
+We always override some properties to enforce these rules:
+```js
+allowUnreachableCode: true
+noUnusedLocals: false
+noUnusedParameters: false
+```
+
+
 ### TypescriptMutator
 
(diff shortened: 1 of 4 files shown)


simondel:
This can be merged if the build is green and if you agree with my documentation changes 👍

nicojs:
👍 looks good

PR #404, opened 2017-10-02
fix(ConfigReader): Use CLI options with default config file

### LINKED ISSUE #390: Make stryker config file mandatory
Running stryker with command line options won't read a stryker.conf.js file. This results in unexpacted behavior, for example:

```bash
$ stryker run
[2017-09-22 15:30:17.180] [INFO] ConfigReader - Using stryker.conf.js in the current working directory.
...
[2017-09-22 15:30:29.083] [INFO] Stryker - Done in 11 seconds.

$ stryker run --logLevel debug
[2017-09-22 15:30:35.483] [INFO] ConfigReader - No config file specified. Running with command line arguments
```

I think it would be more useful to always require a stryker.conf.js file to be specified.

Be-ngt-oH:
Change ConfigReader behavior to always use stryker.conf.js if present. CLI
options override options from the file.

Run does not fail anymore if a CLI option, but no config file is specified.
closes #390

### DIFF
--- a/packages/stryker/src/ConfigReader.ts
+++ b/packages/stryker/src/ConfigReader.ts
@@ -1,5 +1,6 @@
 import { Config } from 'stryker-api/config';
 import { StrykerOptions } from 'stryker-api/core';
+import * as fs from 'mz/fs';
 import * as log4js from 'log4js';
 import * as _ from 'lodash';
 
@@ -10,6 +11,8 @@ export const CONFIG_SYNTAX_HELP = '  module.exports = function(config) {\n' +
   '      // your config\n' +
   '    });\n' +
   '  };';
+
+const DEFAULT_CONFIG_FILE = 'stryker.conf.js';
 
 const log = log4js.getLogger('ConfigReader');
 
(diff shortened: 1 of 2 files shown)


packages/stryker/src/ConfigReader.ts
```diff
@@ -34,10 +37,21 @@ export default class ConfigReader {
   }
 
   private loadConfigModule(): Function {
-    // we start with a dummy configModule
+    // Dummy module to be returned if no config file is loaded.
     let configModule: Function = function () { };
+
+    if (!this.cliOptions.configFile) {
+      try {
+        fs.accessSync(`${process.cwd()}/${DEFAULT_CONFIG_FILE}`);
```
nicojs: I think it's cleaner to use ```path.resolve(`./${DEFAULT_CONFIG_FILE}`)``` here.

packages/stryker/src/ConfigReader.ts
```diff
@@ -34,10 +37,21 @@ export default class ConfigReader {
...
+        log.info(`Using ${DEFAULT_CONFIG_FILE} in the current working directory.`);
+        this.cliOptions.configFile = DEFAULT_CONFIG_FILE;
+      } catch (e) {
+        log.info('No config file specified. Running with command line arguments.');
```
nicojs: Could we also add a message: ```'Use `stryker init` command to generate your config file'``` here? So we help the end user even more?

simondel:
Awesome! You got #404 🥇

nicojs:
Thanks @Be-ngt-oH . Awesome job 🥇

PR #405, opened 2017-10-03
Refactored the code by adding StrykerCli class with run method, Updat…

bicep:
1) Created StrykerCli class with run method
2) Edited binary file by instantiating class
3) Deleted stryker-cli files

### DIFF
--- a/packages/stryker/src/reporters/ProgressAppendOnlyReporter.ts
+++ b/packages/stryker/src/reporters/ProgressAppendOnlyReporter.ts
@@ -18,22 +18,17 @@ export default class ProgressAppendOnlyReporter extends ProgressKeeper {
   }
 
   private render() {
-    process.stdout.write(`Mutation testing ${this.procent()} (ETC ${this.etc()}) ` +
-      `[${this.progress.killed} ${this.progress.killedLabel}] ` +
-      `[${this.progress.survived} ${this.progress.survivedLabel}] ` +
-      `[${this.progress.noCoverage} ${this.progress.noCoverageLabel}] ` +
-      `[${this.progress.timeout} ${this.progress.timeoutLabel}] ` +
-      `[${this.progress.runtimeError} ${this.progress.runtimeErrorLabel}] ` +
-      `[${this.progress.transpileError} ${this.progress.transpileErrorLabel}]` +
+    process.stdout.write(`Mutation testing ${this.percent()} (ETC ${this.etc()}) ` +
+      `${this.progress.tested}/${this.progress.total} tested (${this.progress.survived} survived)` +
       os.EOL);
   }
 
-  private procent() {
-    return Math.floor(this.progress.testedCount / this.progress.totalCount * 100) + '%';
+  private percent() {
+    return Math.floor(this.progress.tested / this.progress.total * 100) + '%';
   }
 
   private etc() {
-    const etcSeconds = Math.floor(this.timer.elapsedSeconds() / this.progress.testedCount * (this.progress.totalCount - this.progress.testedCount));
+    const etcSeconds = Math.floor(this.timer.elapsedSeconds() / this.progress.tested * (this.progress.total - this.progress.tested));
     if (isFinite(etcSeconds)) {
       return etcSeconds + 's';
     } else {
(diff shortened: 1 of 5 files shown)


packages/stryker/src/StrykerCli.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const log = log4js.getLogger('stryker-cli');
+let command: string = '', strykerConfig: string | null = null;
+
+export default class StrikerCli {
+
+  private argv: string[];
+
+  constructor(argv: string[]) {
+    this.argv = argv;
```
simondel: One of the great things about TypeScript is that you don't need this kind of useless code that you see in many languages. You can change the constructor argument to `private argv: string[]` and it will set the private variable `argv` to the argument so you can remove this line, and the declaration of `private argv: string[];` above the constructor 🎉

packages/stryker/src/StrykerCli.ts
```diff
@@ -0,0 +1,93 @@
+import * as program from 'commander';
+import { CONFIG_SYNTAX_HELP } from './ConfigReader';
+import Stryker from './Stryker';
+import StrykerInitializer from './initializer/StrykerInitializer';
+import * as log4js from 'log4js';
+
+const log = log4js.getLogger('stryker-cli');
+let command: string = '', strykerConfig: string | null = null;
```
simondel: We can move these two variables inside of the class so they become properties :)

packages/stryker/bin/stryker
```diff
@@ -5,4 +5,6 @@
 process.title = 'stryker';
 
 // Run the TypeScript compiled stryker-cli.js
-require('../src/stryker-cli');
\ No newline at end of file
+//require('../src/stryker-cli');
+const StrykerCli = require('../src/StrykerCli');
+new StrykerCli(process.argv).run()
```
simondel: You're missing a semicolon here

packages/stryker/bin/stryker
```diff
@@ -5,4 +5,6 @@
 process.title = 'stryker';
 
 // Run the TypeScript compiled stryker-cli.js
-require('../src/stryker-cli');
\ No newline at end of file
+//require('../src/stryker-cli');
+const StrykerCli = require('../src/StrykerCli');
```
simondel: The constructor function that you export in the StrykerCli file is added to the property `default` so I believe this line should be `const StrykerCli = require('../src/StrykerCli').default;`

bicep:
@simondel  Thank you so much for your lengthy reply and guidance. I'll work on them right away and ask questions if I have any!

simondel:
Thanks for the update! I noticed that there are still some package.lock.json files, you can see them [here](https://github.com/stryker-mutator/stryker/pull/405/files).
To answer your questions:
1. It turns out it's not in our gitignore... I thought it was.
2. @nicojs and I personally use Visual Studio Code but you are free to use anything you'd like of course. The best way to test the code is to compile it and then run the tests, so `npm run build` and then `npm test`
3. Using `npm run build` if you want to run tslint as well. If you want to compile in the background and not run tslint (which causes the build to fail right now), you could run `npm start`.

bicep:
Thanks for your reply! Okay, so I have added the package-lock.json files to gitignore. They should be removed now. When I ran `git rm -r --cached .` and restaged the files with `git add .`, I think I ended up also unstaging files like these:

	deleted:    packages/stryker/testResources/globTestFiles/sample/src/Add.js
	deleted:    packages/stryker/testResources/globTestFiles/sample/src/Array.js
	deleted:    packages/stryker/testResources/globTestFiles/sample/src/Circle.js
	deleted:    packages/stryker/testResources/globTestFiles/sample/src/Error.js

Also am I supposed to do these two things every time before I push a new commit to this branch?
2) fetch+merge local master with the main stryker repo through upstream
3) rebase my branch

I have been doing that per this link (https://gist.github.com/Chaser324/ce0505fbed06b947d962) but this time I was not allowed to rebase. It warned that all the package-lock.json files which are untracked could be replaced.

simondel:
Hi! It seems like you accidentally deleted too much. You could remove your last commit using
```
git reset HEAD~1
git push --force
```
Then I would suggest manually deleting the package.lock.json files and adding it to the .gitignore. We squash all pull requests so they won't be in our version history :)

bicep:
I'm not sure what's going on with the git commands. Why would I push my code to the repo `git push --force` when I am trying to revert back to the last commit?

bicep:
Gah! Something happened as I was fiddling with git and I think I lost all the files..

PR #410, opened 2017-10-07
 test(log4js mock): Prevent leaking state 

nicojs:


packages/stryker/test/unit/initializer/StrykerInitializerSpec.ts
```diff
@@ -69,6 +73,7 @@ describe('StrykerInitializer', () => {
     });
 
     it('should prompt for test runner, test framework, mutator, transpilers and reporters', async () => {
+      console.error('RUNNING should prompt for test runner, test framework, mutator, transpilers and reporters');
```
simondel: Why did you add this statement?

nicojs:
Because i was debugging and it was late. Good catch!

simondel:
Awesome thanks!

PR #411, opened 2017-10-07
fix(BetterCodeHub): Use component-depth 2

simondel:
This should bring us up to a score of 10/10 again :)


simondel:
Ok this doens't work yet..

PR #412, opened 2017-10-08
fix(BetterCodeHub): Move bettercodehub.yml to the root

simondel:


simondel:
@nicojs Turns out I placed the yml file in the tasks folder by accident..

PR #415, opened 2017-10-10
refactor(coverage): Use transpiler pipeline

nicojs:
Move code coverage intrumentation to a transpiler so we can handle the code transformation as all other transpilers. Simplifies the process by eliminating the need for keeping a seperate transformation process. Also gets rid of the streaming api, so no need to know that part about node internals anymore.


nicojs:
I've tested it together with "fix(mocha framework): Select tests based on name" and perTest report still works as it should.

PR #416, opened 2017-10-11
fix(html-reporter): Remove warning and fix the example image

simondel:
I haven't updated the report at https://stryker-mutator.github.io/stryker-html-reporter yet so that still shows the html reporter pre 0.9.0.


simondel:
I renamed this pr to a fix so we can actually release it. I believe that docs don't trigger patch releases.

nicojs:
Great thinking!

PR #426, opened 2017-10-18
Refactor cli class

bicep:
I actually really messed up the old PR, so I'm submitting a new one. It should have the refactored code and the package-lock.json files should not be there any more!


simondel:
Thanks for fixing it in a new pull request! I have merged it 🎉

PR #428, opened 2017-10-20
test(html reporter): Fix bootstrap dependency

nicojs:


nicojs:
This should fix the build

PR #431, opened 2017-10-23
fix(html report): Add consistency in numbering of mutants

### LINKED ISSUE #422: html report numbering inconsistent
http://grave-digger-snake-27711.bitballoon.com/printer.ts.html

The numbers in the list on the right side do not correspond with the labels in the source code.

nicojs:
The recently added "mutant table" in the html had inconsistent numbering in a lot of cases when compared to the id's in the report. I.e. mutant 25 in the report would not allign with mutant 25 in the mutant table.

fixes #422


packages/stryker-html-reporter/package.json
```diff
@@ -49,10 +49,12 @@
   },
   "devDependencies": {
     "@types/file-url": "^1.0.28",
+    "@types/jsdom": "^11.0.2",
```
simondel: What do we do with types that are only used in one package? Do we add them to the package or to the root package.json for lerna?

packages/stryker-html-reporter/src/templates/sourceFile.tsx
```diff
@@ -7,7 +7,15 @@ import { ScoreResult, SourceFile, MutantResult, MutantStatus } from 'stryker-api
 import Breadcrumb from '../Breadcrumb';
 import { mutantTable } from './mutantTable';
 
+export interface NumberedMutant {
+    index: number;
```
simondel: Do we need to save the index or can we generate the index when we loop over it?

packages/stryker-html-reporter/src/templates/sourceFile.tsx
```diff
@@ -7,7 +7,15 @@ import { ScoreResult, SourceFile, MutantResult, MutantStatus } from 'stryker-api
...
+    mutant: MutantResult;
+}
+
 export function sourceFile(result: ScoreResult, sourceFile: SourceFile | undefined, mutants: MutantResult[], breadcrumb: Breadcrumb, thresholds: MutationScoreThresholds) {
+    const numberedMutants = _
+        .sortBy(mutants, m => m.range[0] * 10000 + m.range[1] * -1)
```
simondel: What? Why not just order by the start range?

nicojs: Lol, i think it is a left over from when we did ordering based on position (line + column). I'll replace it.

packages/stryker-html-reporter/src/templates/sourceFile.tsx
nicojs: We don't need it indeed. But it does make it more explicit. Do you want me to remove it?

packages/stryker-html-reporter/package.json
nicojs: These types cannot be added to the root package.json, as it would force `lib: ['dom']` for all packages. 

The most important thing is to keep the types where we also keep de (dev) dependencies. For example: i had to add `'@types/jsdom'` here, so i also added `'jsdom'` in this same package.

packages/stryker-html-reporter/src/templates/sourceFile.tsx
simondel: Having an index is fine, it's just that this adds another interface.

simondel:
How do we do the sorting in the component that shows the actual code? (at the left of the page) If we change the order of the numberedmutants, will that component change as well?

nicojs:
Yew it will. That part is also in this file: ...ter/src/templates/sourceFile.tsx

packages/stryker-html-reporter/src/templates/sourceFile.tsx
nicojs: True. Maybe we should add mutant id to the `MutantResult` interface at a later stage, so all reporters report on the same mutants with one Id

PR #433, opened 2017-10-23
Async transpiler plugin support

Archcry:
In order to add webpack support in Stryker, Stryker has to support async transpiler plugins.

This pull-request will:
- Add support for asynchronous transpilers in Stryker.
- Change Transpiler interface accordingly.
- Change Typescript transpiler plugin accordingly.
- Fix all tests that broke because of the changes that were made.


packages/stryker/src/transpiler/CoverageInstrumenterTranspiler.ts
```diff
@@ -24,7 +24,7 @@ export default class CoverageInstrumenterTranspiler implements Transpiler {
     this.log = getLogger(CoverageInstrumenterTranspiler.name);
   }
 
-  transpile(files: File[]): TranspileResult {
+  public async transpile(files: File[]): Promise<TranspileResult> {
```
nicojs: i think `async` is not needed here. It is only used to mark methods in which you can use `await`

packages/stryker/src/transpiler/MutantTranspiler.ts
```diff
@@ -74,7 +74,7 @@ export default class MutantTranspiler {
     }
   }
 
-  private transpileMutant(mutant: TestableMutant): Promise<TranspileResult> {
+  private async transpileMutant(mutant: TestableMutant): Promise<TranspileResult> {
```
nicojs: Not needed here as well

packages/stryker/src/transpiler/TranspilerFacade.ts
```diff
@@ -17,11 +17,11 @@ export default class TranspilerFacade implements Transpiler {
     }
   }
 
-  transpile(files: File[]): TranspileResult {
+  public async transpile(files: File[]): Promise<TranspileResult> {
```
nicojs: Not needed here as well

packages/stryker/src/transpiler/CoverageInstrumenterTranspiler.ts
```diff
@@ -30,9 +30,9 @@ export default class CoverageInstrumenterTranspiler implements Transpiler {
         outputFiles: files.map(file => this.instrumentFileIfNeeded(file)),
         error: null
       };
-      return this.addCollectCoverageFileIfNeeded(result);
+      return Promise.resolve(this.addCollectCoverageFileIfNeeded(result));
     } catch (error) {
-      return this.errorResult(errorToString(error));
+      return Promise.resolve(this.errorResult(errorToString(error)));
     }
```
Archcry: We might want to leave the function `async` because otherwise the two return statements have to be a `Promise`.

nicojs: I think this is fine as it is. It looks clean in the resulting JavaScript.

nicojs:
@simondel it can be merged IMO. What commit message should we use? Strictly speaking its a breaking change, but we could also state that the transpiler api is still experimental (no one is creating stryker transpilers at the moment anyway)

Archcry:
:+1:

simondel:
@Archcry Nice job!
@nicojs I'd go with a feat for now due to the fact that it's still experimental

PR #442, opened 2017-10-29
fix missing commas in default config

vird:


simondel:
Awesome! Thanks for fixing it 👍

PR #445, opened 2017-11-02
chore(ts 2.6): Support TypeScript 2.6

nicojs:
* Remove unused variables
* Alter types for co- and contra variance

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "sinon-chai": "^2.10.0",
     "source-map-support": "^0.4.16",
     "tslint": "^5.6.0",
-    "typescript": "^2.2.2"
+    "typescript": "^2.6.1"
   },
   "scripts": {
     "clean": "lerna run prebuild",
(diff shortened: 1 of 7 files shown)


simondel:
Awesome! Why did you have to specify the datatype in the HTML reporter but not in the function above? Vscode seems to know that type implicitly

nicojs:
> Why did you have to specify the datatype in the HTML reporter but not in the function above?

Do you mean this method?

```typescript
  async rows() {
    const rows: ElementFinder[] = await this.host.$$('tbody tr');
    return rows.map(row => new ResultTableRow(row));
  }
```

The reason that `rows` need to be typed explicitly is the return type of `await this.host.$$('tbody tr');` is of type `any[] | ElementFinder[]`. Maybe the typings could have been defined better, but being what they are now they provide a compiler error because of the awesome addition of `--strictFunctions` (enabled by default, because we're running in `--strict` mode).

simondel:
I was talking about this line:
```ts
  head() {
    return this.host.$$('thead th');
  }
```

nicojs:
> Does that not cause issues because we don't await it?

Why should it? It's not used anywhere  :sweat_smile:

PR #448, opened 2017-11-03
feat(karma): Ability to override options

### LINKED ISSUE #434: Not possible to exclude files that are included in a karma config file
When specifying the `karmaConfigFile` config parameter, it would come in use to have the possibility to include or exclude files that are specified in the karma configuration. This should be possible with the `karmaConfig: { files: [] }` config parameter.

The documentation currently states "The following Karma config options cannot be overridden: files, ..." 

The advantage of being able to override the files config parameter is that it prevents the need to copy the entire karma config when the list of files in karma and Stryker differ.

The following Stryker configuration is an example of a Stryker configuration file which overrides the list specified by the karma configuration

```
  config.set({
    files: ['!**/*.*', 'src/app/**/*.ts','src/app/*.ts'],
    testRunner: "karma",
    mutator: "typescript",
    transpilers: ["typescript"],
    reporter: ["clear-text", "progress"],
    testFramework: "jasmine",
    coverageAnalysis: "off",
    karmaConfigFile: "karma.conf.js",
    karmaConfig: { files: ['!e2e/**/*.ts'] },
    tsconfigFile: 'tsconfig.json',
    mutate: ["src/**/*.ts"]
  });
};

### PR DESCRIPTION
**Pull request by nicojs**
Add the ability to override options.

Change 2 things:

1. Files from karma config are now imported *before the files in the stryker.conf.js*. This give the user the ability to reset the `files` by adding `'!**/*.*` (or something)
2. Options from `karmaConfig` will take precedence over options in the `karmaConfigFile`. This way a user can override options easily. 

These options:

```javascript
// stryker.conf.js
{
  karmaConfig: {
    files: ['first file'],
    some: 'override'
  },
  karmaConfigFile: 'karma.conf.js',
  files: ['bar']
}

// karma.conf.js
{ 
  some: 'this file is overridden',
  optionFromKarmaConfigFile: 'will remain if not overridden'
}  
```

Will result in:

```javascript
{
  karmaConfig: {
    some: 'override',
    optionFromKarmaConfigFile: 'will remain if not overridden'
  },
  karmaConfigFile: 'karma.conf.js',
  files: ['first file', 'bar']
}
```

Fixes #434


packages/stryker-karma-runner/src/KarmaConfigEditor.ts
```diff
@@ -3,48 +3,43 @@ import * as karma from 'karma';
 import { InputFileDescriptor } from 'stryker-api/core';
 import { ConfigEditor, Config as StrykerConfig } from 'stryker-api/config';
 import KarmaConfigReader from './KarmaConfigReader';
+import { KARMA_CONFIG, KARMA_CONFIG_FILE } from './configKeys';
 
 const log = log4js.getLogger('KarmaConfigEditor');
```
simondel: We still set the logger outside of the class here

packages/stryker-karma-runner/src/KarmaConfigEditor.ts
```diff
@@ -3,48 +3,43 @@ import * as karma from 'karma';
...
-  private static importFiles(strykerConfig: StrykerConfig, karmaConfig: karma.ConfigOptions) {
+  private static importFiles(strykerConfig: StrykerConfig) {
+    const karmaConfig: karma.ConfigOptions = strykerConfig[KARMA_CONFIG];
     if (!strykerConfig.files) { strykerConfig.files = []; }
-    if (!karmaConfig.files) { karmaConfig.files = []; }
-    if (!karmaConfig.exclude) { karmaConfig.exclude = []; }
+    if (!Array.isArray(karmaConfig.files)) { karmaConfig.files = []; }
+    if (!Array.isArray(karmaConfig.exclude)) { karmaConfig.exclude = []; }
 
     const files: (karma.FilePattern | string)[] = karmaConfig.files;
```
simondel: Do we still need this assignment?

packages/stryker-karma-runner/src/KarmaConfigEditor.ts
```diff
@@ -3,48 +3,43 @@ import * as karma from 'karma';
...
     const exclude: string[] = karmaConfig.exclude;
```
simondel: Do we still need this assignment?

packages/stryker-karma-runner/src/KarmaConfigEditor.ts
nicojs: Yes, we don't no for sure it's an array and we expect arrays for the remainder of this function

packages/stryker-karma-runner/src/KarmaConfigEditor.ts
nicojs: Yes, let's fix that as well in this PR

simondel: You fixed it in a different file 👍

PR #452, opened 2017-11-05
fix(InputFileResolver): Presume .zip and .tar are binary files.

### LINKED ISSUE #447: Extend binary file extentions list
Hi,

Earlier Me and Nico where trying to use Stryker on an Typescript project. We noticed we had some failing tests. After some debugging we found out our .zip files were correctly interpreted as binary files.

Some additional file extensions would be nice.
`tar`, `zip`

(Sorry for the unclear explanation)

Greets

nicojs:
Fixes #447

### DIFF
--- a/packages/stryker/test/unit/InputFileResolverSpec.ts
+++ b/packages/stryker/test/unit/InputFileResolverSpec.ts
@@ -10,7 +10,7 @@ import * as fs from 'mz/fs';
 import { FileDescriptor, TextFile } from 'stryker-api/core';
 import currentLogMock from '../helpers/log4jsMock';
 import BroadcastReporter from '../../src/reporters/BroadcastReporter';
-import { Mock, mock, textFile } from '../helpers/producers';
+import { Mock, mock, textFile, binaryFile } from '../helpers/producers';
 
 const files = (...namesWithContent: [string, string][]): TextFile[] =>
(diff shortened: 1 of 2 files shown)


packages/stryker/test/unit/InputFileResolverSpec.ts
```diff
@@ -244,4 +246,22 @@ describe('InputFileResolver', () => {
...
+    sut = new InputFileResolver([], binaryFiles, reporter);
+
+    // Act
+    const actual = await sut.resolve();
+
+    // Assert
+    expect(actual).deep.eq(binaryFiles.map(file => binaryFile({ 
+      name: path.resolve(file), 
+      mutated: false, 
+      transpiled: true
```
simondel: Why should `transpiled` be true?

nicojs: Because that is the default when no `transpiled` flag is given. At the same time the default for `binaryFile` is false.

simondel: > At the same time the default for binaryFile is false.
But you just expect that it's true here.

nicojs: Yes it should be true. The `InputFileResolver` should set it to true by default, right?

PR #455, opened 2017-11-13
feat(mocha 4): Add support for mocha version 4

nicojs:


nicojs:
Nothing major has changed, only an exit argument that we need for karma tests (karma itself has no clean way of closing).

PR #458, opened 2017-11-14
chore(stryker.conf.js): Don't transpile package.json

nicojs:


nicojs:
Since v2.6 of typescript, `.json` is added as a valid extension for typescript files:

```javascript
        Extension["Ts"] = ".ts";
        Extension["Tsx"] = ".tsx";
        Extension["Dts"] = ".d.ts";
        Extension["Js"] = ".js";
        Extension["Jsx"] = ".jsx";
        Extension["Json"] = ".json";
```

However, we automatically also ask for diagnostic messages for any typescript file. Resulting in this error: 

```
[2017-11-14 07:16:28.363] [ERROR] StrykerCli - an error occurred Error: Could not find file: 'C:\z\github\stryker-mutator\stryker\packages\stryker\package.json'.
    at getValidSourceFile (C:\z\github\stryker-mutator\stryker\node_modules\typescript\lib\typescript.js:95356:23)
    at Object.getSemanticDiagnostics (C:\z\github\stryker-mutator\stryker\node_modules\typescript\lib\typescript.js:95538:36)
    at C:\z\github\stryker-mutator\stryker\packages\stryker-typescript\src\transpiler\TranspilingLanguageService.js:56:92
```

Is there an other way of fixing these issues?

simondel:
Why do we even have it in our stryker.conf.js?

simondel:
I don't think we have to add json files to some filter right now. If other people start having this issue I think we should consider it.

PR #459, opened 2017-11-14
fix(init stryker typescript): Add init section

### LINKED ISSUE #454: Stryker-typescript default config
in general, i found the stryker-cli auto-generated project for a typescript project harder to follow than writing the file by hand.
specifically:
* the files section was just completely wrong. the autogenerated segments had globs for JS files. also the weirdness of needing to negate out the source files then re-add them was strange, but seems to be a standard part of configuring a TS project?
* no default tsconfigFile option, which led to lots of weird messages when stryker tried to transpile the source. I feel like a default of tsconfig.json is probably a pretty safe default and will highlight to new users that the option exists

Originally reported by @mscharley on Gitter

This can be easily added. Let's do so

nicojs:
* Add a `initStrykerConfig` section to the stryker-typescript package.json file. 
* Remove `mutate` section from `stryker-karma-runner`. A test runner shouldn't force mutation of certain files.

Closes #454

### DIFF
--- a/packages/stryker-typescript/package.json
+++ b/packages/stryker-typescript/package.json
@@ -56,9 +56,6 @@
     "tsconfigFile": "tsconfig.json",
     "files": [],
     "coverageAnalysis": "off",
-    "transpilers": [
-      "typescript"
-    ],
     "mutate": [
       "src/**/*.ts"
     ]
--- a/packages/stryker-typescript/package.json
+++ b/packages/stryker-typescript/package.json
@@ -56,7 +56,6 @@
     "tsconfigFile": "tsconfig.json",
     "files": [],
     "coverageAnalysis": "off",
-    "mutator": "typescript",
     "transpilers": [
       "typescript"
     ],
--- a/packages/stryker-typescript/package.json
+++ b/packages/stryker-typescript/package.json
@@ -51,5 +51,17 @@
   "peerDependencies": {
     "stryker-api": "^0.11.0",
     "typescript": "^2.5.0"
+  },
+  "initStrykerConfig": {
+    "tsconfigFile": "tsconfig.json",
+    "files": [],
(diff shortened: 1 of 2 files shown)


packages/stryker-typescript/package.json
```diff
@@ -51,5 +51,17 @@
   "peerDependencies": {
     "stryker-api": "^0.11.0",
     "typescript": "^2.5.0"
+  },
+  "initStrykerConfig": {
+    "tsconfigFile": "tsconfig.json",
+    "files": [],
+    "coverageAnalysis": "off",
+    "mutator": "typescript",
```
simondel: Why set this? Shouldn't this be set by default?

packages/stryker-typescript/package.json
```diff
@@ -51,5 +51,17 @@
...
+    "transpilers": [
+      "typescript"
```
simondel: Overwriting this can cause issues when we start supporting webpack

nicojs: Yeah, i know. What else can we do? Maybe we should have slit it up into more packages? Or have an other way of opting in-out of initial config somehow?

packages/stryker-typescript/package.json
nicojs: Ah yes, you are right. I'll remove it.

packages/stryker-typescript/package.json
simondel: We could remove it here and rely on the initialiser to set the correct config for the transpilers array

nicojs: wow, is that also set by the initializer? Sorry, i didn't realize 👍

nicojs:
I think all remarks are addressed. Sorry for the obvious mistakes.

simondel:
Awesome 👍

PR #460, opened 2017-11-14
feat(typescript-mutator): Implement new ArrowFunctionMutator

### LINKED ISSUE #456: The block remover TypeScript mutator doesn't remove lambda function bodies
```
export const isNumber: TypeGuard<number> = (n): n is number =>
  typeof n === "number";
```

Currently the only thing mutated about the above function is the equality. Considering there is a tslint rule to force all functions to be defined this way, which I use, this is a reasonably serious bug.

mscharley:
The current implementation of BlockMutator doesn't catch anonymous
functions that use the shorthard return value syntax such as:

```
const lengthOf = (s: string) => s.length
```

Closes #456


simondel:
Awesome thanks!! 🥇

nicojs:
@mscharley thanks for implementing this so cleanly (and fast!). If you want, you can also add the string mutator we've discussed earlier  :wink:.

PR #461, opened 2017-11-15
feat(typescript-mutation): Add a StringLiteral mutator.

mscharley:
This mutator will empty out string values. Some of them will cause
compilation errors, but others are useful. It ensures error messages and
other such messages are tested.

### DIFF
--- a/packages/stryker-typescript/src/mutator/StringLiteralMutator.ts
+++ b/packages/stryker-typescript/src/mutator/StringLiteralMutator.ts
@@ -20,7 +20,20 @@ export default class StringLiteralMutator extends NodeMutator<AllStringLiterals>
   }
 
   protected identifyReplacements(str: AllStringLiterals, sourceFile: ts.SourceFile): NodeReplacement[] {
-    return [{ node: str, replacement: '""' }];
+    if (
+      // Check for empty strings first.
+      (str.kind === ts.SyntaxKind.StringLiteral && str.text === '')
+      // Only check for the Token form of template literals. It's impossible to have a TemplateExpression
+      // that is empty as it's only used when there is a value embedded. I haven't found a case where the
+      // cast fails but the worst that can happen is it is undefined and we fall through to the wrong statement.
+      || (str.kind === ts.SyntaxKind.FirstTemplateToken && (str as ts.Node & { text: string }).text === '')
+    ) {
+      console.log(str);
+      return [{ node: str, replacement: '"Stryker was here!"' }];
+    }
+    else {
+      return [{ node: str, replacement: '""' }];
+    }
   }
 
 }
--- a/packages/stryker-typescript/src/mutator/StringLiteralMutator.ts
+++ b/packages/stryker-typescript/src/mutator/StringLiteralMutator.ts
@@ -11,7 +11,6 @@ export type AllStringLiterals =
 
 export default class StringLiteralMutator extends NodeMutator<AllStringLiterals> {
   name = 'StringLiteral';
-  imports: string[] = [];
 
   guard(node: ts.Node): node is AllStringLiterals {
     return node.kind === ts.SyntaxKind.StringLiteral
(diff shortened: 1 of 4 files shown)


packages/stryker-typescript/src/mutator/StringLiteralMutator.ts
```diff
@@ -0,0 +1,26 @@
...
+  imports: string[] = [];
+
+  guard(node: ts.Node): node is AllStringLiterals {
+    return node.kind === ts.SyntaxKind.StringLiteral
+      || node.kind === ts.SyntaxKind.TemplateExpression
+      || node.kind === ts.SyntaxKind.FirstTemplateToken;
+  }
+
+  protected identifyReplacements(str: AllStringLiterals, sourceFile: ts.SourceFile): NodeReplacement[] {
+    return [{ node: str, replacement: '""' }];
```
nicojs: What if the string was already empty? Maybe we should empty strings to `"Stryker was here"` or something.

packages/stryker-typescript/test/unit/mutator/StringLiteralMutatorSpec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  it('should mutate a template string', () => {
+    expectMutation(sut, 'const b = `Hello world!`;', 'const b = "";');
+  });
+
+  it('should mutate a template string referencing another variable', () => {
+    expectMutation(sut, 'const a = 10; const b = `${a} mutations`;', 'const a = 10; const b = "";');
+  });
+
+  it('should mutate a template string referencing another variable', () => {
```
nicojs: This test has the same name as the previous one.

packages/stryker-typescript/test/unit/mutator/StringLiteralMutatorSpec.ts
```diff
@@ -0,0 +1,48 @@
...
+    expectMutation(sut, 'const a = 10; const b = `mutations: ${a}`;', 'const a = 10; const b = "";');
+  });
+
+  it('should mutate a template string referencing another variable', () => {
```
nicojs: Same here. Are you testing something else? Or can this be removed?

packages/stryker-typescript/test/unit/mutator/StringLiteralMutatorSpec.ts
```diff
@@ -0,0 +1,48 @@
...
+    expectMutation(sut, 'const a = 10; const b = `mutations: ${a} out of 10`;', 'const a = 10; const b = "";');
+  });
+
+  it('should mutate import statements', () => {
```
nicojs: Is there a way to not mutate import strings? There is no point, it will always result in a transpile error.

mscharley: I tried few things but as of right now `node.parent` is always undefined, for all nodes. If it was not then theoretically, `node.parent.kind === ts.SyntaxKind.ImportDeclaration` and we could drop out in that case.

nicojs: Hmm until now we didn't need the `parent` property, but i can see why we need it here. I've created a separate PR for it here: https://github.com/stryker-mutator/stryker/pull/463. Once that's merged you should be able to use `node.parent`.

mscharley:
This latest commit will fail till #463 is in, but this is the fix for not mutating imports.

nicojs:
Ok, we will merge #463 tomorrow probably

simondel:
@mscharley You were two minutes faster than me with merging master into your fork 🥇 I got an error when I tried to push after the merge because you already updated it. Nice :)

mscharley:
@simondel I happened to be checking in at just the right time and noticed the PR got merged 😁

nicojs:
Thanks for a great job!

PR #462, opened 2017-11-15
feature(mutator specification): Create the generic mutator specification

nicojs:
Create a generic mutator specification as a seperate npm package in order to reuse between the javascript-mutator and typescript-mutator (and other mutators in the future).

### DIFF
--- a/packages/stryker-mutator-specification/src/ArrayLiteralMutatorSpec.ts
+++ b/packages/stryker-mutator-specification/src/ArrayLiteralMutatorSpec.ts
@@ -20,7 +20,7 @@ export default function ArrayLiteralMutatorSpec(name: string, expectMutation: Ex
     });
 
     it('should mutate empty array literals as a filled array', () => {
-      expectMutation('[]', '[[]]');
+      expectMutation('[]', '[\'Stryker was here\']');
     });
   });
 }
--- a/packages/stryker-mutator-specification/src/ArrayLiteralMutatorSpec.ts
+++ b/packages/stryker-mutator-specification/src/ArrayLiteralMutatorSpec.ts
@@ -1,0 +1,26 @@
+import { expect } from 'chai';
+import ExpectMutation from './ExpectMutation';
+
+export default function ArrayLiteralMutatorSpec(name: string, expectMutation: ExpectMutation) {
+
+  describe('ArrayLiteralMutator', () => {
+
+    it('should have name "ArrayLiteral"', () => {
+      expect(name).eq('ArrayLiteral');
+    });
+
+    it('should mutate filled array literals as empty arrays', () => {
+      expectMutation('[a, 1 + 1]', '[]');
+      expectMutation(`['val']`, '[]');
+    });
+
+    it('should not mutate array initializers (leave that for ArrayNewExpressionMutator)', () => {
+      expectMutation(`new Array()`);
+      expectMutation(`new Array(1, 2, 3)`);
+    });
+
+    it('should mutate empty array literals as a filled array', () => {
+      expectMutation('[]', '[[]]');
+    });
+  });
+}
--- a/packages/stryker-mutator-specification/src/ArrowFunctionMutatorSpec.ts
+++ b/packages/stryker-mutator-specification/src/ArrowFunctionMutatorSpec.ts
@@ -1,0 +1,19 @@
+import { expect } from 'chai';
+import ExpectMutation from './ExpectMutation';
+
+export default function ArrowFunctionMutatorSpec(name: string, expectMutation: ExpectMutation) {
+  describe('ArrowFunctionMutator', () => {
+
+    it('should have name "ArrowFunction"', () => {
+      expect(name).eq('ArrowFunction');
+    });
+
+    it('should mutate an anonymous function with an inline return', () => {
+      expectMutation('const b = () => 4;', 'const b = () => undefined;');
+    });
+
+    it('should not mutate an anonymous function with a block as a body', () => {
+      expectMutation('const b = () => { return 4; }');
+    });
+  });
+}
--- a/packages/stryker-mutator-specification/src/BlockMutatorSpec.ts
+++ b/packages/stryker-mutator-specification/src/BlockMutatorSpec.ts
@@ -1,0 +1,33 @@
+import { expect } from 'chai';
+import ExpectMutation from './ExpectMutation';
+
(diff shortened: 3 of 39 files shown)


packages/stryker-mutator-specification/src/ArrayLiteralMutatorSpec.ts
```diff
@@ -0,0 +1,26 @@
...
+      expectMutation(`['val']`, '[]');
+    });
+
+    it('should not mutate array initializers (leave that for ArrayNewExpressionMutator)', () => {
+      expectMutation(`new Array()`);
+      expectMutation(`new Array(1, 2, 3)`);
+    });
+
+    it('should mutate empty array literals as a filled array', () => {
+      expectMutation('[]', '[[]]');
```
simondel: Do we always want to support this mutation? Adding it to the spec project would encourage that

packages/stryker-mutator-specification/src/ArrowFunctionMutatorSpec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default function ArrowFunctionMutatorSpec(name: string, expectMutation: ExpectMutation) {
+  describe('ArrowFunctionMutator', () => {
+
+    it('should have name "ArrowFunction"', () => {
+      expect(name).eq('ArrowFunction');
+    });
+
+    it('should mutate an anonymous function with an inline return', () => {
+      expectMutation('const b = () => 4;', 'const b = () => undefined;');
```
simondel: Does every language that we (plan) to support with this spec have arrow functions?

packages/stryker-mutator-specification/src/BlockMutatorSpec.ts
```diff
@@ -0,0 +1,33 @@
...
+    it('should have name "Block"', () => {
+      expect(name).eq('Block');
+    });
+
+    it('should mutate the block of a function into an empty block', () => {
+      expectMutation('function() { return 4; }', 'function() {}');
+    });
+
+    it('should mutate a single block', () => {
+      expectMutation('const a = 3; { const b = a; }', 'const a = 3; {}');
```
simondel: Does every language that we (plan) to support with this spec have the `const` keyword?

packages/stryker-mutator-specification/src/ConditionalExpressionMutatorSpec.ts
```diff
@@ -0,0 +1,16 @@
...
+export default function ConditionalExpressionMutatorSpec(name: string, expectMutation: ExpectMutation) {
+
+  describe('ConditionalExpressionMutator', () => {
+
+    it('should have name "ConditionalExpression"', () => {
+      expect(name).eq('ConditionalExpression');
+    });
+
+    it('should replace conditional expressions', () => {
+      expectMutation('a < 3? b : c', 'false? b : c');
```
simondel: I would prefer to add a space between the condition and the `?`

packages/stryker-mutator-specification/src/ForStatementMutatorSpec.ts
```diff
@@ -0,0 +1,22 @@
...
+    it('should have name "ForStatement"', () => {
+      expect(name).eq('ForStatement');
+    });
+
+    it('should mutate the condition of a for statement', () => {
+      expectMutation('for(let i=0;i<10; i++) { console.log(); }', 'for(let i=0;false; i++) { console.log(); }');
+    });
+
+    it('should mutate the condition of a for statement without a condition', () => {
+      expectMutation('for(let i=0;; i++) { console.log(); }', `for (let i = 0; false; i++) {${os.EOL}    console.log();${os.EOL}}`);
```
simondel: We are formatting our code here. Is that intentional?  We don't format the previous for statement. How will we make sure that all mutators return this exact formatting change?

packages/stryker-mutator-specification/src/IfStatementMutatorSpec.ts
```diff
@@ -0,0 +1,17 @@
...
+export default function IfStatementMutatorSpec(name: string, expectMutation: ExpectMutation) {
+
+  describe('IfStatementMutator', () => {
+
+    it('should have name "IfStatement"', () => {
+      expect(name).eq('IfStatement');
+    });
+
+    it('should mutate an expression to `true` and `false`', () => {
+      expectMutation('if(something){ a++ }', 'if(true){ a++ }', 'if(false){ a++ }');
```
simondel: This code may be formatted by mutators by introducing a space between `if` and `(something)`

packages/stryker-mutator-specification/src/ArrayLiteralMutatorSpec.ts
nicojs: Maybe we should change `['Stryker was here!']`, would be more explicit.

packages/stryker-mutator-specification/src/ArrowFunctionMutatorSpec.ts
nicojs: Yes, the idea is that all "javascript like languages" should support mutation of arrow functions.

packages/stryker-mutator-specification/src/ConditionalExpressionMutatorSpec.ts
nicojs: That shouldn't matter right? Formatting does not change the intent here.

packages/stryker-mutator-specification/src/ForStatementMutatorSpec.ts
nicojs: Agreed, let's make the code here readable and change the expect implementation for typescript.

packages/stryker-mutator-specification/src/ArrowFunctionMutatorSpec.ts
nicojs: I'll add it in the readme 👍

packages/stryker-mutator-specification/src/BlockMutatorSpec.ts
nicojs: Yep, JSX, javascript and TypeScript support it. If it becomes a problem we can change it later

packages/stryker-mutator-specification/src/IfStatementMutatorSpec.ts
nicojs: Ok, i'll do that

simondel:
Awesome!

PR #463, opened 2017-11-16
refactor(typescript): Set parent nodes

nicojs:
Use a slightly more low level api to parse typescript files so we can force typescript to set parent nodes. This is useful for creating node mutators, as we can use information about the parent to determine what tho mutate.

### DIFF
--- a/packages/stryker-typescript/src/helpers/tsHelpers.ts
+++ b/packages/stryker-typescript/src/helpers/tsHelpers.ts
@@ -74,10 +74,16 @@ export function isHeaderFile(file: File) {
   return file.name.endsWith('.d.ts');
 }
 
-export function filterOutTypescriptFiles(files: File[]): TextFile[] {
+/**
+ * Returns all the files that are considered typescript files (text files with *.ts or something like that)
+ */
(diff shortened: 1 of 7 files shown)


nicojs:
@mscharley would you mind taking a look as reviewer?

packages/stryker-typescript/src/helpers/tsHelpers.ts
```diff
@@ -54,15 +62,14 @@ export function printNode(node: ts.Node, originalSourceFile: ts.SourceFile): str
...
 const allExtensions: string[] = Object.keys(ts.Extension).map(extension => ts.Extension[extension as any]);
-export function isToBeTranspiled(file: File) {
+export function isTypescriptFile(file: File) {
   return file.kind === FileKind.Text &&
-    file.transpiled &&
     allExtensions.some(extension => file.name.endsWith(extension)) && !file.name.endsWith('.d.ts');
 }
 
 
 export function filterOutTypescriptFiles(files: File[]): TextFile[] {
```
simondel: You can read the name of this method two ways:
> Give me all typescript files in `files`  

Or
> Give me everything **but** the typescript files

PR #473, opened 2017-11-26
fix(BabelParser): Support parsing modules

simondel:


simondel:
I would suggest merging this as a bugfix because we thought this worked but we never tested it.

packages/stryker-javascript-mutator/src/helpers/BabelParser.ts
```diff
@@ -5,7 +5,13 @@ import { NodePath } from 'babel-traverse';
 
 export default class BabelParser {
   static getAst(code: string): babel.types.File {
-    return babylon.parse(code);
+    let ast: babel.types.File;
+    try {
+      ast = babylon.parse(code, { sourceType: 'script' });
+    } catch {
```
Archcry: Isn't there an error code we can check?

simondel: As discussed, the message that you get is `'import' and 'export' may appear only with 'sourceType: "module"' (1:0)` where the range at the end may differ depending on your code. As this message may change in the future, I think checking on it is not the smartest move here. Also, if a different error is thrown, it will be trown again if we try parsing it as a module. Nice catch though!

nicojs: Nice `try` @simondel

simondel: da dumm tssss

Archcry: Ye, good `catch`

PR #475, opened 2017-11-26
feature(stryker): #336 Configurable tmp folder location 

avassem85:


packages/stryker/README.md
```diff
@@ -278,3 +278,11 @@ Set `break` to `null` (default) to never let the process crash.
  Set the `log4js` log level that Stryker uses (default is `info`). Possible values: `fatal`, `error`, `warn`, `info`, `debug`, `trace`, `all` and `off`.  
  *Note*: Test runners are run as child processes of the Stryker Node process. All output (stdout) of the `testRunner` is logged as `trace`.  
  Thus, to see logging output from the test runner set the `logLevel` to `all` or `trace`.
+
+#### Temporary folder  
+**Command line:** `--tempFolder .stryker-tmp`    
+**Config file:** `logLevel: '.stryker-tmp'`
```
nicojs: Still refers to `logLevel` here

packages/stryker/src/StrykerCli.ts
```diff
@@ -47,6 +47,7 @@ export default class StrykerCli {
       .option('--timeoutFactor <number>', 'Tweak the standard deviation relative to the normal test run of a mutated test', parseFloat)
       .option('--maxConcurrentTestRunners <n>', 'Set the number of max concurrent test runner to spawn (default: cpuCount)', parseInt)
       .option('--logLevel <level>', 'Set the log4js log level. Possible values: fatal, error, warn, info, debug, trace, all and off. Default is "info"')
+      .option('--tempFolder <name>', 'Set name of the folder that contains temporary files')
```
nicojs: Should we call it `tempFolder` or `tempDir` (or `tempDirectory`). I think `tempDir` is more clear, we should always refer to `directory` instead of `folder`.

packages/stryker/src/utils/TempFolder.ts
```diff
@@ -15,6 +15,7 @@ export class TempFolder {
   initialize(tempDirName = '.stryker-tmp') {
```
nicojs: This variable `tempDirName` should not have a default value, as this is a duplicate from the `Config` class.

packages/stryker/test/unit/utils/TempFolderSpec.ts
```diff
@@ -50,18 +50,21 @@ describe('TempFolder', () => {
...
         const tempFolderInstance = TempFolder.instance();
-        const result = tempFolderInstance.clean();
+        const result = await tempFolderInstance.clean();
 
         expect(fileUtils.deleteDir).to.have.been.calledWith(
           tempFolderInstance.baseTempFolder
         );
 
-        result.then(data => expect(data).equals('delResolveStub'));
+        expect(result).equals('delResolveStub');
```
nicojs: This is better indeed 👍

packages/stryker/README.md
```diff
@@ -278,3 +278,11 @@ Set `break` to `null` (default) to never let the process crash.
...
+**Default value:** `.stryker-tmp`  
+**Mandatory**: no 
+**Description:**  
+ Set name of the folder that contains temporary files
```
simondel: I think it's smart to mention that stryker will attempt to delete this folder after it's done.

packages/stryker/src/StrykerCli.ts
simondel: I think `tempDir` is better. But if we change it we should also change the `TempFolder` class.

simondel:
@nicojs Why would an absolute path break mocha?

packages/stryker/README.md
avassem85: Done

packages/stryker/src/StrykerCli.ts
avassem85: Renamed to tempDir

packages/stryker/README.md
avassem85: Done

avassem85:
@simondel Because mocha uses Require that search node_modules up in the directories of the application. When the temp folder is not within the root of the application is will fail.

packages/stryker-api/src/config/Config.ts
```diff
@@ -19,6 +19,7 @@ export default class Config implements StrykerOptions {
   mutator: string = 'es5';
   transpilers: string[] = [];
   maxConcurrentTestRunners: number = Infinity;
+  tempFolder: string = '.stryker-tmp';
```
nicojs: This also needs to be renamed to `tempDir`

packages/stryker/src/utils/StrykerTempFolder.ts
```diff
@@ -1,74 +0,0 @@
-import * as fs from 'mz/fs';
```
nicojs: Good catch! Removing code makes me happy

packages/stryker/README.md
```diff
@@ -278,3 +278,11 @@ Set `break` to `null` (default) to never let the process crash.
...
  *Note*: Test runners are run as child processes of the Stryker Node process. All output (stdout) of the `testRunner` is logged as `trace`.  
  Thus, to see logging output from the test runner set the `logLevel` to `all` or `trace`.
+
+#### Temporary folder  
+**Command line:** `--tempDir .stryker-tmp`    
+**Config file:** `tempDir: '.stryker-tmp'`
+**Default value:** `.stryker-tmp`  
+**Mandatory**: no 
+**Description:**  
+ Set name of the folder that contains temporary files
```
nicojs: Let's make this a little more descriptive. Something like: 

Set the name of the directory that is used by Stryker as a working directory. This directory is used to keep copies of your files that are used by Stryker to mutate your code safely. This directory will be cleaned after a successful run. **Warning!** If you configure an absolute path (for example `/tmp/stryker`), node-based test runners like mocha will fail, because your `node_modules` cannot be loaded. Use with caution.

packages/stryker/src/StrykerCli.ts
```diff
@@ -47,6 +47,7 @@ export default class StrykerCli {
       .option('--timeoutFactor <number>', 'Tweak the standard deviation relative to the normal test run of a mutated test', parseFloat)
       .option('--maxConcurrentTestRunners <n>', 'Set the number of max concurrent test runner to spawn (default: cpuCount)', parseInt)
       .option('--logLevel <level>', 'Set the log4js log level. Possible values: fatal, error, warn, info, debug, trace, all and off. Default is "info"')
+      .option('--tempDir <name>', 'Set name of the folder that contains temporary files. Stryker will try to remove the content when the tests ran')
```
nicojs: Let's also update this one. Something like:

"Set the name of the directory that is used by Stryker as a working directory.  This directory will be cleaned after a successful run. "

nicojs:
Only the deepest directory is deleted. Is this acceptable? 
image

Also it will be deleted regardless of any other data in the directory. So if you would point it to `'/'`, your entire computer will be wiped clean after a successful run.

avassem85:
@nicojs 

1. If you delete more than the deepest directory you could end up deleting to much (temporary locations for other tools)
2. You can't accidentally delete root(/) because the directory is always relative from the root of your project

nicojs:
> You can't accidentally delete root(/) because the directory is always relative from the root of your project

Well... i could incidentally use  `tempDir: '../../../../../'`

avassem85:
@nicojs true, but what do you want to do about it?

nicojs:
> @nicojs true, but what do you want to do about it?

Well.. Maybe we could clean the directory first. Remove all sandbox sub directories. If the directory is *empty* after that, we can remove it safely.

To remove all sandbox sub directories, we can just use rimraf and give it the pattern: `'.stryker-tmp/sandbox*'`. Also i've noticed that we are creating a temp folder within the temp folder. I don't know why. I'm talking about this line: https://github.com/stryker-mutator/stryker/blob/24699e16e9c3c8ea77ecb52f4f1b6a8729ac5306/packages/stryker/src/utils/TempDir.ts#L17

I think we can safely remove that.

nicojs:
This PR is stale. Closing for now, please create a new PR.

brunoqueiros:
any plans of configurable tempFolder?

nicojs:
We've got lot of plans 😅 

This is pretty easy to solve if we can simply give it another name. If you want support for it inside your OS temp folder it will take some more time. What would you prefer?

brunoqueiros:
I was thinking something similar with this PR where you just specify another name inside the project itself. `—tempDir=.another-path-tmp`. The main reason for that is to be able to run multiple `stryker run` commands in parallel.

Do you need help with it?

nicojs:
Help is always welcome!

The best solution would be to make [`TempFolder`](https://github.com/stryker-mutator/stryker/blob/master/packages/core/src/utils/TempFolder.ts) injectable and implement the `Disposable` interface from [`typed-inject`](https://github.com/nicojs/typed-inject#readme). That way we can also remove the explicit call to [`TempFolder.instance().clean`](https://github.com/stryker-mutator/stryker/blob/ed839384a428078696af4fd56777612d76624486/packages/core/src/Stryker.ts#L87) (since it will be injectable).

If you need help setting up the dependency injection, please let me know.

Inside the `TempFolder` you can let the `options` be injected. We add an option as you propose. 

We should also rename `TempFolder` to `TempDir` or `TemporaryDirectory`. Calling directories folders is kind of a pet peeve of mine 😅 .

brunoqueiros:
i created a PR https://github.com/stryker-mutator/stryker/pull/1644, could you check if I am going in the right track regarding the DI.

PR #480, opened 2017-11-28
feat(Babel): Add support for Babel projects

simondel:


CrazyBS:
!
I've been waiting for this!  I'm excited!

simondel:
@CrazyBS What kind of project setup do you have?

nicojs:
Just to clarify: this PR will make it easy to run stryker in babel projects *without additional bundeling* no webpack yet (but its comming)

nicojs:
I just realized this PR is waiting for my approval. Hang on, I'll get started right away 😅

packages/stryker-babel/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "log4js": "^1.1.1"
+  },
+  "devDependencies": {
+    "@types/babel-core": "^6.25.3",
+    "babel-plugin-transform-es2015-spread": "^6.22.0",
+    "babel-preset-es2015": "^6.24.1",
+    "stryker-api": "^0.11.0"
+  },
+  "peerDependencies": {
+    "stryker-api": "^0.11.0"
```
nicojs: peer dependency to babel-core is missing? Or by design? If it is by design, we should remove the remark in the README.md

packages/stryker-babel/src/BabelConfigEditor.ts
```diff
@@ -0,0 +1,37 @@
...
+    this.log.trace(`babelConfig set to: ${JSON.stringify(config[CONFIG_KEY_OPTIONS])}`);
+  }
+
+  private readConfig(config: Config) {
+    if (typeof config[CONFIG_KEY_FILE] === 'string') {
+      const babelrcPath = path.resolve(config[CONFIG_KEY_FILE]);
+      this.log.info(`Reading .babelrc file from path "${babelrcPath}"`);
+      return this.getConfigFile(babelrcPath);
+    } else {
+      this.log.warn(`No .babelrc file configured. Please set the "${CONFIG_KEY_FILE}" property in your config.`);
```
nicojs: Should this be a warning? You cannot remove a ConfigEditor, all config editors are loaded for each run. Maybe add it as an info?

packages/stryker-babel/src/BabelTranspiler.ts
```diff
@@ -0,0 +1,87 @@
...
+import { CONFIG_KEY_OPTIONS } from './helpers/keys';
+
+class BabelTranspiler implements Transpiler {
+  private babelConfig: babel.GeneratorOptions;
+  private knownExtensions: string[];
+
+  public constructor(options: TranspilerOptions) {
+    this.babelConfig = options.config[CONFIG_KEY_OPTIONS];
+
+    this.knownExtensions = ['.js', '.jsx', '.ts'];
```
nicojs: Where did you find this list? Why would typescript be allowed? If that is the case, maybe we also need all these extensions: https://github.com/Microsoft/TypeScript/blob/0b9e4987f477f46e66848a08e8f3db38338c84e0/src/compiler/types.ts#L4247

packages/stryker-babel/src/BabelTranspiler.ts
```diff
@@ -0,0 +1,87 @@
...
+  public transpile(files: File[]): Promise<TranspileResult> {
+    const outputFiles: File[] = [];
+    const errors: string[] = [];
+
+    files.forEach((file) => {
+      if (this.shouldBeTranspiled(file)) {
+        try {
+          outputFiles.push({
+            name: file.name,
+            content: this.transform((file as TextFile).content),
```
nicojs: You could also use a typeguard for this, but you can leave it as well.

packages/stryker-babel/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "version": "0.1.0",
+  "description": "A plugin for babel projects using Stryker",
+  "main": "src/index.js",
+  "scripts": {
+    "start": "tsc -w",
+    "clean": "rimraf \"+(test|src)/**/*+(.d.ts|.js|.map)\" reports",
+    "prebuild": "npm run clean",
+    "build": "tsc -p .",
+    "postbuild": "tslint -p tsconfig.json",
+    "test": "nyc --reporter=html --report-dir=reports/coverage --check-coverage --lines 85 --functions 90 --branches 60 mocha  \"test/helpers/**/*.js\" \"test/unit/**/*Spec.js\" \"test/integration/**/*Spec.js\""
```
nicojs: Why this low `branches` coverage check?

integrationTest/test/babel/src/Casino.js
```diff
@@ -0,0 +1,28 @@
+var Bank = require('./Bank').default;
```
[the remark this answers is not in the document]
simondel: I explicitly did not use an `import` to test if stryker-babel and stryker-javascript-mutator can deal with require statments instead of imports (modules vs scripts)

packages/stryker-babel/src/BabelConfigEditor.ts
simondel: If we don't have any babel config, the babeltranspiler won't transpile. It will just output the same code.

integrationTest/test/babel/src/Casino.js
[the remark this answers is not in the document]
nicojs: Great reason! 👍

packages/stryker-babel/src/BabelConfigEditor.ts
nicojs: I understand that you need it for transpilation, but who says the babel transpiler is even configured?

CrazyBS:
@simondel It's posted in https://github.com/stryker-mutator/stryker/issues/101

simondel:
@CrazyBS Unfortunately that project includes webpack which is not done yet. Sorry man :/

integrationTest/test/babel/src/Casino.js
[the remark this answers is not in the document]
simondel: I added a comment so we don't forget it

packages/stryker-babel/package.json
simondel: Fixed!

packages/stryker-babel/src/BabelTranspiler.ts
simondel: I removed typescript from the list

packages/stryker-babel/package.json
simondel: Fixed! users may not have babel-core directly if they use the babel-cli. But that does have babel-core as a dependency

packages/stryker-babel/src/BabelTranspiler.ts
simondel: We could, but the guard is already implemented (separate method: `shouldBeTranspiled`)

PR #481, opened 2017-12-01
fix(Babel): Rename stryker-babel to stryker-babel-transpiler

simondel:


packages/stryker-babel-transpiler/src/BabelConfigEditor.ts
```diff
@@ -0,0 +1,34 @@
+import * as fs from 'fs';
```
nicojs: Shouldn't this file and clas be called `BabelConfigReader` now?

simondel:
I renamed it. Let's see if this build succeeds 🥇

Issue #521, opened 2017-12-04
Async/Await javascript support


webdevian:
I get a parse error thrown in stryker / esprima for async methods on a class. It appears upgrading esprima to v4 will fix this - https://github.com/jquery/esprima/issues/1875#issuecomment-341885659

I've got no idea if this will have other side effects though

simondel:
Odd. That means you are still using the ES5 mutator. Is this the same program that you opened #477 for? If that's not the case, could you try upgrading to the `stryker-javascript-mutator` as you did with the program in #477?

We won't be upgrading Esprima because the ES5 mutator is deprecated and will be removed soon.

webdevian:
No it's a fresh install and init with `"stryker-javascript-mutator": "^0.2.3",`

webdevian:
My config looks like this - 
```js
module.exports = config => {
  config.set({
    files: [
      {
        pattern: 'lib/**/*.js',
        included: true
      },
      {
        pattern: 'src/**/*.js',
        mutated: true,
        included: true
      },
      'test/**/*.js'
    ],
    testRunner: 'mocha',
    mutator: 'javascript',
    transpilers: [],
    reporter: ['html', 'baseline', 'clear-text', 'progress'],
    testFramework: 'mocha',
    coverageAnalysis: 'perTest'
  })
}
```

The error thrown is this:

```
[2017-12-04 12:55:42.624] [ERROR] StrykerCli - an error occurred Error: Could not transpile input files: Execute CoverageInstrumenterTranspiler: Error: Could not instrument "/Users/iegner/Code/huestatus/src/Hue.js" for code coverage. Error: Line 30: Unexpected identifier
Error: Line 30: Unexpected identifier
    at constructError (/Users/iegner/Code/huestatus/node_modules/istanbul/node_modules/esprima/esprima.js:2407:21)
    at createError (/Users/iegner/Code/huestatus/node_modules/istanbul/node_modules/esprima/esprima.js:2426:17)
    at unexpectedTokenError (/Users/iegner/Code/huestatus/node_modules/istanbul/node_modules/esprima/esprima.js:2500:13)
    at throwUnexpectedToken (/Users/iegner/Code/huestatus/node_modules/istanbul/node_modules/esprima/esprima.js:2505:15)
    at parseClassBody (/Users/iegner/Code/huestatus/node_modules/istanbul/node_modules/esprima/esprima.js:5147:21)
[...]

PR #522, opened 2017-12-04
Upgrade Esprima

webdevian:
Possible fix for #521


simondel:
As discussed in #521 we'll close this.

PR #523, opened 2017-12-04
chore(travis): Build against latest LTS

webdevian:
I noticed that Node v8 is missing from the travis builds since 9 was released


simondel:
Nice catch! Could you also drop v7 support as it's no longer maintained?

PR #550, opened 2017-12-14
feat(cvg analysis): New coverage instrumenter

nicojs:
Use the new `istanbul-lib-instrumenter` instrumenter to replace the old instanbul instrumenter. The new instrumenter is based on babel and supports the latest javascript constructs. Use the function map as well as statement map, because functions are no longer seen as 'statements


packages/stryker/src/MutantTestMatcher.ts
```diff
@@ -1,19 +1,35 @@
...
+  statement
+}
+
+/**
+ * Represents a location inside the coverage data of a file
+ * Either the function map, or statement map
+ */
+interface CoveredCodeIndicator {
+  type: CoveredCodeIndicatorType;
+  index: string;
```
simondel: Why set the index here and not the actual value? Right now it looks like we set the index to either the `statementIndex` or `functionIndex` and then read it later on. That would also remove the need for the vague name 'index'

packages/stryker/package.json
```diff
@@ -53,13 +53,14 @@
     "stryker": "./bin/stryker"
   },
   "dependencies": {
+    "@types/istanbul-lib-instrument": "^1.7.0",
```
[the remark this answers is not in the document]
nicojs: :+1:

packages/stryker/src/MutantTestMatcher.ts
nicojs: What actual value? We need thr index, because the index in the statement/fn map is also the key of the coverage data objects (`s` and `f`)

packages/stryker/src/transpiler/CoverageInstrumenterTranspiler.ts
```diff
@@ -1,26 +1,30 @@
...
 const COVERAGE_CURRENT_TEST_VARIABLE_NAME = '__strykerCoverageCurrentTest__';
 
-export interface StatementMapDictionary {
-  [file: string]: StatementMap;
+export interface FileCoverageMaps {
+  statementMap: { [key: string]: Range };
+  fnMap: { [key: string]: { loc: Range } };
+}
+
+export interface FileCoverageDataDictionary {
```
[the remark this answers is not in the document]
nicojs: Yes, the interface is just to function as a quick access. So we don't have to iterate the array once per mutant. Remove it and use an array?

packages/stryker/src/transpiler/CoverageInstrumenterTranspiler.ts
```diff
@@ -1,26 +1,30 @@
...
 import { TestFramework } from 'stryker-api/test_framework';
 import { Logger, getLogger } from 'log4js';
+import { FileCoverageData, Range } from 'istanbul-lib-coverage';
 
-const coverageObjRegex = /\{.*"path".*"fnMap".*"statementMap".*"branchMap".*\}/g;
 const COVERAGE_CURRENT_TEST_VARIABLE_NAME = '__strykerCoverageCurrentTest__';
 
-export interface StatementMapDictionary {
-  [file: string]: StatementMap;
+export interface FileCoverageMaps {
```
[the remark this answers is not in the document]
nicojs: Sorry i don't know what you mean. We need both the statement map as well as the fn map. We could use the istanbul interface directly, but we don't usr all its properties. It still might be clearer though. What do you think?

nicojs:
@simondel I've simplified some parts. For example: the coverage maps interface is now defined as:

```typescript
export interface CoverageMaps {
  statementMap: { [key: string]: Range };
  fnMap: { [key: string]: Range };
}
```

This makes the code inside `MutantTestMatcher` easier, as both maps have the same structure (i could remove this method: `findMatchingFunction` ). 

I've did another test and the coverageAnalysis `'off'` resulted in *exactly* the same result as `'perTest'`.

packages/stryker/src/MutantTestMatcher.ts
```diff
@@ -152,25 +150,6 @@ export default class MutantTestMatcher {
     }
   }
 
-  /**
-   * 
-   * @param mutant The mutant
-   * @param functionMap The function map of the covering file
-   * @returns The index of the smallest function covering the mutant, or null if not found
-   */
-  private findMatchingFunction(mutant: TestableMutant, functionMap: { [key: string]: { loc: Range } }): string | null {
```
simondel: Was this obsolete code?

packages/stryker/src/transpiler/CoverageInstrumenterTranspiler.ts
```diff
@@ -8,19 +8,19 @@ import { FileCoverageData, Range } from 'istanbul-lib-coverage';
...
+export interface CoverageMaps {
   statementMap: { [key: string]: Range };
-  fnMap: { [key: string]: { loc: Range } };
+  fnMap: { [key: string]: Range };
 }
 
-export interface FileCoverageDataDictionary {
-  [file: string]: FileCoverageMaps;
+export interface CoverageMapsByFile {
+  [file: string]: CoverageMaps;
```
simondel: Why don't we store the filename in the coveragemap?

packages/stryker/src/MutantTestMatcher.ts
simondel: We set the index in `findMatchingCoveringIndicator` and in `isCoveredByCoverageCollection` we read the value with the index that we have set. Why can't we read the value in `findMatchingCoveringIndicator` and add that value to the object instead of the index?

simondel:
As discussed on the phone. We'll leave it like it is for now because this will be refactored in the future. The code is already cleaner like this 👍

packages/stryker/src/MutantTestMatcher.ts
nicojs: Yeah we didn't need it.

packages/stryker/src/transpiler/CoverageInstrumenterTranspiler.ts
nicojs: As a sneak preview, i'm refactoring it like this:

```
abstract class CoveredStatement {
  fileName: string;
  range: Range;
  abstract getCoveringTestIds(): number[];
}
```

Probably a lot cleaner like this.

Issue #545, opened 2017-12-07
Can Stryker use for projects with Flow type? (https://flow.org/)


vmatheu:
Can Stryker use for projects with Flow type?

simondel:
Do you use `babel-preset-flow`? If that is the case, the [stryker-babel-transpiler](https://www.npmjs.com/package/stryker-babel-transpiler) combined with the [stryker-javascript-mutator](https://www.npmjs.com/package/stryker-javascript-mutator) may be what you're looking for. You can set this up when using our [quickstart](https://stryker-mutator.github.io/quickstart.html)

guigonc:
Hi @simondel I work in this project with @vmatheu and our configurations is defined as follows.

`stryker.conf.json`
```
module.exports = function(config) {
  config.set({
    files: [
      { pattern: "src/**/*.*", mutated: true, included: false },
      "test/unit/**/*.js",
    ],
    testRunner: "mocha",
    mochaOptions: {
      recursive: true,
      compilers: ['js:babel-register', 'js:babel-polyfill'],
      require: ['ignore-styles']
    },
    mutator: "javascript",
    transpilers: ["babel"],
    reporter: ["html", "clear-text", "progress"],
    testFramework: "mocha",
    coverageAnalysis: "off",
    babelrcFile: ".babelrc"
  });
};
```
`package.json`
```
...
devDependencies: {
    ...
    "babel-preset-es2015": "^6.18.0",
    "babel-preset-flow": "^6.23.0",
    "babel-preset-react": "^6.16.0",
...
```
`.babelrc`
```
{
  "presets": ["es2015", "react", "flow"],
  "plugins": ["transform-object-rest-spread", "transform-class-properties"]
}
```
But we still receive an error after the initial test run, and we can only pass through this error by erasing the flow types definition.
Here is the error.

```
[2017-12-11 13:37:35.441] [INFO] InitialTestExecutor - Initial test run succeeded. Ran 447 tests in 10 seconds.
[2017-12-11 13:37:35.453] [ERROR] StrykerCli - an error occurred { SyntaxError: Unexpected token (5:12)
    at Parser.pp$5.raise (/Users/ggoncalv/Projects/Latam/booking-webapp/node_modules/stryker-javascript-mutator/node_modules/babylon/lib/index.js:4454:13)
[...]

PR #552, opened 2017-12-18
Feat(React): Add support for jsx and flow

Archcry:
Should fix the issue in: https://github.com/stryker-mutator/stryker/issues/545
Also adds a check for knownextensions ['.js', '.jsx'] so it will not try to mutate other files like html or css.

### DIFF
--- a/packages/stryker-javascript-mutator/src/JavaScriptMutator.ts
+++ b/packages/stryker-javascript-mutator/src/JavaScriptMutator.ts
@@ -15,7 +15,6 @@ function defaultMutators(): NodeMutator[] {
 
 export default class JavaScriptMutator implements Mutator {
   private log = getLogger(JavaScriptMutator.name);
-  private knownExtensions: Array<string> = ['.js', '.jsx'];
 
   constructor(config: Config, private mutators: NodeMutator[] = defaultMutators()) {
   }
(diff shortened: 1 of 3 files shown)


packages/stryker-javascript-mutator/src/JavaScriptMutator.ts
```diff
@@ -7,21 +7,23 @@ import BabelParser from './helpers/BabelParser';
...
 import NodeMutator from './mutators/NodeMutator';
+import * as path from 'path';
 
 function defaultMutators(): NodeMutator[] {
   return NodeMutatorFactory.instance().knownNames().map(name => NodeMutatorFactory.instance().create(name, undefined));
 }
 
 export default class JavaScriptMutator implements Mutator {
   private log = getLogger(JavaScriptMutator.name);
+  private knownExtensions: Array<string> = ['.js', '.jsx'];
```
simondel: We can move this list to the hasValidExtension function since we only use it there

packages/stryker-javascript-mutator/src/JavaScriptMutator.ts
```diff
@@ -41,6 +43,16 @@ export default class JavaScriptMutator implements Mutator {
     return mutants;
   }
 
+  private hasValidExtension(file: TextFile): boolean {
+    for (let extension of this.knownExtensions) {
```
simondel: We could also use `return this.knownExtensions.some(extension => path.extname(file.name) === extension);`

packages/stryker-javascript-mutator/src/helpers/BabelParser.ts
```diff
@@ -6,10 +6,20 @@ import { NodePath } from 'babel-traverse';
 export default class BabelParser {
   static getAst(code: string): babel.types.File {
     let ast: babel.types.File;
+    
+    const options: babylon.BabylonOptions = {
+      sourceType: 'script',
+      plugins: [
+        'jsx',
+        'flow'
```
simondel: You added flow here but I'm missing a testcase for flow

packages/stryker-javascript-mutator/src/JavaScriptMutator.ts
simondel: Could you also change it to a `string[]` or remove the type entirely?

PR #560, opened 2017-12-26
 chore(dependencies): Upgrade to chalk 2.3.0

simondel:
The types are now included with the release of chalk so they have been removed from the devDependencies.


nicojs:
Awesome to see that TypeScript is adopted by library maintainers.

PR #588, opened 2018-01-19
Stryker webpack transpiler

nicojs:
> A plugin to use webpack in Stryker

Some dude I know...

(See readme for more details)


packages/stryker-webpack-transpiler/src/compiler/WebpackCompiler.ts
```diff
@@ -0,0 +1,66 @@
...
+  private _inputFS: InputFileSystem;
+  private _outputFS = new OutputFileSystem();
+
+  public constructor(webpackConfig: Configuration) {
+    this._compiler = this.createCompiler(webpackConfig);
+  }
+
+  private createCompiler(webpackConfig: Configuration): Compiler {
+    const compiler = webpack(webpackConfig);
+    this._inputFS = new InputFileSystem();
```
Archcry: Can't you pass an initialized `InputFileSystem` in the constructor, thereby using dependency injection? That would also make it easier to mock away the filesystem in the unit tests.

packages/stryker-webpack-transpiler/src/compiler/WebpackCompiler.ts
```diff
@@ -0,0 +1,66 @@
...
+import { Compiler, Configuration } from 'webpack';
+import webpack from './Webpack';
+import * as path from 'path';
+import InputFileSystem from '../fs/InputFileSystem';
+import OutputFileSystem from '../fs/OutputFileSystem';
+
+export default class WebpackCompiler {
+  private _compiler: Compiler;
+  private _inputFS: InputFileSystem;
+  private _outputFS = new OutputFileSystem();
```
Archcry: Can't you pass an initialized InputFileSystem in the constructor, thereby using dependency injection? That would also make it easier to mock away the filesystem in the unit tests.

packages/stryker-webpack-transpiler/src/compiler/WebpackCompiler.ts
```diff
@@ -0,0 +1,66 @@
...
+  public writeFilesToFs(files: Array<File>): void {
+    for (let file of files) {
+      if (file.kind !== FileKind.Web) {
+        this.writeToFs(file);
+      }
+    }
+  }
+
+  private writeToFs(file: TextFile | BinaryFile): void {
+    // Create the directory
```
Archcry: Is this comment really necessary?

packages/stryker-webpack-transpiler/src/compiler/ConfigLoader.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Configuration } from 'webpack';
+import { StrykerWebpackConfig } from '../WebpackTranspiler';
+import { getLogger } from 'log4js';
+
+const PROGRESS_PLUGIN_NAME = 'ProgressPlugin';
+
+export default class ConfigLoader {
+  private log = getLogger(ConfigLoader.name);
+
+  public constructor(private loader: NodeRequireFunction = require) {
```
Archcry: My personal preference is to explicitly define the private variables instead of using the TypeScript shortcut, I think this makes the code clearer for people that haven't worked with TypeScript.

packages/stryker-webpack-transpiler/src/fs/InputFileSystem.ts
```diff
@@ -0,0 +1,126 @@
+import MemoryFS from './MemoryFS';
+import { webpack, Callback } from '../types';
+import * as fs from 'fs';
+const errors = require('errno');
+import { dirname } from 'path';
+import { getLogger } from 'log4js';
+
+export default class InputFileSystem implements webpack.InputFileSystem {
```
Archcry: It would be nice to see `public` keywords defined explicitly.

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
```diff
@@ -0,0 +1,94 @@
...
+const binaryFileExtensions = Object.freeze(['.ico', '.zip']);
+
+export default class OutputFileSystem implements webpack.OutputFileSystem {
+
+  private _files: {
+    [name: string]: string | Buffer;
+  };
+
+  constructor() {
+    this.purge();
```
Archcry: I get why you call `purge` here (to initialize `this_files`) but it's weird to see a purge in a constructor.

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
```diff
@@ -0,0 +1,94 @@
+import { PathLike } from 'fs';
+import * as path from 'path';
+import { webpack, EmptyCallback, Callback } from '../types';
+import { BinaryFile, TextFile, FileKind } from 'stryker-api/core';
+import ChunkSorter, { Chunk } from '../compiler/ChunkSorter';
+
+const binaryFileExtensions = Object.freeze(['.ico', '.zip']);
+
+export default class OutputFileSystem implements webpack.OutputFileSystem {
```
Archcry: Could you specify the `public` keywords explicitly?

packages/stryker-webpack-transpiler/test/helpers/sinonInit.ts
```diff
@@ -0,0 +1,17 @@
...
+    debug: sandbox.stub(),
+    info: sandbox.stub(),
+    warn: sandbox.stub(),
+    error: sandbox.stub()
+  };
+  sandbox.stub(log4js, 'getLogger').returns(global.logMock);
+});
+
+afterEach(() => {
+  global.sandbox.restore();
```
Archcry: Can be shortened by doing `afterEach(() => global.sandbox.restore())`

packages/stryker-webpack-transpiler/test/unit/compiler/ConfigLoaderSpec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  it('should load webpack config from given location', () => {
+    requireStub.returns('resolved');
+    const result = sut.load(createStrykerWebpackConfig({ configFile: 'webpack.foo.config.js' }));
+    expect(result).eq('resolved');
+    expect(requireStub).calledWith(path.resolve('webpack.foo.config.js'));
+  });
+
+  it('should remove "ProgressPlugin" if silent is `true`', () => {
+    // Arrange
```
Archcry: Why are there suddenly comments here?

packages/stryker-webpack-transpiler/src/compiler/WebpackCompiler.ts
Archcry: I know this is possible using sandboxes but the cleaner way would be injecting them in the constructor.

packages/stryker-webpack-transpiler/src/compiler/ConfigLoader.ts
nicojs: I understand that sentiment. However, that would mean having to repeat myself (variable name + type). The benefits out way the readability argument here.

It took me a while to get use to this syntax as well, but in the end: people should know the language (at least a little bit) before contributing.

packages/stryker-webpack-transpiler/src/compiler/WebpackCompiler.ts
nicojs: I agree that constructor params are more explicit. I'll change it to:

```ts
  public constructor(webpackConfig: Configuration,
    private _inputFS = new InputFileSystem(),
    private _outputFS = new OutputFileSystem()) {
    this._compiler = this.createCompiler(webpackConfig);
  }
```

packages/stryker-webpack-transpiler/src/compiler/WebpackCompiler.ts
nicojs: Nope, a remnant of the past. Removed 👍

packages/stryker-webpack-transpiler/src/fs/InputFileSystem.ts
nicojs: Meh... I don't mind either way. We don't have a strict rule for it either. I will add them here.

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
nicojs: It's more DRY this way. It's not that strange right? You start out with a purged file system.

packages/stryker-webpack-transpiler/test/helpers/sinonInit.ts
nicojs: I know, but I've burned myself too often on this:

Lets say you have this line of code:

```ts
beforeEach(() => thing.initialize());
```

Later on you decide to add a statement. That is a pain already, since you need to add `{ ...; }`:

```ts
beforeEach(() => {
  thing = new Thing();
  thing.initialize();
});
```

Do you see the problem here? Since the result of `initialize()` is no longer (implicitly) returned, the test will run before the `initiliaze` is finished. 

This is why I decided to always use brackets when dealing with `beforeEach()`, `it()` and `afterEach()`

packages/stryker-webpack-transpiler/test/unit/compiler/ConfigLoaderSpec.ts
nicojs: Because the tests  has 8 lines of code. I usually add them when there are more lines, just to keep it clean.

nicojs:
@Archcry I've commented on/fixed all issues. Are you happy?

packages/stryker-webpack-transpiler/test/helpers/sinonInit.ts
Archcry: Ah that's fair.

Archcry:
Looks good to me, good to go.

nicojs:
@simondel can we merge this?

simondel:
@nicojs not yet, I want to review this. Do you want me to give this one priority over the source map one?

packages/stryker-webpack-transpiler/src/compiler/ChunkSorter.ts
```diff
@@ -0,0 +1,57 @@
...
+const toposort = require('toposort');
+
+export interface Chunk {
+  id: string;
+  files: string[];
+  parents?: Array<string | Chunk>;
+}
+
+/**
+ * A plugin that can sort the webpack files given there dependencies
```
simondel: their dependencies*

packages/stryker-webpack-transpiler/src/compiler/ChunkSorter.ts
```diff
@@ -0,0 +1,57 @@
...
+ * so that they can be loaded in the browser in the correct order.
```
simondel: You can remove the work *that*

packages/stryker-webpack-transpiler/src/compiler/ChunkSorter.ts
```diff
@@ -0,0 +1,57 @@
...
+ * 
+ * Given this dependency graph
+ *       A   D
+ *      / \ /
+ *     B   C
+ * 
+ * The order might be:
```
simondel: Input or output order?

packages/stryker-webpack-transpiler/src/compiler/ChunkSorter.ts
```diff
@@ -0,0 +1,57 @@
...
+    // We build a map (chunk-id -> chunk) for faster access during graph building.
+    const chunkMap: { [chunkId: string]: Chunk } = {};
+    allChunks.forEach((chunk) => {
+      chunkMap[chunk.id] = chunk;
+    });
+    const edges: [Chunk, Chunk][] = [];
+
+    allChunks.forEach(chunk => {
+      if (chunk.parents) {
+        // Add an edge for each parent (parent -> child)
```
simondel: This code may be clearer if you explain the *parent -> child* relationship when you create the edges array.

packages/stryker-webpack-transpiler/src/WebpackTranspiler.ts
```diff
@@ -0,0 +1,64 @@
...
+
+      this.webpackCompiler.writeFilesToFs(files);
+      return this.createSuccessResult(await this.webpackCompiler.emit());
+    } catch (err) {
+      return this.createErrorResult(`${err.name}: ${err.message}`);
+    }
+  }
+
+  private initialize(strykerWebpackConfig?: StrykerWebpackConfig): void {
+    strykerWebpackConfig = this.getStrykerWebpackConfig(strykerWebpackConfig);
```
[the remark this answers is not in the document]
nicojs: Agreed 👍

packages/stryker-webpack-transpiler/src/WebpackTranspiler.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.webpackCompiler = new WebpackCompiler(new ConfigLoader().load(strykerWebpackConfig));
```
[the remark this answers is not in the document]
nicojs: Well... it creates its own `log` property. Let's keep it as an instance method, once we migrate to log4js 2.x we might be able to refactor this.

packages/stryker-webpack-transpiler/src/compiler/ChunkSorter.ts
nicojs: :frowning:

packages/stryker-webpack-transpiler/src/compiler/ChunkSorter.ts
nicojs: Good call. I've changed it in this:

```ts
   // Build the *edges* of the graph. 
    // It will contain an entry per parent -> child relationship
    const edges: [Chunk, Chunk][] = [];
```

packages/stryker-webpack-transpiler/src/compiler/Webpack.ts
```diff
@@ -0,0 +1,2 @@
+import * as webpack from 'webpack';
```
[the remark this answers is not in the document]
nicojs: Yep. I've added a comment.

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
```diff
@@ -0,0 +1,94 @@
...
+          content: fileContent
+        });
+      } else {
+        files.push({
+          name: fileName,
+          kind: FileKind.Text,
+          mutated: true,
+          included: true,
+          transpiled: true,
+          content: fileContent.toString()
```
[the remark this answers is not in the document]
nicojs: Yeah. I've tested this and this is the case. JS/TS files are written to the output fs as a Buffer, we are responsible to call the `toString` on it.

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
```diff
@@ -0,0 +1,94 @@
...
+        });
+      }
+    });
+
+    const sortedFiles = new ChunkSorter()
```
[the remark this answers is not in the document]
nicojs: True. I'll create it static (easier to mock away than a function)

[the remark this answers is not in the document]
Archcry: Furthermore, do we really want to sort in the `OutputFileSystem`? I mean, the current implementation is tightly coupled with Stryker because it uses all kinds of Stryker descriptors (file creation). Wouldn't it be better to make the OutputFileSystem more generic and create / sort the files where it makes more sense?

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  private isBinary(fileName: string) {
+    return binaryFileExtensions.indexOf(path.extname(fileName)) >= 0;
+  }
+
+  public collectFiles(chunks: Chunk[]): Array<BinaryFile | TextFile> {
+    const files: Array<BinaryFile | TextFile> = [];
+    Object.keys(this._files).forEach(fileName => {
+      const fileContent = this._files[fileName];
```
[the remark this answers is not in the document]
nicojs: Well... If we create a `File[]` array, we would need to make sure that no duplicates are added. I rather use an object for that in this case.

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
```diff
@@ -0,0 +1,94 @@
...
+          included: true,
+          transpiled: true,
+          content: fileContent.toString()
+        });
+      }
+    });
+
+    const sortedFiles = new ChunkSorter()
+      .sortedFileNames(chunks);
+    files.sort((a, b) => {
```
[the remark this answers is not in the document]
nicojs: > Is this an alphanumeric sort

No, this is the order based on the index of the `sortedFiles` array (which is the topological order from the `ChunkSorter`).

```ts
files.sort((a, b) => {
      const aName = path.basename(a.name);
      const bName = path.basename(b.name);
      return sortedFiles.indexOf(aName) - sortedFiles.indexOf(bName);
    });
```

> But you could configure bundels to end up in different folders, right?

Yes, it could. And the ChunkSorter will be able to handle that.

nicojs:
@Archcry could you approve this PR? (you're review still says "requested changes")

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
```diff
@@ -0,0 +1,94 @@
...
+          mutated: true,
+          included: true,
+          transpiled: true,
+          content: fileContent.toString()
+        });
+      }
+    });
+
+    const sortedFiles = new ChunkSorter()
+      .sortedFileNames(chunks);
```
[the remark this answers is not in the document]
nicojs: A "chunk" in webpack terms is a piece of code that can be split up into multiple output files. The `ChunkSorter` can abstract these file names in the correct order.

If you don't think this is the correct place, where can we move it to?

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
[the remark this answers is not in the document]
nicojs: > Furthermore, do we really want to sort in the `OutputFileSystem`

Hmm good point. I'll move it to the `WebpackCompiler`.

packages/stryker-webpack-transpiler/src/fs/OutputFileSystem.ts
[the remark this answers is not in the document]
nicojs: I'll rename it to `FileSorter` which has one `sort` method that expects `File[]` and `Chunk[]`

nicojs:
Thanks for the reviews guys. I've addressed all feedback. Can you guys please have another look?

nicojs:
I just did a rebase. Now doing some final tweaks to the package.json. After that build is green I will merge

nicojs:
BTW it seems that BHC check is gone now, since rebase.

PR #596, opened 2018-01-26
fix(webpack-transpiler): treat font files as binary

nicojs:


simondel:
Feel free to merge when the build passes 👍

PR #601, opened 2018-02-01
 fix(webpack-transpiler): Correctly use in-memory files

### LINKED ISSUE #597: Module not found error when using the stryker-webpack-transpiler
### Summary

The InputFileSystem of the Stryker webpack transpiler does not mix-and-match correctly. 

See this line:
https://github.com/stryker-mutator/stryker/blob/9b86254265f04bf5793945f1798a326a62f996b5/packages/stryker-webpack-transpiler/src/fs/InputFileSystem.ts#L38-L40

When we create the path in-memory and later read from the directory we end up communicating less directories than exist on disk.

More info including reproduction can be found here: https://github.com/nicojs/angular-stryker-example/issues/1

nicojs:
Correctly use in-memory files. Instead of caching them ourselves, extend webpack's `CachedInputFileSystem` and rely on that for caching files. Still use memory-FS to load the in-memory files.

One downside is that I had to set the typescript target to "es2015", which means we have to drop support for node-4 on the stryker-webpack-transpiler package.

Fixes #597

### DIFF
--- a/packages/stryker-webpack-transpiler/src/fs/InputFileSystem.ts
+++ b/packages/stryker-webpack-transpiler/src/fs/InputFileSystem.ts
@@ -1,33 +1,34 @@
 import MemoryFS from './MemoryFS';
 import { webpack, Callback } from '../types';
 import * as fs from 'fs';
-const errors = require('errno');
 import { dirname } from 'path';
-import { getLogger } from 'log4js';
+import { CachedInputFileSystem, NodeJsInputFileSystem, Stats } from 'enhanced-resolve';
 
-export default class InputFileSystem implements webpack.InputFileSystem {
+// Cache duration is same as webpack has
+// => https://github.com/webpack/webpack/blob/efc576c8b744e7a015ab26f1f46932ba3ca7d4f1/lib/node/NodeEnvironmentPlugin.js#L14
+const CACHE_DURATION = 60000;
 
-  private log = getLogger(InputFileSystem.name);
+export default class InputFileSystem extends CachedInputFileSystem implements webpack.InputFileSystem {
+
   private memoryFS = new MemoryFS();
 
+  constructor(innerFS = new NodeJsInputFileSystem()) {
+    super(innerFS, CACHE_DURATION);
+  }
+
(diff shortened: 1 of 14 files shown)


packages/stryker-webpack-transpiler/src/fs/InputFileSystem.ts
```diff
@@ -1,125 +1,66 @@
...
-
-  public readlink(path: any, callback?: any) {
-    this.memoryFS.readlink(path, (err: Error, file: string) => {
-      if (err) {
-        fs.readlink(path, callback);
-      }
-      else {
-        callback(err, file);
+  public readFile(...args: any[]) {
+    const memoryFSArgs = args.filter((_, index) => index !== args.length - 1);
```
simondel: Why is this last item of the args so important?

packages/stryker-webpack-transpiler/src/fs/InputFileSystem.ts
```diff
@@ -1,125 +1,66 @@
...
+    memoryFSArgs.push((error: NodeJS.ErrnoException, content: any) => {
+      if (error) {
+        super.readFile.apply(this, args);
+      } else {
+        args.find((_, index) => index === args.length - 1)(error, content);
```
simondel: Cant we use `args[args.length - 1]`?

packages/stryker-webpack-transpiler/src/fs/InputFileSystem.ts
simondel: Can't we use `args.slice(0, args.length -1);`

nicojs:
Great review. Didn't know what I was thinking, fixing that tomorrow

Archcry:
Hmm, breaking compatibility with node4 seems reasonable since its LTS will end in April.

packages/stryker-webpack-transpiler/src/fs/InputFileSystem.ts
nicojs: Yes we can!

PR #603, opened 2018-02-01
build(integrationTest): Run integration tests in parallel

nicojs:


simondel:
This doesn't really seem to have a notable effect on the build times :/

nicojs:
I know. The problem is that the angular project has a lot of dependencies. We might be able to remove som and add a package-lock.json file. That would speed things up. However this is a good start i think

PR #604, opened 2018-02-03
refactor(webpack-transpiler): support new transpiler api

nicojs:
Throw an error when the webpack transpiler is asked to produce source maps (for now).


nicojs:
Merging this to fix the build (there are 2 problems i'm fixing in 2 separate PR's)

PR #614, opened 2018-02-07
fix(babel-transpiler): log babel config on debug

nicojs:
Log babel config on debug. Instead of logging only babel config from the `babelConfig` option, now log the resulting config from `babelrcFile` as well.


simondel:
LGTM! (If the build passes)

PR #627, opened 2018-02-23
docs(website): update stryker-mutator website url

nicojs:


simondel:
Looks good! We may want to consider using https instead of http

Issue #13, opened 2016-02-16
Exclude mutation types


simondel:
It should be possible to configure which mutation types are and which aren't used. The importance of this feature increases as the number of mutation types increases.

shybyte:
This feature request seems to be  a duplicate of #293.

shybyte:
I wonder if it makes sense to exclude/disable/enable mutation types by a system similar to the rules configuration of tslint (See https://palantir.github.io/tslint/usage/configuration/)

Example:
```js
{
  mutations: {
    StringLiteral: false,
    Block: false
  }
}
```
Th advantage of this system
* Familiar to tslint users
* Can be easily extended in the future to configure rules in a more complex way. This can be useful in cases where people disagree about what should be how mutated (Example issue #465)


Example:
```js
{
  mutations: {
    StringLiteral: {
      relaceExportedConst: false,
      replacements: ["", "<script>window.alert('Evil Hack!')</script>"]
    },
  }
}
```

shybyte:
@simondel  @nicojs  Is there any ongoing work in the direction of this ticket? Otherwise I would like to volunteer to implement it, at least for the TypescriptMutator. For me it seems that in the current design, the class TypescriptMutator is responsible for knowing all typescript "child" mutators and for executing them. 
Therefore I think the best place to disable/enable/configure the "child" mutators is the constructor of https://github.com/stryker-mutator/stryker/blob/5dd8be70f0ec53152837abae6145214b9d06d364/packages/stryker-typescript/src/TypescriptMutator.ts#L24

In the long run, it might be better to do this work on a higher level, but  I'm particular interested in TypeScript and it could be a good start.

simondel:
There are plans to implement this, but it's not at the top of our todo-list. I would be a fan of working with a blacklist that you could set in your config. That way when we release a new mutation type, you will actually start using it.

Perhaps we could simply work with an array property that is added to the config:
```
excludedMutations: ['ArrayNewExpression', 'PrefixUnaryExpression'],
```
The names of the excluded mutations should be exactly the same (case and all) with the reported value:
https://github.com/stryker-mutator/stryker/blob/06d6bace999265ac6a8d0ae98ba6f6413cc3a1e0/packages/stryker-api/src/report/MutantResult.ts#L6
That way you could always look at your report and decide what you want to exclude.

The constructor you found is used for test purposes. A more solid option might be to filter the `NodeMutator[]` with the list in the config when the `mutate` function is called.

[...]

PR #652, opened 2018-03-14
feat(stryker): Add excludedMutations as a config option (#13)

hlbowen:
This adds the ability to exclude specific mutation types ('BooleanSubstitution', 'ArrayLiteral', etc.) from being performed during the test run. Can be used via either command-line or config file. Defaults to an empty array, meaning all mutation types are used.

### DIFF
--- a/packages/stryker-api/src/core/StrykerOptions.ts
+++ b/packages/stryker-api/src/core/StrykerOptions.ts
@@ -54,10 +54,10 @@ interface StrykerOptions {
    * This is often dependent on the language of your source files.
    *
    * This value can be either a string, or an object with 2 properties:
-   * * `string`: The name of the mutant generator to use. For example: 'es5', 'typescript'
+   * * `string`: The name of the mutant generator to use. For example: 'javascript', 'typescript'
    * * { name: 'name', excludedMutations: ['mutationType1', 'mutationType2'] }:
    *    * The `name` property is mandatory and contains the name of the mutant generator to use.
-   *    * For example: 'es5', 'typescript'
+   *    * For example: 'javascript', 'typescript'
    *    * The `excludedMutations` property is mandatory and contains the names of the specific mutation types to exclude from testing.
    *    * The values must match the given names of the mutations. For example: 'BinaryExpression', 'BooleanSubstitution', etc.
    */
--- a/packages/stryker-api/src/core/StrykerOptions.ts
+++ b/packages/stryker-api/src/core/StrykerOptions.ts
@@ -56,6 +56,13 @@ interface StrykerOptions {
   mutator?: string;
 
   /**
+   * The names of the specific mutation types to exclude from testing.
+   * The names must match the given names of the mutations.
+   * For example: 'BinaryExpression', 'BooleanSubstitution', etc.
+   */
+  excludedMutations?: string[];
+
+  /**
    * The names of the transpilers to use (in order). Default: [].
    * A transpiler in this context is a plugin that can transform input files (source code)
    * before testing.
--- a/packages/stryker-api/src/core/StrykerOptions.ts
+++ b/packages/stryker-api/src/core/StrykerOptions.ts
@@ -1,5 +1,6 @@
 import InputFileDescriptor from './InputFileDescriptor';
 import MutationScoreThresholds from './MutationScoreThresholds';
+import MutatorDescriptor from './MutatorDescriptor';
 
 interface StrykerOptions {
   // this ensures that custom config for for example 'karma' can be added under the 'karma' key
@@ -49,18 +50,18 @@ interface StrykerOptions {
   testRunner?: string;
 
   /**
-   * The name of the mutant generator to use to generate mutants based on your input file. 
+   * The mutant generator to use to generate mutants based on your input file.
    * This is often dependent on the language of your source files.
-   * For example: 'es5', 'typescript'
+   *
+   * This value can be either a string, or an object with 2 properties:
+   * * `string`: The name of the mutant generator to use. For example: 'es5', 'typescript'
+   * * { name: 'name', excludedMutations: ['mutationType1', 'mutationType2'] }:
+   *    * The `name` property is mandatory and contains the name of the mutant generator to use.
+   *    * For example: 'es5', 'typescript'
+   *    * The `excludedMutations` property is mandatory and contains the names of the specific mutation types to exclude from testing.
+   *    * The values must match the given names of the mutations. For example: 'BinaryExpression', 'BooleanSubstitution', etc.
    */
-  mutator?: string;
-
-  /**
-   * The names of the specific mutation types to exclude from testing.
-   * The names must match the given names of the mutations.
-   * For example: 'BinaryExpression', 'BooleanSubstitution', etc.
-   */
-  excludedMutations?: string[];
+  mutator?: string | MutatorDescriptor;
 
   /**
    * The names of the transpilers to use (in order). Default: [].
--- a/packages/stryker/src/ConfigValidator.ts
+++ b/packages/stryker/src/ConfigValidator.ts
@@ -22,6 +22,7 @@ export default class ConfigValidator {
     this.validateIsStringArray('plugins', this.strykerConfig.plugins);
     this.validateIsStringArray('reporter', this.strykerConfig.reporter);
     this.validateIsStringArray('transpilers', this.strykerConfig.transpilers);
+    this.validateIsStringArray('excludedMutations', this.strykerConfig.excludedMutations);
     this.validateCoverageAnalysis();
     this.validateCoverageAnalysisWithRespectToTranspilers();
     this.crashIfNeeded();
--- a/packages/stryker/src/ConfigValidator.ts
+++ b/packages/stryker/src/ConfigValidator.ts
@@ -1,5 +1,6 @@
+import * as _ from 'lodash';
 import { TestFramework } from 'stryker-api/test_framework';
-import { MutationScoreThresholds } from 'stryker-api/core';
+import { MutatorDescriptor, MutationScoreThresholds } from 'stryker-api/core';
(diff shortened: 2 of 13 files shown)


hlbowen:
Hi there! I picked up this particular feature request because my current work team is using Stryker, and being able to limit the mutation types would be very helpful for us :)

I tried to follow the approach proposed in the comments of issue #13 as best as possible. It seemed easier to implement a separate config option ('excludedMutations') for the time being, rather than modifying the existing 'mutator' option; less chance of screwing something up. There's no reporting on the excluded mutations yet, but that could be added in future.

Open to feedback & suggestions!

packages/stryker/test/helpers/producers.ts
```diff
@@ -236,5 +236,14 @@ export const testableMutant = (fileName = 'file') => new TestableMutant('1337',
   textFile({ name: fileName, content: 'const a = 4 + 5' })
 ));
 
+export const excludedTestableMutant = (fileName = 'file') => new TestableMutant('2448', mutant({
```
nicojs: I would rather add the parameter `mutatorName` to the `testableMutant` factory method. That way, it is really clear from the test code what is happening.

packages/stryker/src/Stryker.ts
```diff
@@ -68,14 +69,15 @@ export default class Stryker {
 
   private mutate(inputFiles: File[], initialTestRunResult: InitialTestRunResult) {
     const mutator = new MutatorFacade(this.config);
-    const mutants = mutator.mutate(inputFiles);
-    if (mutants.length) {
-      this.log.info(`${mutants.length} Mutant(s) generated`);
+    const allMutants = mutator.mutate(inputFiles);
+    const includedMutants = this.removeExcludedMutants(allMutants);
+    if (includedMutants.length) {
+      this.log.info(`${includedMutants.length} Mutant(s) generated`);
```
nicojs: Could we change this log message to include the number of excluded mutants? 

```ts
const numberExcluded = allMutants.length - includedMutants.length;
this.log.info(`${includedMutants.length} Mutant(s) generated${numberExcluded ? ' (${numberExcluded} mutant(s) excluded)' : '' }`);
```

Same could be added to the mutant-free world message.

packages/stryker/test/unit/StrykerSpec.ts
```diff
@@ -163,7 +166,8 @@ describe('Stryker', function () {
     });
 
     it('should log to have quit early if no mutants were generated', async () => {
```
nicojs: I personally think it would be clearer if we add a separate unit test for the exclusion of mutations. So keep the original test and add one where you explicitly exclude a few. That way it would also be easier to test the new log message.

nicojs:
@simondel what do you think about the `excludedMutations` name? I think you suggested it in the issue as well. 

Go with that name and have a possible breaking change later? Or update this PR to use the other config method?

simondel:
I'd prefer implementing it with an object straight away. We could help with that. 

Do you want the stryker config to default to:
```js
mutator: {
   name: ${configuredMutator},
   excludedMutations: [/* Array with the names of mutation types (as strings) that should be excluded */]
}
```
for all new projects?

nicojs:
> Do you want the stryker config to default to ..

No I would keep them both, allowing for the short hand version as well as the longer version. We shouldn't over complicate the configuration if people don't want to use it. 

@hlbowen do you think you can handle these changes? Or shall we do a PR on your branch to help you out a bit?

simondel:
The reason why I'd prefer to default to the more complicated setup for new projects is to make people aware of the fact that the setting exists

dennythecoder:
I've only dabbled with Stryker, but I think the better option might be to improve documentation on the matter.  Increased required  complexity can be outputting for folks looking at mutation testing for the first time.  Just my thoughts 😅 totally dig the work you're doing

hlbowen:
Hi folks, I believe I can handle the changes you requested! One point of clarification though. It sounds like you want the default config to be:

```
mutator: {
   name: "${configuredMutator}",
   excludedMutations: [/* Array with the names of mutation types (as strings) that should be excluded */]
}
```

but to also continue handling the following:

`mutator: "${configuredMutator}"`

and to get rid of the top-level `excludedMutations` standalone option. Is this correct? How would you want to implement this option on the command-line, if at all?

Thanks so much for the feedback!

simondel:
Let's stick with _less is more_!
@dennythecoder Thanks for dropping by! This is one the reasons why I love open source. Users can provide feedback when they want to instead of when we explicitly ask for it/deploy software ❤️ 

----
As I understand it the plan for now is:
By default the `mutator` property will be a string indicating which plugin should be loaded. All mutations will be enabled with this configuration.

If someone wants to change which mutations are enabled they can do so by specifying the `mutator` property as an object such as:
```js
mutator: {
   name: "typescript",
   excludedMutations: ["ArrayLiteral", "BinaryExpression"]
}
```

There will be no top-level `excludedMutations` property and you cannot exclude mutations from the commandline (only from config file).

Does everyone agree? @nicojs @hlbowen @dennythecoder @Archcry

Archcry:
Yes, I agree, I think this groups the exclusions nicely with the mutator that is used.

hlbowen:
Okay, I think all the requested changes have been accounted for and this feature is ready. Take a look and let me know what you think!

nicojs:
I'll have a look within 2 days, thanks for all your work so far!

packages/stryker-api/src/core/StrykerOptions.ts
```diff
@@ -49,18 +50,18 @@ interface StrykerOptions {
...
-   * The names must match the given names of the mutations.
-   * For example: 'BinaryExpression', 'BooleanSubstitution', etc.
-   */
-  excludedMutations?: string[];
+   *
+   * This value can be either a string, or an object with 2 properties:
+   * * `string`: The name of the mutant generator to use. For example: 'es5', 'typescript'
+   * * { name: 'name', excludedMutations: ['mutationType1', 'mutationType2'] }:
+   *    * The `name` property is mandatory and contains the name of the mutant generator to use.
+   *    * For example: 'es5', 'typescript'
```
nicojs: `'es5'` is actually deprecated. The new one is `'javascript'` (babel based)

packages/stryker/src/ConfigValidator.ts
```diff
@@ -34,6 +34,21 @@ export default class ConfigValidator {
     }
   }
 
+  private validateMutator() {
+    const mutator = this.strykerConfig.mutator;
+    if (typeof mutator !== 'string' && !_.isObject(mutator)) {
+      // console.log(`***** FAILED at first assertion: Mutator is: ${mutator}`);
```
nicojs: Can we remove these comments?

packages/stryker/src/ConfigValidator.ts
```diff
@@ -34,6 +34,21 @@ export default class ConfigValidator {
...
+      this.invalidate(`Value "${mutator}" is invalid for \`mutator\`. Expected either a string or an object`);
+    }
+    if (_.isObject(mutator)) {
```
nicojs: Please use `typeof mutator === 'object'` instead. We don't want to depend on lodash for these simple assertions.

packages/stryker/src/ConfigValidator.ts
```diff
@@ -34,6 +34,21 @@ export default class ConfigValidator {
     }
   }
 
+  private validateMutator() {
+    const mutator = this.strykerConfig.mutator;
+    if (typeof mutator !== 'string' && !_.isObject(mutator)) {
```
nicojs: This `if` can be rewritten as:

```ts
const mutatorType = typeof mutator;
if(mutatorType === 'object') {
 // object validation
} else if(mutatorType !== 'string'){
 // Invalidate because not supported.
}
```

packages/stryker/src/MutatorFacade.ts
```diff
@@ -12,7 +12,15 @@ export default class MutatorFacade implements Mutator {
...
-      .create(this.config.mutator, this.config)
+      .create(this.getMutatorName(this.config.mutator), this.config)
       .mutate(inputFiles);
   }
+
+  private getMutatorName(mutator: string | MutatorDescriptor) {
+    if (typeof mutator === 'string') {
+      return mutator;
+    } else {
+      return (mutator as MutatorDescriptor).name;
```
nicojs: This cast is not needed and can be removed. TypeScript's awesome type inference has your back.

packages/stryker/src/Stryker.ts
```diff
@@ -98,7 +98,11 @@ export default class Stryker {
   }
 
   private removeExcludedMutants(mutants: Mutant[]): Mutant[] {
-    return mutants.filter(mutant => this.config.excludedMutations.indexOf(mutant.mutatorName) === -1);
+    if (typeof this.config.mutator === 'string') {
+      return mutants;
+    }
+    const mutatorDescriptor = this.config.mutator as MutatorDescriptor;
+    return mutants.filter(mutant => mutatorDescriptor.excludedMutations.indexOf(mutant.mutatorName) === -1);
```
nicojs: Please move this block in an "else". That way it is immediately clear that this method has 2 branches and it is not an oversight of the developer.

hlbowen:
Thanks so much for the feedback, @nicojs! Requested changes have been added to the PR.

Also, I feel it is worth pointing out that if the commits on this PR are squashed before merging, there should be NO breaking changes. The old config functionality remains intact and users will simply have a new option of using an object for the mutator. So this feature could still go into release 1.0.0, but it doesn't have to wait for that. :)

nicojs:
> So this feature could still go into release 1.0.0, but it doesn't have to wait for that. :)

That's what we figured :shipit: . We generally don't want to wait for releases, we just release whenever new functionality is available. 

We just released v0.20.1 with this feature. You'll also need to update all plugins as our release plugin also did a major release of `stryker-api` for the new functionality.

PR #664, opened 2018-03-23
Feat(Webpack): Add Webpack 4 zero config support

Archcry:
Makes the Stryker Webpack Transpiler plugin compatible with the new version of webpack (4). Also adds support for some of the zero config features in Webpack 4.


packages/stryker-webpack-transpiler/test/integration/transpiler.it.ts
```diff
@@ -34,13 +30,33 @@ describe('Webpack transpiler', () => {
...
     const files = readFiles();
+
     const transpiledFiles = await sut.transpile(files);
+
     expect(transpiledFiles.error).null;
     expect(transpiledFiles.outputFiles).lengthOf(1);
   });
 
+  it('should be able to transpile "zeroConfig" sample without a Webpack config file', async function() {
+    this.timeout(10000);
```
nicojs: we should set this timeout for all tests instead of per-spec.

packages/stryker-webpack-transpiler/testResources/zeroConfig/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "scripts": {
+    "build": "webpack --mode development"
+  },
+  "keywords": [
+    "webpack 4"
+  ],
+  "author": "Sander Koenders",
+  "license": "MIT",
+  "devDependencies": {
+    "webpack": "^4.1.1",
```
nicojs: do we need dependencies here?

packages/stryker-webpack-transpiler/src/WebpackTranspiler.ts
```diff
@@ -4,7 +4,7 @@ import WebpackCompiler from './compiler/WebpackCompiler';
 import ConfigLoader from './compiler/ConfigLoader';
 import { setGlobalLogLevel } from 'log4js';
 
-const DEFAULT_STRYKER_WEBPACK_CONFIG = Object.freeze({ configFile: 'webpack.config.js', silent: true });
+const DEFAULT_STRYKER_WEBPACK_CONFIG = Object.freeze({ configFile: 'webpack.config.js', silent: true, context: process.cwd() });
 
 export default class WebpackTranspiler implements Transpiler {
```
nicojs: I think we should still crash is we configure a config file location to a file that does not exist + integration test

packages/stryker-webpack-transpiler/testResources/zeroConfig/package.json
Archcry: Ofcourse we do, more dependencies is always better!

Issue #650, opened 2018-03-09
Projectroot in stryker configuration object


Archcry:
Currently we are using `process.cwd()` in a lot of plugins to resolve the current working directory in some of our stryker packages. This is undesirable in tests we sometimes want to mock the current working directory to "pretend" we're somewhere else. The only way to do this at the moment is to mock `process.cwd()` which can cause side effects. It would be better to define a `projectRoot` or `base` property on the Stryker config object that points to the current working directory. This way we can make our own Stryker config with a `projectRoot` or `base` that points somewhere else without harming `process.cwd()`.

nicojs:
How about `basePath`? (same as Karma)

A list of components that should use this property:

* stryker: input file resolver and mutate property
* karma-runner: use to locate karma configuration
* typescript-transpiler: locate the tsconfig file
* babel-transpiler: use to locate babelrc
* webpack-transpiler: use to locate wepack config file.
* mocha-runner: locate mocha.opts file

Let's first finish #589

Archcry:
> How about `basePath`? (same as Karma)

Sounds good

nicojs:
OK, the way it should work:

`basePath` will always be set by stryker and defaults to `.` (current directory). 

If `basePath` is set in stryker.conf.js: that basePath is used, relative to the stryker.conf.js file. For example: `path.resolve(path.dirname('path/to/stryker.conf.js'), options.basePath)`

This should also solve #245

nicojs:
I think this also solves #137

simondel:
I'll get to work on this 👍

simondel:
As discussed we won't do this right now.

PR #665, opened 2018-03-23
feat(identify files): automatically detect files using git

### LINKED ISSUE #589: Feature: use git to identify files (instead of the files array)
## Current problem

Right now, a lot of reported issues come from people misunderstanding what the `files` array does in your stryker.conf.js files. And who can blame them. It's kind of a mess.

Here is a list of the most common issues:

1. "Why do I need to specify `files`?" 
2. "Why do I need to put `!**.*` as the first argument?" (psst, both `stryker-typescript` and `stryker-karma-runner` expand the files array even before your config is loaded, resulting in a mess. I know, complicated indeed!)
3. "Why do i need to add *insert-file-name-here* to my `files`?"
4. "What is the difference between `mutate` and `files`?"

We can do better than this.

## Feature request: use git

Let's replace the current functionality with using the project's `.gitignore` file by default. This is much more reliable and user-friendly than any custom solution we can come up with.  ~For users that still need to specify the files, we can use a `.strykerignore` file construction. It will work the same way as `.gitignore` works, so no need to learn something new.~ EDIT: We won't use a `.strykerignore` file, instead we still support a `files` array with strings for which files to include.

### Way of working

The new way of working will be:

1. Use `files` array if present.
2. If not: use `.gitignore` file. Use the  output of the `git ls-files --others --exclude-standard --cached`  command to get the list without having to parse the .gitignore files ourselves.
3. ~If not in a git repo (or git isn't installed), use a sane default (ignore only `node_modules` or something).~ EDIT: We shouldn't use a default. If we did that, the same project on different machines would result in different files being used and different overall result. Instead, we should throw an error which clearly specifies that the input files could not be specified

### Advantages:

This way of working has a lot of advantages

1. It will be zero config for most users. Users don't even have to know about sandboxes or anything.
2. Marking files to be mutated will be easier, as you can only do that by specifying them in the `mutate` array. No more mix and matching by combining the `files` and `mutate` array. 

### Downsides

We have some downsides to this new approach.

1. It is a major (breaking) change
2. Sandboxes will be bigger than they would need to be. This will not be a big problem I think.
3. We will lose support for additional file properties you can now specify with your files: `transpiled`, `included` and `mutated`. We can mitigate this:
  * `mutated`: This one is simple: use the `mutate` property instead.
  * `included`: This should move to the test runner configuration (with sane defaults). This might give some problems with both the mocha and karma test runners.
  * `transpiled`: This should be moved to the transpiler configuration. Not a big deal, as all transpilers have their own config anyway (webpack.config.js, tsconfig, .babelrc, etc)

Todo list:

Work on this feature is done in the [feat/identify-files branch](https://github.com/stryker-mutator/stryker/tree/feat/identify-files) 

- [x] stryker-api: Remove the `FileDescriptor`, `TextFile`, `WebFile` and `BinaryFile` interfaces. Add the `File` class which represents a binary or text file.
- [x] stryker: add "downgrade" from old `InputFileDescriptor` objects to strings with a deprecation message.
- [x] stryker: Don't input file resolver: don't read deleted files
- [x] stryker: Support coverage analysis of typescript projects
- [x] stryker: Update the `InputFileResolver` class to now use the new way of identifying files
- [x] stryker: Update the integration with `Transpiler` and `TestRunner`.
- [x] stryker: Serialize/deserialize the new `File` class instance using [surrial](https://www.npmjs.com/package/surrial)
- [x] stryker-typescript: Remove the feature of magically adding `files` from the `TypescriptConfigEditor`
- [x] stryker-typescript: Replace the feature of using the `transpiled` property, instead use the files defined in the `tsconfig.json` file directly
- [x] stryker-karma-runner: Remove the feature of magically adding `files` from the `StrykerKarmaEditor`
- [x] stryker-karma-runner: Add functionality of using the `files` array from the `karma.conf.js` file.
- [x] stryker-karma-runner: Make sure the `testHooks` is always added (first file).
- [x] stryker-babel-transpiler: Don't use the `transpiled` property anymore
- [x] stryker-mocha-runner: Don't use the `included` property anymore: instead use the mocha opts (or default to `test` folder as is the default for mocha)
- [x] stryker-mocha-runner: Make sure the `testHooks` is always evaluated.
- [x] stryker-mocha-runner: Fix this issue: 
   ```bash
    [2018-03-23 10:00:58.113] [TRACE] TestRunnerChildProcessAdapter - [2018-03-23 10:00:58.114] [DEBUG] IsolatedTestRunnerAdapterWorker - UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: No files discovered (tried pattern(s) [
      "test/*.js"
    ]). Please specify the files (glob patterns) containing your tests in mochaOptions.files in your stryker.conf.js file.
   ```
- [x] stryker-javascript-mutator: Implement new api.
- [x] stryker-webpack-transpiler: Implement new api.
- [x] documentation: Document the new way of working.
- [x] update readme's
- [ ] update `initStrykerConfig` keys in package.json's

nicojs:
Fixes #589


packages/stryker/src/input/InputFileResolver.ts
```diff
@@ -0,0 +1,174 @@
+import * as path from 'path';
```
Archcry: This file should be looked at, contains more than one class and is hard to understand.

packages/stryker/src/utils/objectUtils.ts
```diff
@@ -88,4 +73,8 @@ export function setExitCode(n: number) {
 
 export function base64Decode(base64EncodedString: string) {
   return Buffer.from(base64EncodedString, 'base64').toString('utf8');
+}
+
+export function normalizeWhiteSpaces(str: string) {
```
Archcry: What does this function do?

packages/stryker-api/src/core/File.ts
```diff
@@ -1,29 +1,55 @@
...
- 
-export interface BinaryFile extends FileDescriptor {
-  content: Buffer;
-  kind: FileKind.Binary;
-}
\ No newline at end of file
+  /**
+   * Sets the underlying content as buffer.
+   */
+  set content(value: Buffer) {
```
Archcry: File class should become immutable. Also applies for the `set textContent` function.

packages/stryker-babel-transpiler/src/BabelConfigReader.ts
```diff
@@ -8,19 +8,20 @@ export default class BabelConfigReader {
...
-  private getConfigFile(config: Config) {
+  private getConfigFile(config: Config): babel.TransformOptions | null {
     if (typeof config[CONFIG_KEY_FILE] === 'string') {
       const babelrcPath = path.resolve(config[CONFIG_KEY_FILE]);
       this.log.info(`Reading .babelrc file from path "${babelrcPath}"`);
       if (fs.existsSync(babelrcPath)) {
         try {
-          let configFile = fs.readFileSync(babelrcPath, 'utf8');
-          return JSON.parse(configFile);
+          const configFileContent = fs.readFileSync(babelrcPath, 'utf8');
```
Archcry: We could inline this

packages/stryker-babel-transpiler/src/BabelTranspiler.ts
```diff
@@ -1,69 +1,85 @@
...
+      return file;
+    } else if (typeof result.code === 'undefined') {
+      throw new Error(`Could not transpile file "${file.name}". Babel transform function delivered \`undefined\`.`);
+    } else {
+      return new File(toJSFileName(file.name), result.code);
     }
+  }
 
-    return result;
+  private determineProjectRoot(options: TranspilerOptions): string {
```
Archcry: Should replace this with `config.basePath` when it's available https://github.com/stryker-mutator/stryker/issues/650

packages/stryker-karma-runner/src/TestHooksMiddleware.ts
```diff
@@ -0,0 +1,25 @@
...
+import * as url from 'url';
+
+export const TEST_HOOKS_FILE_NAME = '__test_hooks_for_stryker__.js';
+
+export default class TestHooksMiddleware {
+
+  public currentTestHooks: string = '';
+
+  public handler(): RequestHandler {
+    const self = this;
```
Archcry: Is this really needed? We're using a arrow function so the context of `this` should still be the same.

packages/stryker-karma-runner/src/utils.ts
```diff
@@ -1,3 +1,18 @@
...
   return require(name);
+}
+
+/**
+ * Creates an empty file if it did not exist
+ * @param fileName the name of the file
+ */
+export function touchSync(fileName: string) {
+  try {
+    const handle = fs.openSync(fileName, 'wx');
```
Archcry: Do we need the file to be executable? If not `w` should be enough.

Archcry: We can inline this.

packages/stryker-mocha-runner/package.json
```diff
@@ -35,10 +35,12 @@
   },
   "homepage": "https://github.com/stryker-mutator/stryker/tree/master/packages/stryker-mocha-runner#readme",
   "dependencies": {
+    "fast-glob": "^2.1.0",
```
Archcry: We might want to find an alternative for this in the future because it has a lot of dependencies.

packages/stryker-mocha-runner/src/MochaOptionsLoader.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private parseOptsFile(optsFileContent: string): MochaRunnerOptions {
+    const options = optsFileContent.split('\n').map(val => val.trim());
+    const mochaRunnerOptions: MochaRunnerOptions = Object.create(null);
+    options.forEach(option => {
+      const args = option.split(' ').filter(Boolean);
+      if (args[0]) {
+        switch (args[0]) {
+          case '--require':
```
Archcry: Doesn't `case '--require' || '-r':` work?

packages/stryker-mocha-runner/src/MochaTestRunner.ts
```diff
@@ -82,7 +95,7 @@ export default class MochaTestRunner extends EventEmitter implements TestRunner
   }
 
   private additionalRequires() {
-    if (this.mochaRunnerOptions && this.mochaRunnerOptions.require) {
```
Archcry: Are we sure `this.mochaRunnerOptions` is a valid object here instead of undefined?

packages/stryker-mocha-runner/src/MochaTestRunner.ts
```diff
@@ -1,36 +1,48 @@
...
+export default class MochaTestRunner implements TestRunner {
+  private fileNames: string[];
   private log = getLogger(MochaTestRunner.name);
-  private mochaRunnerOptions: MochaRunnerOptions | undefined;
+  private mochaRunnerOptions: MochaRunnerOptions;
 
   constructor(runnerOptions: RunnerOptions) {
-    super();
-    this.files = runnerOptions.files;
     this.mochaRunnerOptions = runnerOptions.strykerOptions[mochaOptionsKey];
```
Archcry: `runnerOptions.strykerOptions[mochaOptionsKey]` could be undefined.

packages/stryker-typescript/src/transpiler/TranspileFilter.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Represents a transpile filter. This is the component that decides on which files needs to be transpiled.
+ * 
+ * It is implemented using the composite pattern.
+ * If there is a tsConfig, that will be used. If not, a default is used (transpile all TS-like files)
+ */
+export default abstract class TranspileFilter {
+  public abstract isIncluded(fileName: string): boolean;
+
+  public filterIsIncluded(files: File[]) {
```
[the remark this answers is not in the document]
nicojs: Yeah, i was kind of unhappy with them. Thanks, i've removed it.

packages/stryker/src/child-proxy/ChildProcessProxy.ts
```diff
@@ -44,7 +45,8 @@ export default class ChildProcessProxy<T> {
   }
 
   private send(message: WorkerMessage) {
-    this.worker.send(serialize(message));
+    const messageStr = serialize(message);
```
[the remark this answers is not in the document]
nicojs: Yeah, probably used that to inspect the variable during debugging. I've inlined it again.

packages/stryker/src/utils/objectUtils.ts
nicojs: It consolidates multiple consecutive white spaces into a single space. I'll add a comment.

packages/stryker-babel-transpiler/src/BabelTranspiler.ts
nicojs: True, that's why the babel transpiler is in the list.

packages/stryker-mocha-runner/package.json
nicojs: Yeah, but let's use this for now. We can experiment later.

packages/stryker-mocha-runner/src/MochaOptionsLoader.ts
nicojs: Nope. Apparently not (tried it)

packages/stryker-mocha-runner/src/MochaTestRunner.ts
nicojs: Yes, because we know all config editors are loaded and the mocha config editor does this:

`config[mochaOptionsKey] = new MochaOptionsLoader().load(config);`

It is not ideal, but at the same time, i don't want to keep rechecking that variable to be defined.

nicojs:
@Archcry I've refacted / updated all you're remarks. I will work on the last todo's tonight/this weekend. Let's keep the ball rolling!

simondel:
There are some parts of Stryker's README that are a bit off, the formatting near **Mandatory:** is a bit off.
image
Also, it still mentions the InputFileDescriptor which is gone, right? And there is an odd **+** at the end

simondel:
The typescript plugin mentions it changes the files array, but I believe that has been removed from the code: https://github.com/stryker-mutator/stryker/tree/feat/identify-files/packages/stryker-typescript#typescriptconfigeditor

packages/stryker-mocha-runner/src/MochaTestRunner.ts
```diff
@@ -1,36 +1,53 @@
...
-import { FileDescriptor } from 'stryker-api/core';
 import LibWrapper from './LibWrapper';
 import StrykerMochaReporter from './StrykerMochaReporter';
 import MochaRunnerOptions, { mochaOptionsKey } from './MochaRunnerOptions';
+import RequireCacheRecorder from './RequireCacheRecorder';
+import { evalGlobal } from './utils';
 
-export default class MochaTestRunner extends EventEmitter implements TestRunner {
-  private files: FileDescriptor[];
+const DEFAULT_TEST_PATTERN = 'test/*.js';
```
simondel: We should describe this in the README

simondel: As well as how we use `this.mochaRunnerOptions.files`

packages/stryker-mocha-runner/src/MochaTestRunner.ts
```diff
@@ -1,36 +1,53 @@
...
+    setGlobalLogLevel(runnerOptions.strykerOptions.logLevel || 'info');
     this.mochaRunnerOptions = runnerOptions.strykerOptions[mochaOptionsKey];
     this.additionalRequires();
   }
 
-  private purgeFiles() {
-    this.files.forEach(f => delete require.cache[f.name]);
+  init(): Promise<void> {
+    const globPatterns = this.mochaRunnerOptions.files || [DEFAULT_TEST_PATTERN];
+    return LibWrapper.glob(globPatterns)
```
simondel: Why do we wrap glob?

nicojs:
@simondel how do you want to merge this in master? A squash commit is not helpful I think (as we would only have 1 note in the changelog). I was thinking on rebasing it and do 1 commit per package. That way we can outline exactly what changed in that package. Most packages will have breaking changes, so we need to manually release this.

packages/stryker/src/Stryker.ts
```diff
@@ -50,33 +50,34 @@ export default class Stryker {
   async runMutationTest(): Promise<MutantResult[]> {
     this.timer.reset();
     const inputFiles = await new InputFileResolver(this.config.mutate, this.config.files, this.reporter).resolve();
```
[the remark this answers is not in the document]
nicojs: I didn't want to change this functionality too much, but I'm thinking about removing it entirely in favor of something like [`fast-glob`](https://www.npmjs.com/package/fast-glob). So, unless you have remarks on the new code, let's leave the way of working like it is for now.

packages/stryker/src/input/InputFileCollection.ts
```diff
@@ -0,0 +1,31 @@
+import { File } from 'stryker-api/core';
+import { Logger } from 'log4js';
+import { normalizeWhiteSpaces } from '../utils/objectUtils';
+
+export default class InputFileCollection {
+  public readonly filesToMutate: ReadonlyArray<File>;
+  public readonly mutateFileNames: ReadonlyArray<string>;
+
+  constructor(public readonly files: ReadonlyArray<File>, mutateGlobResult: ReadonlyArray<string>) {
+    this.mutateFileNames = mutateGlobResult.filter(mutateFile => files.some(file => file.name === mutateFile));
```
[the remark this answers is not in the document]
nicojs: True. It was just there for convenience but only used twice. I'll remove it.

packages/stryker/src/process/InitialTestExecutor.ts
```diff
@@ -17,46 +20,50 @@ const INITIAL_RUN_TIMEOUT = 60 * 1000 * 5;
...
-      await sandbox.dispose();
-      return {
-        runResult,
-        transpiledFiles: transpileResult.outputFiles,
-        coverageMaps: coverageInstrumenterTranspiler.fileCoverageMaps
-      };
-    }
+    const transpiledFiles = await this.transpileInputFiles();
+    const sourceMapper = SourceMapper.create(transpiledFiles, this.options);
+    const { fileCoverageMaps, instrumentedFiles } = await this.annotateForCodeCoverage(transpiledFiles, sourceMapper);
```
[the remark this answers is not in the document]
nicojs: Good point. I'll be happy to take naming suggestions. For the time being I'll add some comments.

packages/stryker-mocha-runner/src/MochaTestRunner.ts
nicojs: Yes! I've rewritten the readme. I've taken a page from the [configuring jest](https://facebook.github.io/jest/docs/en/configuration.html) layout, let me know what you think.

nicojs:
@simondel Thanks for your review comments! I've refactored the code. Could you please take another look?

simondel:
@nicojs I'll take another look at it :) I'm fine with creating one merge commit. We should also consider a blog post.

packages/stryker-mocha-runner/README.md
```diff
@@ -37,15 +37,53 @@ module.exports = function (config) {
...
             asyncOnly: false
         }
     });
 }
 ```
 
-You can pass mocha options in 2 ways: either specify a mocha.opts file or configure `mochaOptions` in stryker.conf.js. When both are specified, the values in stryker.conf.js will overrule the values in your mocha.opts file.
+### `mochaOptions.files` [`string` or `string[]`]
+
+Default: `'test/*.js'`
```
simondel: Wait, shouldn't this be `test/**/*.js`? Or does the globbing implementation work differently?

packages/stryker/src/input/InputFileCollection.ts
```diff
@@ -17,14 +17,14 @@ export default class InputFileCollection {
...
-        log.info(`Found ${this.mutateFileNames.length} of ${this.files.length} file(s) to be mutated.`);
+      if (this.filesToMutate.length) {
+        log.info(`Found ${this.filesToMutate.length} of ${this.files.length} file(s) to be mutated.`);
       } else {
         log.warn(`No files marked to be mutated, stryker will perform a dry-run without actually mutating anything.`);
       }
       if (log.isDebugEnabled) {
         log.debug(`All input files: ${JSON.stringify(this.files.map(file => file.name), null, 2)}`);
-        log.debug(`Files to mutate: ${JSON.stringify(this.mutateFileNames, null, 2)}`);
+        log.debug(`Files to mutate: ${JSON.stringify(this.filesToMutate.map(_ => _.name), null, 2)}`);
```
simondel: Funny, you use the temp variable `_` here while it's `file` in the line above

packages/stryker-mocha-runner/README.md
nicojs: It is the default in mocha configuration. That's why I've chosen it as a default. At the same time, I think it is a silly default, so let's make it `test/**/*.js` 👍

packages/stryker/src/input/InputFileCollection.ts
nicojs: Written the code on a different day of the week I guess. I'll align on `file`.

nicojs:
I've cleanup our changelog (force push). Ready to merge. Let's wait for one last green build

PR #678, opened 2018-04-06
test(Webpack): use Webpack 4 in integration test

### LINKED ISSUE #630: Support webpack 4
Webpack release 4 contains a lot of breaking changes. 

https://medium.com/webpack/webpack-4-migration-guide-for-plugins-loaders-20a79b927202

We should migrate the stryker-webpack-transpiler to support this version. It might make sense to not support 3 anymore (anyone interested in webpack 3 can still use the old version).

simondel:
Fixes #630 together with #664

### DIFF
--- a/integrationTest/package.json
+++ b/integrationTest/package.json
@@ -24,7 +24,7 @@
     "mz": "^2.6.0",
     "ts-node": "^5.0.0",
     "typescript": "^2.4.2",
-    "webpack": "^3.10.0"
+    "webpack": "^4.2.0"
   },
   "scripts": {
     "postinstall": "install-local && link-parent-bin -c test --link-local-dependencies true",
--- a/integrationTest/test/angular-project/package.json
+++ b/integrationTest/test/angular-project/package.json
@@ -45,6 +45,8 @@
   "localDependencies": {
     "stryker": "../../../packages/stryker",
     "stryker-webpack-transpiler": "../../../packages/stryker-webpack-transpiler",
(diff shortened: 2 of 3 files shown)


simondel:
Reading the `parents` property from a `chunk` is causing issues. I think this is the code: https://github.com/stryker-mutator/stryker/blob/master/packages/stryker-webpack-transpiler/src/compiler/FileSorter.ts

nicojs:
@simondel @Archcry 
Double check: we can remove the file chunk sorter right? Since we now leave file ordering up to test runners (transpilers are no longer responsible for file order)

simondel:
From the webpack release notes:
> Chunk.chunks/parents/blocks are no longer Arrays. A Set is used internally and there are methods to access it.

This matches the error `Error: Chunk.parents: Use ChunkGroup.getParents() instead` that we get on Travis.

We should either try to remove the use of `chunk.parents` as @nicojs mentions or build a switch to see if we run on webpack 4 because then we have to use `ChunkGroup.getParents()`.

Archcry:
I think we should remove the use of chunk.parents as @nicojs, the whole point in the refactoring we did in all the test-runners is partially because we do not want to rely on the ordering of the transpilers. So removing the sorting of chunks from the transpiler should resolve the problem.

Archcry:
The issue is caused by `CommonsChunkPlugin` that is added to the Webpack configuration used while testing an Angular project. `CommonsChunkPlugin` has been removed from Webpack 4 but Angular still uses it. Therefore the integration tests are failing.

PR #682, opened 2018-04-06
fix(Input files): passthrough all input files

### LINKED ISSUE #681: webpack transpiler v0.3.0 swallows all input files
Since the latest 0.3.0 release of the webpack transpiler it seems to swallow all input files.

nicojs:
Fixes #681


nicojs:
I have confident that the build will be green. It is green on some of the nodes. This is a pretty big issue, so I will release it now.

PR #690, opened 2018-04-10
feat(Mutator): Add StringLiteral mutator

### LINKED ISSUE #685: Support StringLiteral mutations for stryker-javascript-mutator
The stryker-javascript-mutator should support the StringLiteralMutator.

Spec: https://github.com/stryker-mutator/stryker/blob/master/packages/stryker-mutator-specification/src/StringLiteralMutatorSpec.ts

This feature will most likely have an impact on the mutation score of projects due to extra mutants being generated.

simondel:
Mutates strings into an empty string: `"Hello world!"` -> `""`
Mutates empty strings into non-empty strings: `""` -> `"Stryker was here!"`
Does not mutate `require`, `import`, `"use strict"` or JSX element properties

Fixes #685

### DIFF
--- a/packages/stryker-javascript-mutator/src/helpers/BabelParser.ts
+++ b/packages/stryker-javascript-mutator/src/helpers/BabelParser.ts
@@ -2,11 +2,12 @@ import * as babel from 'babel-core';
 import * as babylon from 'babylon';
 import generate from 'babel-generator';
 import { NodePath } from 'babel-traverse';
+import { NodeWithParent } from './ParentNode';
 
 export default class BabelParser {
   static getAst(code: string): babel.types.File {
     let ast: babel.types.File;
-    
+
     const options: babylon.BabylonOptions = {
       sourceType: 'script',
       plugins: [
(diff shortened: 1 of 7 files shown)


packages/stryker-javascript-mutator/src/helpers/BabelParser.ts
```diff
@@ -28,14 +29,15 @@ export default class BabelParser {
...
   }
 
-  static getNodes(ast: babel.types.File): babel.types.Node[] {
-    const nodes: babel.types.Node[] = [];
+  static getNodes(ast: babel.types.File): NodeWithParent[] {
+    const nodes: NodeWithParent[] = [];
 
     babel.traverse(ast, {
       enter(path: NodePath<babel.types.Node>) {
         const node = path.node;
```
nicojs: If you do the cast here, it is only needed once

simondel: I was having issues with this cast because the value that I assign does not have a parent property, that's on `path.parent`

nicojs: Good point. But at the same time, the first node will not have a parent (undefined). So it might need to be optional, right?

simondel: Good point! I'll make it optional

nicojs: Good point! You do that.

(couldn't help myself)

simondel:
@nicojs I fixed your comments 👍

packages/stryker-javascript-mutator/src/mutators/StringLiteralMutator.ts
```diff
@@ -22,7 +22,8 @@ export default class StringLiteralMutator implements NodeMutator {
       }
 
       nodes.push(mutatedNode);
-    } else if (types.isStringLiteral(node) && !types.isImportDeclaration(node.parent) && !types.isJSXAttribute(node.parent)) {
+    } else if ((node.parent === undefined || (node.parent && !types.isImportDeclaration(node.parent) && !types.isJSXAttribute(node.parent)))
```
nicojs: Can be simplified:

`if (!node.parent || (!types.isImportDeclaration(node.parent) && !types.isJSXAttribute(node.parent)))`

PR #694, opened 2018-04-13
fix(Dependencies): set correct stryker-api dependency

simondel:


simondel:
Yep. I noticed it in your demo yesterday :(

PR #712, opened 2018-04-20
Refactor(Webpack): Remove FileSorter

Archcry:
Removes the Filesorter from the Stryker-webpack-plugin that is no longer needed because of the new implementations of the test-runners.


packages/stryker-webpack-transpiler/src/compiler/FileSorter.ts
```diff
@@ -1,69 +0,0 @@
-import * as path from 'path';
```
nicojs: You can remove the `toposort` dependency from the package.json as well.

PR #714, opened 2018-04-20
fix(Sandbox pool): remove race condition

### LINKED ISSUE #713: Race condition in SandboxPool
There is a race condition in the `SandboxPool`. If the problem occurs, the SandboxPool will already signal that it is done disposing when there are actually still sandbox processes starting. This can result in the following error being logged:

```
[2018-04-20 14:23:16.837] [ERROR] TestRunnerChildProcessAdapter - C:\z\test\tsnodemocha\node_modules\stryker\src\isolated-runner\IsolatedTestRunnerAdapterWorker.js:59
        process.chdir(message.runnerOptions.sandboxWorkingFolder);
```

After stryker reports that it should be done.

nicojs:
Remove a race condition where the initalization of a sandbox takes
longer than it takes to run stryker.

Fixes #713

### DIFF
--- a/packages/stryker/src/SandboxPool.ts
+++ b/packages/stryker/src/SandboxPool.ts
@@ -9,8 +9,7 @@ import Sandbox from './Sandbox';
 export default class SandboxPool {
 
   private readonly log = getLogger(SandboxPool.name);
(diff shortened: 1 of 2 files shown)


simondel:
Awesome, nice find!

PR #715, opened 2018-04-20
feat(integration): Add Webpack zero config integration test with Karma

Archcry:
Add integration test for Webpack zero config with Karma.

Depends on https://github.com/stryker-mutator/stryker/pull/678, build will be red untill https://github.com/stryker-mutator/stryker/pull/678 is merged into master and master is merged here.


integrationTest/test/webpack-zero-conf-karma/.gitignore
```diff
@@ -0,0 +1,27 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
```
nicojs: Is this `.gitignore` file really needed? We have a pretty good gitignore file at the root, no?

Archcry: Does it filter the dist folder?

Archcry:
Tests fail because the port for Karma is already in use, Node LTS and 4 don't have this problem it seems.

nicojs:
Node 4 doesn't have the problem because the angular project tests are not run on node 4. Because angular-cli requires node 6 and up.

Archcry:
It's not because of angular-cli, it's because karma tries to start a server on a port that is already occupied.

nicojs:
I've resolved the port conflict in the integration tests. Will be merged when the build is green.

Issue #667, opened 2018-03-23
Remove side effects from stryker plugins


nicojs:
Right now it is quite difficult to to tryout 1 plugin by linking it in your project (`npm i file:./path/to/plugin`). This is because plugins work based on global side effects.

For example: `foo-transpiler`. The index.js file would look like:

```js
const { TranspilerFactory } = require('stryker-api/transpile');
const FooTranspiler = /*...*/;
TranspilerFactory.instance().register('foo', FooTranspiler);
```

The problem is that the instance of `require('stryker-api/transpile');` is different in the linked package. This is by design (standard nodejs require functionality).

We can solve this issue by not relying on side effects. Instead plugins would export:

```js
const FooTranspiler = /*...*/
exports.FooTranspiler = FooTranspiler;

/* OR in typescript: */
export { FooTranspiler } from './FooTranspiler';
```

@simondel what do you think?

The magic will be the postfix. This could be one of these: `ConfigEditor`, `Transpiler`, `Runner`, `Framework`, `Mutator`.

If a plugin uses one of the factories we should create a deprecation warning. We can remove the factories in the 1.0 release.

UPDATE:

Lot of work is done. Still todo:

- [ ] Replace calls to `getLogger` to DI
- [ ] Deprecate old api (using `process.emitWarning`): https://nodejs.org/dist/latest-v11.x/docs/api/process.html#process_process_emitwarning_warning_type_code_ctor
- [ ] Log deprecated message when loading `config` via DI (using `process.emitWarning`): https://nodejs.org/dist/latest-v11.x/docs/api/process.html#process_process_emitwarning_warning_type_code_ctor

nicojs:
This is actually also the reason that people cannot simply use `npm i ../stryker/packages/stryker-webpack-transpiler` to install an altered plugin.

nicojs:
In order to understand exactly what happens in NodeJS land, it makes sense to look at how node modules are loaded.

The text of a javascript file being `required` is actually just placed in a _wrapper function_. The wrapper function looks like this: (from https://nodejs.org/api/modules.html#modules_the_module_wrapper)

```js
(function(exports, require, module, __filename, __dirname) {
// Module code actually lives in here
});
```

Both `exports` and `module` can be used to export stuff from the module. To illustrate what happens (from https://nodejs.org/api/modules.html#modules_exports_shortcut):

```js
  const module = { exports: {} };
  const { require, filename, dirname } = /*...*/;
[...]

PR #723, opened 2018-04-26
WIP: feat(Vue): add Vue mutator

### LINKED ISSUE #579: stryker-javascript-mutator ignores Vue files
The stryker-javascript-mutator plugin ignores .vue files as they are not a part of the known extensions.

simondel:
Fixes #579


packages/stryker-vue-mutator/src/VueMutator.ts
```diff
@@ -0,0 +1,80 @@
...
+import { Mutant, Mutator, MutatorFactory } from 'stryker-api/mutant';
+const compiler = require('vue-template-compiler');
+
+export default class VueMutator implements Mutator {
+  private mutators: { [name: string]: Mutator; };
+
+  constructor(private config: Config) { }
+
+  mutate(inputFiles: File[]): Mutant[] {
+    this.initializeMutators();
```
nicojs: Can't we move this to the constructor. That way it works nicely with TS 2.7 `--strictPropertyInitialization`

packages/stryker-vue-mutator/src/VueMutator.ts
```diff
@@ -0,0 +1,80 @@
...
+    factory.knownNames().forEach(name => {
+      if (name !== 'es5' && name !== 'vue') {
+        this.mutators[name] = factory.create(name, this.config);
+      }
+    });
+  }
+
+  private getVueScriptMutator(script: any): Mutator {
+    const lang: string | undefined = script.attrs.lang;
+    let mutatorName: string;
```
nicojs: We can make it a `const` if we move the case switch to a separate method with `return` statements in the `case`s

packages/stryker-vue-mutator/src/VueMutator.ts
```diff
@@ -0,0 +1,80 @@
...
+        break;
+      case 'ts':
+      case 'typescript':
+        mutatorName = 'typescript';
+        break;
+      default:
+        throw new Error(`Found unsupported language attribute 'lang="${lang}"' on a <script> block.`);
+    }
+
+    let mutator = this.mutators[mutatorName];
```
nicojs: `const` please :)

packages/stryker-vue-mutator/src/VueMutator.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (mutator === undefined) {
+      throw new Error(`The '${mutatorName}' mutator is required to mutate a <script> block but it was not found. Please read the README of this package for information on configuration.`);
+    }
+    return mutator;
+  }
+
+  private getMutator(file: File): Mutator {
+    let mutator = this.mutators['typescript'] || this.mutators['javascript'];
```
nicojs: `const` please

packages/stryker-vue-mutator/test/integration/VueMutator.it.ts
```diff
@@ -0,0 +1,87 @@
...
+      var fs = require('fs');
+      function checkAge(user) {
+        return user.age >= 18;
+      }`;
+      const files = [new File('AppComponent.vue', vueCode), new File('age.js', jsCode)];
+      const sut = new VueMutator(new Config());
+
+      const mutants = sut.mutate(files);
+
+      expect(mutants.length).to.equal(8);
```
nicojs: This test is pretty flaky, right? Whenever we update our js mutator, this might fail?

packages/stryker-vue-mutator/test/unit/VueMutatorSpec.ts
```diff
@@ -0,0 +1,315 @@
...
+            }
+          }
+        </script>`);
+
+      const files = [vueFile];
+
+      expect(() => sut.mutate(files)).throws(`The 'javascript' mutator is required to mutate a <script> block but it was not found. Please read the README of this package for information on configuration.`);
+    });
+
+    describe('when the first file is not a Vue file', () => {
```
nicojs: I think this describe is not helping with the explanation of what is tested in the `it`. Could we move the `it` to the top level and remove this `describe`?

packages/stryker-vue-mutator/src/VueMutator.ts
```diff
@@ -0,0 +1,80 @@
...
+      } else {
+        const mutator = this.getMutator(file);
+        mutants = mutants.concat(mutator.mutate([file]));
+      }
+    });
+
+    return mutants;
+  }
+
+  private initializeMutators() {
```
nicojs: Can we move this method to a separate file? That way the unit tests can be independent of the current `MutatorFactory` way of working. We will remove the mutator factory as part of #667  .

packages/stryker-vue-mutator/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "vue-template-compiler": "^2.5.15"
+  },
+  "devDependencies": {
+    "stryker-api": "^0.17.0",
+    "stryker-javascript-mutator": "^0.7.0",
+    "stryker-typescript": "^0.11.0"
+  },
+  "peerDependencies": {
+    "stryker-api": "^0.17.0",
+    "vue-template-compiler": "^2.0.0"
```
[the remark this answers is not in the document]
simondel: This plugin is used to extract the template code from the vue file. We need it in order to run but we need the version that is exactly the same as the version of vue itself. Else the plugin won't work.

packages/stryker-vue-mutator/src/VueMutator.ts
simondel: Fixed

packages/stryker-vue-mutator/src/VueMutator.ts
simondel: Fixed

packages/stryker-vue-mutator/src/VueMutator.ts
simondel: Fixed

packages/stryker-vue-mutator/src/VueMutator.ts
simondel: I'd prefer we don't do this to keep the readability that we have right now

packages/stryker-vue-mutator/src/VueMutator.ts
```diff
@@ -0,0 +1,80 @@
...
+  mutate(inputFiles: File[]): Mutant[] {
+    this.initializeMutators();
+    let mutants: Mutant[] = [];
+
+    inputFiles.forEach(file => {
+      if (file.name.endsWith('.vue')) {
+        const script = compiler.parseComponent(file.textContent).script;
+        const mutator = this.getVueScriptMutator(script);
+        const vueFile = new File(
+          file.name + '.js', /* TODO: Remove this once all mutators understand that vue files are OK to mutate */
```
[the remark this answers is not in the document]
simondel: Fixed

packages/stryker-vue-mutator/README.md
```diff
@@ -0,0 +1,63 @@
...
+[![NPM](https://img.shields.io/npm/dm/stryker-vue-mutator.svg)](https://www.npmjs.com/package/stryker-vue-mutator)
+[![Node version](https://img.shields.io/node/v/stryker-vue-mutator.svg)](https://img.shields.io/node/v/stryker-vue-mutator.svg)
+[![Gitter](https://badges.gitter.im/stryker-mutator/stryker.svg)](https://gitter.im/stryker-mutator/stryker?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
+[![BCH compliance](https://bettercodehub.com/edge/badge/stryker-mutator/stryker)](https://bettercodehub.com/)
+
+![Stryker](https://github.com/stryker-mutator/stryker/raw/master/stryker-80x80.png)
+
+# Stryker Vue mutator
+
+A mutator that supports JavaScript for [Stryker](https://stryker-mutator.io), the mutation testing framework for JavaScript and friends.
```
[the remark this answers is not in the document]
simondel: Fixed

packages/stryker-vue-mutator/src/VueMutator.ts
simondel: Fixed

packages/stryker-vue-mutator/test/unit/VueMutatorSpec.ts
simondel: Fixed

packages/stryker-vue-mutator/package.json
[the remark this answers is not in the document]
nicojs: Ok, in that case, please remove the dependency on `"vue-template-compiler"`

packages/stryker-vue-mutator/src/helpers/MutatorHelpers.ts
```diff
@@ -0,0 +1,15 @@
+import { Mutator, MutatorFactory } from 'stryker-api/mutant';
+import { Config } from 'stryker-api/config';
+
+const generateMutators = (config: Config): { [name: string]: Mutator; } => {
```
nicojs: This name... In what way is it actually generating mutators? More like `discoverMutators` right?

nicojs:
Is this ready to be reviewed?

simondel:
Allright, one integration test is done. Do we also want a project with jest and typescript @nicojs ?

nicojs:
Hmm. What do you think? Couldn't hurt of course, but does it test something new?

simondel:
It redirects the code to a different path because all the Vue mutator does is send code to the correct mutator.

packages/stryker-vue-mutator/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "homepage": "https://github.com/stryker-mutator/stryker/tree/master/packages/stryker-vue-mutator#readme",
+  "license": "Apache-2.0",
+  "devDependencies": {
+    "stryker-api": "^0.17.0",
+    "stryker-javascript-mutator": "^0.7.0",
+    "stryker-typescript": "^0.11.0",
+    "vue-template-compiler": "^2.0.0"
+  },
+  "peerDependencies": {
+    "stryker-api": "^0.17.0",
```
simondel: This has to be updated

simondel:
The package itself does have tests to see if the code works though.

packages/stryker-vue-mutator/src/VueMutator.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const mutator = this.mutators[mutatorName];
+    if (mutator === undefined) {
+      throw new Error(`The '${mutatorName}' mutator is required to mutate a <script> block but it was not found. Please read the README of this package for information on configuration.`);
+    }
+    return { mutator, extension };
+  }
+
+  private getMutator(file: File): Mutator {
+    const mutator = this.mutators['typescript'] || this.mutators['javascript'];
```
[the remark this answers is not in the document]
nicojs: Also, couldn't we fail faster? As soon as the `mutate` method is called and no mutators are registered for example?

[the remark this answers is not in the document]
simondel: I'll save the string names `'typescript'` and `'javascript'` somewhere. I think the typescript mutator should always be able to process javascript code. Javascript is typescript after all, right?

We could fail faster, but we would still have to check if the correct mutator is available for Vue code. For instance, we can't just say 'if either of the two mutators exists, all is fine' because the JavaScript mutator may be present while there is a Vue file with TypeScript code.

simondel:
@nicojs Thanks for the review! I've refactored the code.

simondel:
I'll try to make them less flakey later this week 👍

simondel:
I have updated the tests with my latest commit.

nicojs:
I see you mocked away the mutators for unittests. Great. Unittests are a little wet (vs dry) for my taste, but ok to merge

nicojs:
Wet as opposite of DRY indeed. I like my unit tests a little [DAMP](https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests)

PR #724, opened 2018-04-29
feat(node version): drop node 4 support

nicojs:
BREAKING CHANGE: Node 4 is no longer supported.


.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
 - node
+- '8'
```
simondel: How about we don't explicitly build for node 8? In most cases, our software will either not work on the latest version of node or the oldest version. Not something in between.

.travis/ci-release.jsh
```diff
@@ -6,7 +6,7 @@ const exec = (command) => execSync(command, { stdio: [0, 1, 2] });
 const releaseRequirements = {
     pullRequest: 'false',
     branch: 'release',
-    nodeVersion: '6'
+    nodeVersion: '8'
```
simondel: Why not release with node 6?

PR #736, opened 2018-05-04
feat(webpack): Add support for config files exporting a function

ollelauribostrom:
The stryker-webpack-transpiler currently expects that the webpack config file exports an object. This PR adds support for config files that exports a function as well - to avoid having to create multiple webpack config files.


nicojs:
Thanks @ollelauribostrom! We will be releasing this later today.

PR #737, opened 2018-05-05
 feat(Jasmine runner): support jasmine as test runner

### LINKED ISSUE #164: Add support for jasmine-node test runner
We should add support for the jasmine-node test runner.

See https://github.com/bkimminich/juice-shop/issues/219

Follow progress non the `164-feat-jasmine-runner` branch.

Still todo:

- [ ] Add unit tests (worked with integration tests so far)
- [ ] Support the jasmine test framework ( + IT).
- [ ] Add an integration test to run stryker with node-jasmine (integrationTest module)

nicojs:
Add jasmine as a _test runner_.

We've supported jasmine for a while, but only as a test framework when running in karma. It's time to add support for the [jasmine node test runner](https://jasmine.github.io/setup/nodejs.html) (a.k.a [jasmine-npm](https://github.com/jasmine/jasmine-npm)) was well.

See [the readme](https://github.com/stryker-mutator/stryker/tree/ae58480e7e9180aeecf46aaf769ad9dc36dcbfa1/packages/stryker-jasmine-runner#readme) for more information on how to configure.

This feature includes:

* Add `'jasmine'` as a possible test runner.
* Supports testHooks. In other words: you can use `coverageAnalysis: "perTest"` (as long as you also configure the test framework).
* Add integration test for a jasmine-jasmine project.

Fixes #164


packages/stryker-jasmine-runner/src/JasmineTestRunner.ts
```diff
@@ -0,0 +1,70 @@
...
+import { StrykerOptions } from 'stryker-api/core';
+import { Jasmine, toStrykerTestResult, evalGlobal } from './helpers';
+
+export default class JasmineTestRunner implements TestRunner {
+
+  private jasmineConfigFile: string | undefined;
+  private fileNames: ReadonlyArray<string>;
+  private Date: { new(): Date } = Date; // take Date prototype now we still can (user might choose to mock it away)
+
+  constructor({ fileNames, strykerOptions: { jasmineConfigFile } }: { fileNames: ReadonlyArray<string>, strykerOptions: StrykerOptions }) {
```
simondel: To me, this code is more readable:
```typescript
constructor(private fileNames: ReadonlyArray<string>, strykerOptions: StrykerOptions) {
    this.jasmineConfigFile = strykerOptions.jasmineConfigFile;
  }
```

packages/stryker-jasmine-runner/src/JasmineTestRunner.ts
```diff
@@ -0,0 +1,70 @@
...
+    this.jasmineConfigFile = jasmineConfigFile;
+    this.fileNames = fileNames;
+  }
+
+  run(options: { testHooks?: string }): Promise<RunResult> {
```
simondel: Why do you specify the type here like this instead of using `RunOptions`?

packages/stryker-jasmine-runner/src/JasmineTestRunner.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        jasmineDone() {
+          resolve({
+            errorMessages: [],
+            status: RunStatus.Complete,
+            tests
+          });
+        }
+      };
+      jasmine.addReporter(reporter);
```
simondel: This will work because we create a new jasmine runner instance for every testrun right? Because adding a reporter twice may cause issues

packages/stryker-jasmine-runner/test/unit/JasmineTestRunnerSpec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  it('should clear require cache on run', async () => {
+    require.cache['foo.js'] = 'foo';
+    require.cache['bar.js'] = 'bar';
+    await actEmptyRun();
+    expect(require.cache['foo.js']).not.ok;
+    expect(require.cache['bar.js']).not.ok;
+  });
+
+  it('should report completed specs', async () => {
```
simondel: Could you format the test files so there is an empty line between the `setup`, `act` and `assert` stages to improve readability?

packages/stryker-jasmine-runner/test/unit/JasmineTestRunnerSpec.ts
```diff
@@ -0,0 +1,124 @@
...
+      .and.matches(/.*Error: foobar.*/);
+  });
+
+  it('should evaluate testHooks in global context', async () => {
+    const hooks = 'foobar';
+    await actEmptyRun(hooks);
+    expect(evalGlobalStub).calledWith(hooks);
+  });
+
+  function actEmptyRun(testHooks?: string) {
```
simondel: Could you add some comments to describe what this does? (And in what way it's `empty`)

integrationTest/test/jasmine-jasmine/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "private": true,
+  "description": "A module to perform an integration test",
+  "main": "index.js",
+  "scripts": {
+    "pretest": "rimraf \"reports\" \"verify/*.map\" && tsc -p .",
+    "test": "stryker run",
+    "posttest": "mocha --require ts-node/register verify/*.ts"
+  },
+  "author": "",
+  "license": "ISC"
```
simondel: Should this be ISC or Apache-2.0?

packages/stryker-jasmine-runner/README.md
```diff
@@ -0,0 +1,51 @@
...
+    });
+}
+```
+
+### `jasmineConfigFile` [`string`]
+
+Default: `undefined`
+
+Specify your [jasmine configuration file](https://jasmine.github.io/setup/nodejs.html#configuration) to be loaded. 
+Leaving this blank will result in the jasmine defaults, which are undocumented.
```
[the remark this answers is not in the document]
nicojs: The defaults from the jasmine runner itself. I will add a link to the source code where the defaults are defined.

packages/stryker-jasmine-runner/src/JasmineTestRunner.ts
nicojs: True, but defining it like this makes it much easier to test.

We don't need to specify all required stryker options, instead we can focus on only the stuff that's actually used.

For example:

```js
new JasmineTestRunner({
        fileNames: [
           //...
        ], strykerOptions: { jasmineConfigFile: 'spec/support/jasmine.json' }
});
```

packages/stryker-jasmine-runner/src/JasmineTestRunner.ts
nicojs: See my previous remark.

packages/stryker-jasmine-runner/src/JasmineTestRunner.ts
nicojs: Yes indeed. It is required, jasmine does not allow a runner to be reused unfortunately.

integrationTest/test/jasmine-jasmine/package.json
nicojs: Good point! I'll make it `Apache-2.0`

packages/stryker-jasmine-runner/src/JasmineTestRunner.ts
simondel: What could the impact of this notation be if we decide to refactor the StrykerOptions? Would our editors understand that this code is also using the StrykerOptions?

packages/stryker-jasmine-runner/package.json
```diff
@@ -0,0 +1,48 @@
+{
```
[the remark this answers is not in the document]
nicojs: :+1:

packages/stryker-jasmine-runner/src/JasmineTestRunner.ts
nicojs: I would like to keep this code. We ignore the timeout setting.

packages/stryker-jasmine-runner/test/unit/JasmineTestRunnerSpec.ts
nicojs: 👍

packages/stryker-jasmine-runner/test/unit/JasmineTestRunnerSpec.ts
nicojs: Renamed to `actRunWithoutTests`

PR #742, opened 2018-05-11
 feat(jasmine-3): support jasmine 3 as test framework 

nicojs:
This PR adds support for jasmine 3 as a test framework.


packages/stryker-jasmine/src/JasmineTestFramework.ts
```diff
@@ -32,11 +32,10 @@ export default class JasmineTestFramework implements TestFramework {
   }
 
   filter(testSelections: TestSelection[]): string {
-    const ids = testSelections.map(selection => selection.id);
+    const names = testSelections.map(selection => selection.name);
```
simondel: The randomized order does not matter for us with this change because only the id changes, right? (Since we do run non-randomized in the tests, where we use ordering)

nicojs: True

PR #743, opened 2018-05-11
test(Allow js): add integration test for mixing js and ts code

### LINKED ISSUE #741: Support JS files in the stryker-typescript transpiler
A TypeScript project can allow JS files. However, this is not supported yet by the `stryker-typescript-transpiler`. It will try to `emit` the .js file, which results in an error.

This issue is discovered while investigating #684 and should be fixed before we can proceed to troubleshoot that issue.

nicojs:
Fixes #741


nicojs:
It was apparently already supported. Add an integration test so we don't suffer from regression in the future.

Issue #744, opened 2018-05-11
Dashboard Reporter is too tightly coupled to Travis


mthmulders:
### Summary
The Dashboard Reporter is too tightly coupled to [Travis CI](http://travis-ci.org/), that it simply does not work on other CI environments like [Circle CI](https://circleci.com/dashboard):

> Dashboard report is not sent when stryker didn't run on buildserver {TRAVIS=true}

### Your Environment
My personal projects build on CircleCI and that means I currently can't use the Stryker Mutator Dashboard feature.

### Idea
I would suggest to de-couple the `DashboardReporter` from any CI provider. We could then write a `CiProvider` interface with implementations for e.g. Travis, Circle, and maybe other. I could imagine such an interface having methods like:
* `isActve`, so the reporter or a helper can decide which implementation to use for the remaining steps
* `isPullRequest`, so the reporter can decide not to send the mutation score
* `determineBranch` for determining the branch that is currently being built
* `determineRepository` for determining the name of the repository that is currently being built

Any comments, improvements, rants or insults...?

nicojs:
I'm curious, would you like for the `DashboardReporter` to also support dynamic loading of more ci providers? If not, i'd suggest to not over complicate things and just add support for CircleCI. 

@mthmulders Out of curiosity, is your project on GH? We only support the dashboard for open source GH projects for now.

mthmulders:
> Would you like for the DashboardReporter to also support dynamic loading of more ci providers?

I would like Stryker to automatically figure out a) it is running in a CI environment and b) which one. To me, it doesn't make sense to have configuration settings that only apply in a CI/CD-environment. And I don't think we should bother end users with that. It's relatively easy to auto-detect both of 'em, I think...

> Out of curiosity, is your project on GH? 

Yes, it is. It's just that there are CI/CD providers out there than Travis...

> We only support the dashboard for open source GH projects for now.

I noticed :-)

mthmulders:
Sure, I'll give it a shot...

simondel:
@mthmulders @nicojs This can be closed right?

PR #747, opened 2018-05-13
Broader CI support

mthmulders:
As noted in #744, the Dashboard reported will currently only work when building on Travis.
This PR addresses that by de-coupling the reporter from the CI/CD system being used.

### DIFF
--- a/packages/stryker/src/reporters/ci/Provider.ts
+++ b/packages/stryker/src/reporters/ci/Provider.ts
@@ -1,0 +1,29 @@
+import TravisProvider from './TravisProvider';
+
+import { getEnvironmentVariable } from '../../utils/objectUtils';
+
+/**
+ * Represents an object that can provide information about a CI/CD provider.
+ */
+export interface CiProvider {
+  /** Returns whether Stryker is running on a pull request. */
+  isPullRequest: () => boolean;
+  /** Returns the name of the Git branch of the project on which Stryker is running. */
+  determineBranch: () => string;
+  /** Returns the name of the GitHub repository of the project on which Stryker is running. */
+  determineRepository: () => string;
+}
+
+/**
+ * Return an approriate instance of CiProvider.
+ * @returns An instance of CiProvider, or undefined if it appears Stryker is not running in a CI/CD environment.
+ */
+export function determineCiProvider()  {
+  // By far the coolest env. variable from all those listed at
+  // https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables
+  if (getEnvironmentVariable('HAS_JOSH_K_SEAL_OF_APPROVAL')) {
+    return new TravisProvider();
+  }
+
+  return undefined;
+}
--- a/packages/stryker/src/reporters/ci/Provider.ts
+++ b/packages/stryker/src/reporters/ci/Provider.ts
@@ -1,3 +1,4 @@
+import CircleProvider from './CircleProvider';
 import TravisProvider from './TravisProvider';
 
 import { getEnvironmentVariable } from '../../utils/objectUtils';
@@ -23,6 +24,9 @@ export function determineCiProvider()  {
   // https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables
   if (getEnvironmentVariable('HAS_JOSH_K_SEAL_OF_APPROVAL')) {
     return new TravisProvider();
+  } else if (getEnvironmentVariable('CIRCLECI')) {
+    return new CircleProvider();
+
   }
 
   return undefined;
--- a/packages/stryker/test/unit/reporters/ci/CircleProviderSpec.ts
+++ b/packages/stryker/test/unit/reporters/ci/CircleProviderSpec.ts
@@ -1,0 +1,75 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import * as environmentVariables from '../../../../src/utils/objectUtils';
+
+import CircleProvider from '../../../../src/reporters/ci/CircleProvider';
+
+describe('CircleCI Provider', () => {
+  let getEnvironmentVariables: sinon.SinonStub;
+
+  const circleProvider = new CircleProvider();
+
+  beforeEach(() => {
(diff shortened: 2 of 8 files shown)


packages/stryker/test/unit/reporters/ci/CircleProviderSpec.ts
```diff
@@ -0,0 +1,75 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import * as environmentVariables from '../../../../src/utils/objectUtils';
+
+import CircleProvider from '../../../../src/reporters/ci/CircleProvider';
+
+describe('CircleCI Provider', () => {
+  let getEnvironmentVariables: sinon.SinonStub;
+
+  const circleProvider = new CircleProvider();
```
nicojs: I know the `CircleProvider` does not have state. Yet I would feel better if we created a new one for each test. If it would have state in the future, you wouldn't automatically expect state to flow from test to test.

packages/stryker/test/unit/reporters/ci/TravisProviderSpec.ts
```diff
@@ -0,0 +1,70 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import * as environmentVariables from '../../../../src/utils/objectUtils';
+
+import TravisProvider from '../../../../src/reporters/ci/TravisProvider';
+
+describe('Travis Provider', () => {
```
nicojs: This is OK for now, but we might want to consider moving this to a more DRY approach. Defining the specs in a function and calling that function from both the `TravisProviderSpec` and the `CircleProviderSpec`.

packages/stryker/test/unit/reporters/ci/CircleProviderSpec.ts
mthmulders: In fact, I thought of it the other way around, in a more test-driven approach: I don't *want* the Providers to have state. So let's set it up in a way that keeping state would be hard to achieve :-).

packages/stryker/src/reporters/ci/Provider.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Return an approriate instance of CiProvider.
+ * @returns An instance of CiProvider, or undefined if it appears Stryker is not running in a CI/CD environment.
+ */
+export function determineCiProvider()  {
+  // By far the coolest env. variable from all those listed at
+  // https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables
+  if (getEnvironmentVariable('HAS_JOSH_K_SEAL_OF_APPROVAL')) {
+    return new TravisProvider();
+  } else if (getEnvironmentVariable('CIRCLECI')) {
```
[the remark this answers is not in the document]
mthmulders: I've considered that one, because - indeed - it would centralise knowledge of each CI system in one place. What withheld me is that it would require us to _first_ instantiate each provide and _then_ check whether it is active. But if you feel like this drawback isn't that big, it does give a nice advantage, as you pointed out.
WDYT?

packages/stryker/test/unit/reporters/ci/CircleProviderSpec.ts
nicojs: This suggests that the test would be red, while in fact it could just as easily be that the tests are green for 1 test run, but red the next one. You'd never notice, because tests always run just once, *unless* of course during mutation testing. The result is a very high mutation score, seeing as all runs fail. We've been burned by this once before, and it took me hours to figure out (why do we have a mutation score of 98%?? surely our tests aren't *that* good)

I agree that it is a bad practice to keep state in a provider, but on the other hand I think it is a bad practice to share state over tests. Both should stand on its own. Do you agree?

packages/stryker/src/reporters/ci/Provider.ts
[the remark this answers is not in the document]
nicojs: I would personally add it to the `CIProvider` interface. 

You could opt to change the `determineCiProvider` to a member of a class. In this class you can add a constructor that excepts a list of `CIProvider`s.

Something like this:

```ts
abstract class CIProvider {

  abstract isActive(): boolean;
  abstract isPullRequest(): boolean;
  abstract determineBranch(): string;
  abstract determineRepository(): string;
 
   static active( providers = [new TravisProvider(), new CircleProvider()] ): CIProvider | null {
     
   }
}
```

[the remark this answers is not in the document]
nicojs: Hmm all of that seems a little over-engineered in hindsight. Let's go with your implementation. We can always change it later if we want to support custom build pipelines.

packages/stryker/src/reporters/ci/Provider.ts
```diff
@@ -0,0 +1,33 @@
...
+  determineBranch: () => string;
+  /** Returns the name of the GitHub repository of the project on which Stryker is running. */
+  determineRepository: () => string;
+}
+
+/**
+ * Return an approriate instance of CiProvider.
+ * @returns An instance of CiProvider, or undefined if it appears Stryker is not running in a CI/CD environment.
+ */
+export function determineCiProvider()  {
```
nicojs: This should also be renamed to `determineCIProvider`.

nicojs:
I'll squash and merge as soon as the build is green.

nicojs:
There still seems to be one rename left: 

```
stryker@0.23.0 prebuild /home/travis/build/stryker-mutator/stryker/packages/stryker\n> rimraf "+(test|src)/**/*+(.d.ts|.js|.map)" .nyc_output reports coverage\n\n\n> stryker@0.23.0 build /home/travis/build/stryker-mutator/stryker/packages/stryker\n> tsc -p .\n\ntest/unit/reporters/DashboardReporterSpec.ts(23,52): error TS2345: Argument of type \'"determineCiProvider"\' is not assignable to parameter of type \'"determineCIProvider"\'.\n',
```

mthmulders:
Which part of the documentation should be updated?

nicojs:
Reporters: https://github.com/stryker-mutator/stryker/tree/master/packages/stryker#reporters

PR #829, opened 2018-05-31
fix(String mutator): do not mutate prologue directives

avdg:
String literals in a expression statement tend not to have any side effects. So lets assume it's just dead code.

The first string literals in an expression may, given ecma standard rules be converted to prologue directives and have no meaning when being executed.

### DIFF
--- a/packages/stryker-mutator-specification/src/StringLiteralMutatorSpec.ts
+++ b/packages/stryker-mutator-specification/src/StringLiteralMutatorSpec.ts
@@ -50,5 +50,10 @@ export default function StringLiteralMutatorSpec(name: string, expectMutation: E
     it('should not mutate string JSX attributes', () => {
       expectMutation('<Record class="row" />');
     });
+
+    it('should not mutate directive prologues', () => {
+      expectMutation('"use strict";"use asm";');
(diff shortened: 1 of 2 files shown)


nicojs:
Quote from gitter:

> Other statements like `'baz'` in this example will now also not be mutated
> 
> ```ts
> const foo = 'bar';
> 'baz';
> console.log(foo);
> ```
> 
> But that's OK i think. Stryker is not a "dead code" detector.

simondel:
I guess my Gitter notifications broke again..

avdg:
Why mutate something that does not make things fail when we want deliberate failure? Of course, we still have to look out for cases where mutations creates failures, but I fail to see one in this case.

nicojs:
Great job @avdg !

Curious as to why the javascript mutator didn't need a change. Did it already support not-mutating the prologue directives?

avdg:
Maybe this explains: https://github.com/babel/babel/blob/dccaec7691dd3a5217e8464ed3acd91370668f4a/packages/babel-types/src/validators/generated/index.js#L64

Maybe babel doesn't convert directives in string literals

nicojs:
Great work on this PR :+1:

PR #866, opened 2018-06-08
Feat(node): Downgrade node types to version 6

Archcry:
We are currently still supporting node version 6
so it makes sense to downgrade the types we
use to version 6.


Archcry:
Waiting for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26376

PR #926, opened 2018-06-30
feat(karma-runner): add angular-cli support

nicojs:
This adds _native_ angular cli support to the stryker-karma-runner. When configuring the stryker-karma-runner to use the angular cli, it will use `ng test` to start the test runner, instead of `karma start`. This enables angular to do its webpack magic.

You now configure the stryker-karma-runner like this:

```js
karma: {
  project: 'angular-cli', // or 'custom' as the default
  configFile: 'src/karma.conf.js', // location of the karma config file
  config: {
    browsers: ['ChromeHeadless'] // override karma config for stryker here
  }
}
```

The old configuration options `karmaConfigFile` and `karmaConfig` are deprecated, but will still work for the time being.

This is a complete rewrite of the karma test runner. Instead of relying on `new karma.Server(..)`, we now allow `ng test` as an alternative. To hook into the karma server, we supply _our own_ karma config file. From there, we hook into karma for stryker specific logic as well as load additional user configuration.


packages/stryker-karma-runner/src/StrykerReporter.ts
```diff
@@ -0,0 +1,88 @@
...
+      failureMessages: spec.log
+    };
+    this.emit('test_result', testResult);
+  }
+
+  onRunComplete(runResult: karma.TestResults) {
+    this.emit('run_complete', this.collectRunState(runResult));
+  }
+
+  private collectRunState(runResult: karma.TestResults): RunStatus {
```
Archcry: Could you move this function beneath the public functions?

packages/stryker-karma-runner/src/KarmaTestRunner.ts
```diff
@@ -1,235 +1,122 @@
...
+      await this.runServer();
+    }
+    const runResult = this.collectRunResult();
+    this.cleanRun();
+    return runResult;
   }
 
-  init(): Promise<void> {
-    return this.serverStartedPromise;
+  private loadSetup(settings: RunnerOptions): StrykerKarmaSetup {
```
Archcry: Would it be possible to split this function, it does three things now:

- load the karma config.
- Inform the user their karma configuration is deprecated.
- set globals.

Archcry:
seal_of_approval

nicojs:
🤦‍♂️

PR #959, opened 2018-07-08
Fixed a couple of CONTRIBUTING.md typos

JoshuaKGoldberg:
"i.e." roughly means "that is" (I think of it as "in essence"); "e.g." roughly means "for example" (I think of it as "example given").

Missing exclamation mark at the end. You've got to have the right level of oomph!


simondel:
Typo PRs are the best PRs 💓 Thanks for fixing it!

Issue #917, opened 2018-06-28
Yarn support?


JoshuaKGoldberg:
### Your Environment
| software         | version(s)
| ---------------- | -------  
| node             | 10.x
| npm              |  _(nope!)_
| Operating System | Windows 10

I use Yarn on a project. Stryker's commands all run with npm, which cause weird and angry installation errors due to differences in how the two resolve modules.

Two suggestions:
* Offer the user which they're using, with npm as the default?
* Detect a `yarn.lock` without a `package-lock.json` and switch the default to yarn if so?

nicojs:
Thanks @JoshuaKGoldberg for this feature request. I think you're specifically mean the `stryker init` command right? 

Should be fairly straight forward to implement. Do you want to give it a shot? It might take some time for us, as we're focusing on releasing 1.0 right now.

JoshuaKGoldberg:
I'm in a bit of a time crunch right now but sure, why not :smile:. Do you have any tips for getting started?

nicojs:
Haha, I think I know the feeling.

The initializer logic is all in the [initializer](https://github.com/stryker-mutator/stryker/tree/master/packages/stryker/src/initializer) directory. The actual npm call is here: https://github.com/stryker-mutator/stryker/blob/ca16dc0408e7661a3e3709719fc57ead9ef95909/packages/stryker/src/initializer/StrykerInitializer.ts#L142

I think a quick question using the [inquirer](https://github.com/stryker-mutator/stryker/blob/ca16dc0408e7661a3e3709719fc57ead9ef95909/packages/stryker/src/initializer/StrykerInquirer.ts#L9) will be enough.

However, this is only for the install command after the questionnaire. If you've installed the [stryker-cli](https://github.com/stryker-mutator/stryker-cli) and want to configure stryker on a new project using `stryker init`, you will first get a prompt asking to install stryker locally. If you want to add yarn support for that as well, you'll need to create a PR there as well. [This line](https://github.com/stryker-mutator/stryker-cli/blob/31c55a399c22f327df7acd1a79af69c76f411617/src/stryker-cli.ts#L8) is where the npm command is constructed there.

Everything is unit tested (d'uh) using mocha. Please take a look at our [contribution guide](https://github.com/stryker-mutator/stryker/blob/master/CONTRIBUTING.md#vscode-environment-configuration) for some hints to get you started with your dev environment (compiling typescript, debugging tests, etc).

JoshuaKGoldberg:
[...]

PR #962, opened 2018-07-08
Added support for yarn installs to `stryker init`

JoshuaKGoldberg:
Starts on #917.

### DIFF
--- a/packages/stryker/src/initializer/StrykerConfigWriter.ts
+++ b/packages/stryker/src/initializer/StrykerConfigWriter.ts
@@ -32,12 +32,14 @@ export default class StrykerConfigWriter {
     selectedMutator: null | PromptOption,
     selectedTranspilers: null | PromptOption[],
     selectedReporters: PromptOption[],
+    selectedPackageManager: PromptOption,
     additionalPiecesOfConfig: Partial<StrykerOptions>[]): Promise<void> {
     const configObject: Partial<StrykerOptions> = {
       testRunner: selectedTestRunner ? selectedTestRunner.name : '',
       mutator: selectedMutator ? selectedMutator.name : '',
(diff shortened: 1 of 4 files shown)


nicojs:
Hi @JoshuaKGoldberg , you're off to an awesome start! I took the liberty of fixing the compile errors. I used inquires generic type argument for that (and refactored it in the other calls). We might be able to fix this issue in inquirer's typings using mapped types. Would be an awesome use case for it.

I also fixed one unit test. I saw there were some other ones failing, you want me to fix those as well? 

Thanks a lot for all the work already!

JoshuaKGoldberg:
Ha, thanks - I was planning on re-checking this tonight. I can do them!

JoshuaKGoldberg:
I hope you'll forgive spacing some of the test object literals across lines; I found it hard to read having several members on the same line. Otherwise, unit tests are fixed. They caught that I was neglecting to write the selected package manager to the config file! 🚀

nicojs:
> I hope you'll forgive spacing some of the test object literals across lines; I

No problem! I'm still planning of integrating something like [prettier](https://www.npmjs.com/package/prettier) in our build pipeline, but haven't got around to it.

> They caught that I was neglecting to write the selected package manager to the config file! rocket

Well.. that might be an oversight on our part. Do we want the packageManager config setting? I suppose it won't hurt and we could use it in the future I guess. What do you think?

If I had to write the tests again today, i would introduce a helper method to arrange the test cases. Making them a bit more DRY. I seem to go back and forth on the dry vs damp discussion

JoshuaKGoldberg:
I would prefer to leave it in precisely for the future proofing, yes. Seems harmless otherwise?

nicojs:
@JoshuaKGoldberg awesome, thanks for your work 🐍

simondel:
Amazing 🏅

PR #1003, opened 2018-07-14
Feat(Jest): retrieve coverage from testResults Fix #965

### LINKED ISSUE #965: Feature request - Add coverage analysis for Jest
### Summary

Is it possible to add coverage analysis, to provide a better experience on mutation testing with jest ?
For now, `stryker-jest-runner` create mutants on every single part of code, even if statement or branch are not reach.

I would like to add my contribution to stryker, to help bigger application to use stryker, without to add mutation on unreachable statement during initial test run.

### Stryker config

```js
  config.set({
    testRunner: "jest",
    mutator: "javascript",
    transpilers: [],
    reporter: ["html", "clear-text", "progress"],
    coverageAnalysis: "all",
    mutate: ["src/**/*.js"]
  });
};

```

### Stryker environment

```
    "jest": "^23.2.0",
    "stryker": "^0.24.1",
    "stryker-api": "^0.17.2",
    "stryker-html-reporter": "^0.14.2",
    "stryker-javascript-mutator": "^0.7.2",
    "stryker-jest-runner": "^0.7.0"
```

### Your Environment
| ---------------- | -------  
| node             | 10.4
| npm              | 6.1
| Operating System |  OSX

Tobino:
Fixes #965

### DIFF
--- a/packages/stryker-jest-runner/src/JestTestRunner.ts
+++ b/packages/stryker-jest-runner/src/JestTestRunner.ts
@@ -36,7 +36,7 @@ export default class JestTestRunner extends EventEmitter implements TestRunner {
     const errorMessages = results.testResults.map((testSuite: jest.TestResult) => testSuite.failureMessage).filter(errorMessage => (errorMessage)) as Array<string>;
 
     // There *should* be a global "__coverage__" variable here, unfortunately it was cleaned by jest
-    const coverage = (global as any).__coverage__;
+    const coverage = results.testResults.map((testSuite: jest.TestResult) => testSuite.coverage) as any;
     return {
       tests: this.processTestResults(results.testResults),
       status: (results.numRuntimeErrorTestSuites > 0) ? RunStatus.Error : RunStatus.Complete,
--- a/packages/stryker-jest-runner/types/jest/index.d.ts
+++ b/packages/stryker-jest-runner/types/jest/index.d.ts
@@ -34,6 +34,7 @@ declare namespace Jest {
 
   // Taken from https://goo.gl/nAzQ4J, removed all stuff that we are not using
   interface TestResult {
+    coverage?: RawCoverage;
     failureMessage: Maybe<string>;
     testResults: Array<AssertionResult>;
   }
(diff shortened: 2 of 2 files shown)


nicojs:
@Tobino thanks for working on this.

It doesn't seem to work for me:

image

However there does seem to be coverage data, but I think it is because jest itself instrumented the code. Which isn't bad by itself, we should be able to make that work as well, we would just need to alter Stryker to not do the coverage analysis in this case. More changes are needed. @simondel what do you think?

image

nicojs:
I've opened a stack overflow question about globals: https://stackoverflow.com/questions/51723662/retrieve-global-variable-after-test-run-in-jest

simondel:
@nicojs Do we also have to update the Stryker readme? https://github.com/stryker-mutator/stryker/tree/master/packages/stryker#type-of-coverage-analysis

Tobino:
@nicojs on my macbook, I can retrieve the coverage data. Maybe I am working with a different version of jest. :/

The array contain two coverages, one for Add.js and one for Circle.js

Your array have 4 items, what is the other testSuite ?

stryker-jest

nicojs:
@Tobino After reading the docs and dived a bit in the Jest source code, I'm now seeing that Jest actually does it's own instrumentation using istanbul: https://github.com/facebook/jest/blob/6d3da9be06350bfa7414b609807ebd0cbdf0c416/packages/jest-runtime/src/script_transformer.js#L24

This means that we could make it work, but it would come at severe consequences:

1. We need to change Stryker to allow the test runner to instrument the code instead of Stryker itself. 
1. Because Stryker cannot instrument the code, the coverage analysis `'perTest'` will never work, because that instrumentation is different from the regular istanbul instrumentation.
1. The istanbul version that the test runner uses must be inline with the istanbul version Stryker is using. Meaning that the report structure must be the same one Stryker is supporting.
1. This implementation would break once we do our own 'mutation coverage instrumentation' that we are planning to do in the future.

IMHO those downsides are to big to allow this. 

However, really the only thing that we need jest to be able to do is report back on global variables. I've opened this stack overflow question for this. It seems that it is not possible at the moment: https://stackoverflow.com/questions/51723662/retrieve-global-variable-after-test-run-in-jest/51831522

@Tobino Would you be interested in opening up an issue at Jest side to see if they are willing to help us with this? If they could give us a way to programmatically retrieve a global variable value from the testing environment, we would be golden.

nicojs:
I went along and created an issue at the Jest side of things here: https://github.com/facebook/jest/issues/7421

I'm closing this PR for now. Once we figured out what the way of reporting the code will be we can create a new PR.

PFadel:
@nicojs any news for this ?

simondel:
@PFadel As you can see on the issue facebook/jest#7421, unfortunately not :(

PR #1042, opened 2018-07-24
fix(mocha-runner): Don't log individual successful tests

nicojs:


simondel:
Do we still want to log every test that passes/fails or do we just want the result?

nicojs:
I'll merge it when green.

simondel:
Just as a check, the log message every time a test fails is still there on purpose, right?

nicojs:
Yeah, that could be useful imo

PR #1047, opened 2018-07-27
feat(Command test runner): Add command test runner

### LINKED ISSUE #768: New test runner: Use npm test command
We can simplify the use of Stryker quite a bit if we simply use the `npm test` command of a repository. In order to do this, we can create a simple test runner, let's call it the `NpmTestTestRunner` for now. 

The idea is that, instead of hooking into the test runner and test framework, we simply use the `npm test` command. If exit code is 0, the test passed, if it is 1, the test errors. Our timeout system should still work with this approach, so a timeout would also be supported.

Basically, we would trade in performance for simplicity.

[Mutode](https://github.com/TheSoftwareDesignLab/mutode) uses this approach. @DiegoRBaquero is the author and might have some time to implement this feature.

nicojs:
Add the command test runner. This test runner simply runs command from the bash/cmd prompt. It defaults to `npm test`.

It will always report exactly _one_ test result (called `'All tests'`), as it cannot distinguish between the tests. It is implemented differently from other test runners, as we don't want to spawn this one in a separate process (it already uses `require('child_process').exec` to start a test run).

Fixes #768


simondel:
What are your thoughts on the integration with `stryker init` and the use of a test framework in your config?

nicojs:
For `stryker init`, I was thinking to put it as a last option: 

* Use `npm test` (or other command)

How would we communicate that this option is much slower?

Configuring a test framework doesn't make sense anymore. The test runner does not support test selection. Good point, I will implement a fail fast for this.

nicojs:
I've altered the first question to look like this:

image

This should be clear enough. If you select "command", the test framework question is skipped.

I've created https://github.com/stryker-mutator/stryker/issues/1049 for preset support. Let's not do that in this PR.

nicojs:
I just totally forgot about the timeout functionality. Implemented now.

simondel:
Do you have an open source project where you had these issues? Because then we could try it out :) we are pretty thorough with killing processes.

DiegoRBaquero:
I don't quite remember which one it was exactly, but for sure it's in the top 20 most depended upon npm modules and I think it ran karma. I can't really remember which one it was. I should have documented it, sorry :(

simondel:
No problem :)

simondel:
This code is really clean. Nice job @nicojs! 💖

PR #1055, opened 2018-07-30
Update peerDependency

andy128k:
[According to this](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker-jest-runner/src/jestTestAdapters/JestTestAdapterFactory.ts#L15)


nicojs:
Yes, good point. Thanks for the fix!

PR #1063, opened 2018-08-03
fix(stryker): Clear timeouts so stryker exits correctly

nicojs:


simondel:
Looks good! I reverted your change to the stryker.conf.js of the webpack transpiler because I think you did not intend to commit that.

Issue #1080, opened 2018-08-12
Migrate to Lerna 3


simondel:
Lerna 3 has been released. We should checkout what's different and migrate our project to it.

Stukongeluk:
The result of upgrading to Lerna 3.4.0 and running the scripts defined in the `package.json` didn't result in any errors. (Only a deprecated warning for `commands` in the `lerna.json`).
This is expected, because Stryker isn't using any lerna commands which has had (documented) breaking changes. 

How much further do you want to research the new Lerna before it will be used in Stryker? @simondel

simondel:
I'd like to test a release that also uses conventuonal-commits on both Windows and Linux to ensure a setup like ours still works (upgrading from Lerna 1 to 2 broke this for a while)

nicojs:
> I'd like to test a release that also uses conventuonal-commits on both Windows and Linux to ensure a setup like ours still works (upgrading from Lerna 1 to 2 broke this for a while)

How do you want to do this? Just try it out and see what it does? (living on the edge!)

simondel:
Just like how we tested it for V2, with a demo project

simondel:
@nicojs sneaked this in somewhere.

Issue #1076, opened 2018-08-10
Rename 'project' config key to 'projectType'


simondel:
As mentioned by @nrg500 in https://github.com/stryker-mutator/stryker-handbook/pull/2 the config setting `'project'` in the stryker-karma-runner and stryker-jest-runner can be quite confusing because your project name could be a valid value. A name such as `'projectType'` would probably result in less confusion about the intended use of the property.

**Suggestion:** 
Add a property `'projectType'` and make it the default property to use. 
If a user still uses `'project'`, then the value should be moved to the `'projectType'` property and a deprecation warning should be logged.
The projectType `default` should be renamed to `custom` for the jest-runner. There should be a deprecation warning if the user has manually set the projectType to `default`

The stryker-jest-runner and stryker-karma-runner packages are affected by this. 
It's important to take into account that the [initStrykerConfig](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker-karma-runner/package.json#L55) in the package.json of the two packages may be setting the `project` property. This should be changed to `projectType` if it is present.
When we merge a pull request that resolves this, we should also update the [stryker handbook](https://github.com/stryker-mutator/stryker-handbook)

simondel:
Nice catch! I have updated the main post.

nicojs:
While you're at it, please update the name of `default` to `custom` in the jest-runner. It doesn't make sense to say a `projectType` is `default` imo

kmdrGroch:
@simondel, @nicojs I think it can be closed.

nicojs:
Closed via #1107 thanks again @kmdrGroch  👍

Issue #860, opened 2018-06-06
Rename --timeoutMs => --timeoutMS


nicojs:
Casing should be consistent to normal casing rules: `--timeoutMS`

nicojs:
Yeah, deprecate it first. Remove in 1.0

Issue #654, opened 2018-03-15
Prepare 1.0.0 release


simondel:
With the 1.0.0 release with Stryker we will make the following changes:

### Preparation

- [x] #365
- [x] #667 
- [x] #799 (release v1 of dependencies)
- [x] #920 (support big projects)
- [x] [Rename 'project' config key to 'projectType'](https://github.com/stryker-mutator/stryker/issues/1076)
- [x] [Rename 'timeoutMs' config key to 'timeoutMS'](https://github.com/stryker-mutator/stryker/issues/860)

### Execution

- [x] Remove deprecated Factories from stryker-api.
- [x] Remove ES5 mutator
- [x] Set 'javascript' as the default mutator
- [x] Remove deprecated reporter config key
- [x] Remove deprecated project config key
- [x] Remove deprecated 'default' projectType for jest (in favor of 'custom')
- [x] Remove deprecated timeoutMs config key
- [x] Rename packages:
  * `stryker` => `@stryker-mutator/core`
  * `stryker-xxx-yyy` => `@stryker-mutator/xxx-yyy`
  * `stryker-jasmine` => `@stryker-mutator/jasmine-framework`
  * `stryker-typescript` => `@stryker-mutator/typescript`
- [x] Remove `config` key from being injectable
- [x] Remove `getLogger` and `setImplementation` from `stryker-api/logging`
- [x] Remove all deprecated apis/messages/etc
- [x] Update plugin loader to load plugins from `@stryker-mutator` (instead of `stryker-`)
- [x] Enable lerna versioning mode (not independent, other one)
- [x] Update `stryker-cli` to install `@stryker-mutator/core` instead of `stryker` and `stryker-api`
- [x] Update package.json keywords to @stryker-mutator/
- [x] Update keyword plugin to 'stryker-mutator-plugin'

### Clean up

- [x] Deprecate Stryker packages: https://docs.npmjs.com/cli/deprecate (let's wait ~ a week for this)
- [x] Write awesome blog article
- [x] Update handbook https://github.com/stryker-mutator/stryker-handbook/tree/v1-release
- [x] Update website quickstart: https://github.com/stryker-mutator/stryker-mutator.github.io/tree/v1-release

_This original post will be kept up-to-date._

nicojs:
Why would we no longer support mutators as strings? `mutator: 'typescript'` could be a shorthand for ```js
mutator: {
   type: 'typescript',
   excludedMutations: []
}
```

[...]

PR #1082, opened 2018-08-13
Style fix and some #654 refactors

kmdrGroch:
I suppose i solved:
- [Rename 'project' config key to 'projectType'](#1076)
- [Rename 'timeoutMs' config key to 'timeoutMS'](#860)
- [Migrate to Lerna 3](#1080)
Although You should really take a close look on it since I am not 100% sure.

### DIFF
--- a/packages/stryker/src/config/ConfigValidator.ts
+++ b/packages/stryker/src/config/ConfigValidator.ts
@@ -79,7 +79,11 @@ export default class ConfigValidator {
   }
 
   private validateTimeout() {
-    this.validateIsNumber('timeoutMS', this.strykerConfig.timeoutMS);
+    if (this.strykerConfig.timeoutMs !== undefined) {
+      this.invalidate('[Deprecation warning] use timeoutMS instead of timeoutMs');
+    } else {
+      this.validateIsNumber('timeoutMS', this.strykerConfig.timeoutMS);
+    }
     this.validateIsNumber('timeoutFactor', this.strykerConfig.timeoutFactor);
   }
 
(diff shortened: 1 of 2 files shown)


simondel:
Thanks for making the changes!
I have a couple of remarks:
* Could you revert the Lerna change? Previous releases of the framework have caused quite a lot of trouble and since it's at the core of our release process, we want to prevent mistakes at any cost
* Feel free to ignore the mutators that are in the Stryker package directly from now on as we will completely delete the files once we go to v1.0.0. (that's what we mean by 'remove ES5 mutator)
* Our plan is to deprecate the configuration changes that you made instead of completely removing the old functionality. For example, if someone has "timeoutMs" in their config, they should get a warning telling them to stop using "timeoutMs" and start using "timeoutMS". Stryker should still be able to run without a problem. This makes it easy for users to upgrade to the new version. With 1.0.0 we will remove the support for the old property, forcing people to update the config if they haven't already. For 1.0.0 we will have a guide with all the changes you have to make to upgrade.

kmdrGroch:
I can surely revert Lerna change. Honestly i dunno how to implement these ``deprecation`` warnings at this moment, so now these options are just replaced

kmdrGroch:
@simondel, we could try that way. Also reverting Lerna change right now

simondel:
Awesome :) if you make a new branch with just the karma change and then create a new pull request for it, I'll add the test tomorrow!

kmdrGroch:
@simondel and what if in ``styker/.../ConfigReader`` in ``readConfig`` function add
```js
if (config.timeoutMs) {
      config.timeoutMS = config.timeoutMs;
      console.warn('[Deprecation warning] use timeoutMS instead of timeoutMs');
    }
    if (config.project) {
      config.projectType = config.project;
      console.warn('[Deprecation warning] use projectType instead of project');
    }
```

kmdrGroch:
@simondel you can look how i kinda make it work -> test passes, although now it is just throwing an error (since i didn't know how to test warning message) [this commit is to undo although you can see my vision / method of doing this]

simondel:
@kmdrGroch I have an example of deprecating a property in #1088

kmdrGroch:
I'll be able to take a deeper look when I come back home, so after a week. But it looks nice and replicable.

PR #1088, opened 2018-08-17
feat(stryker config): rename config setting `reporter` to `reporters`

### LINKED ISSUE #793: Rename reporter config property
Every stryker.conf.js has a property called `reporter` which contains an array of reporters. Since this is an array, it would be better for the property to be plural, so `repoters` in this case. 

When someone still has the 'old' `reporter` property instead of the new property, we should put the array on the new property to ensure Stryker will works. We should also log a deprecation warning to inform the user that the `reporter` property will be removed in 1.0.0 and that it should be called `reporters` from now on.

New users should automatically have the new property when the run stryker init. Keep in mind that some plugins may override properties using the `initStrykerConfig` option in the package.json of the package. One of the packages that does this is stryker-typescript (but it does not override the reporters however):
https://github.com/stryker-mutator/stryker/blob/357abb673012ea7c2f3b91291de6e3974dafb95a/packages/stryker-typescript/package.json#L56-L61

simondel:
Deprecates the `reporter` property. Reporters should be set using the `reporters` property and they should be an array (instead of an array or string).

Fixes #793


nicojs:
Ok for me. We could later remove the null check in the broadcast reporter. I'll create an issue for it and other options we know are not null (so we can enable typescripts strict null checks)

simondel:
@nicojs So, merge when green?

simondel:
We will still have to update the strkyer-api dependency version though.

PR #1099, opened 2018-08-21
feat(stryker config): rename config setting `timeoutMs` to `timeoutMS`

kmdrGroch:
It should solve [Rename 'timeoutMs' config key to 'timeoutMS'](#860 )

### DIFF
--- a/packages/stryker/src/config/ConfigReader.ts
+++ b/packages/stryker/src/config/ConfigReader.ts
@@ -43,7 +43,7 @@ export default class ConfigReader {
 
     if (config.timeoutMs) {
       config.timeoutMS = config.timeoutMs;
-      this.log.warn(`DEPRECATED: please change the config setting 'timeoutMs: ${JSON.stringify(config.timeoutMs)}' into 'timeoutMS: ${JSON.stringify(config.timeoutMS)}'`);
+      this.log.warn(`DEPRECATED: please change the config setting 'timeoutMs: ${config.timeoutMs}' into 'timeoutMS: ${config.timeoutMS}'`);
     }
 
     return config;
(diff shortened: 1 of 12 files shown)


packages/stryker/src/config/ConfigReader.ts
```diff
@@ -41,6 +41,11 @@ export default class ConfigReader {
       this.log.warn(`DEPRECATED: please change the config setting 'reporter: ${JSON.stringify(config.reporter)}' into 'reporters: ${JSON.stringify(config.reporters)}'`);
     }
 
+    if (config.timeoutMs) {
+      config.timeoutMS = config.timeoutMs;
+      this.log.warn(`DEPRECATED: please change the config setting 'timeoutMs: ${JSON.stringify(config.timeoutMs)}' into 'timeoutMS: ${JSON.stringify(config.timeoutMS)}'`);
```
simondel: Since timeoutMs and timeoutMS are both number you can use:
```ts
this.log.warn(`DEPRECATED: please change the config setting 'timeoutMs: ${config.timeoutMs}' into 'timeoutMS: ${config.timeoutMS}'`);
```
Else the log message will be `... config setting 'timeoutMs: "5000"' into ...` instead of `... config setting 'timeoutMs: 5000' into ...`

kmdrGroch:
@simondel #1076 is also done but without tests. I dunno how to make them. Should I add it here without tests / create new pull request without tests or leave it to you?

packages/stryker/src/config/ConfigReader.ts
kmdrGroch: I have checked files by ctrl + f -> search in all folders -> "timeoutMs"

kmdrGroch: Now let's see if this build works, and then i'll change this

simondel:
@kmdrGroch Nice! note: the build may still fail because you removed the `"` in your test but it's still in the line of code that logs the warning

kmdrGroch:
@simondel that's why copy-paste is bad >.< final try

simondel:
Awesome job!

Please open a new pull request for #1076. I'll help you with the tests 👍

PR #1100, opened 2018-08-21
chore: add compatibility for stryker-api@0.21.x

simondel:


simondel:
Creates compatibiliy for stryker-api with the renamed timeoutMS property. I created a separate PR for this to be able to release a patch verison of our plugins

Issue #436, opened 2017-10-25
Reach consensus and implement stronger tslint rules


nicojs:
Right now, we're using tslint, but without much rules enabled. We should agree on a good rule set and enforce it by fixing all inconsistencies and enforcing them in the build

mscharley:
[I have opinions on this topic](https://github.com/mscharley/generic-type-guard/blob/master/tslint.json)

I spent way too long with the TSLint and ESLint rule lists.

nicojs:
It's hard to comment on this. Could you create a PR for this? 

Could you try to place this TSConfig in the root of the repo and simply use 

```ts
{
  "extends": [
    "../../tsconfig.json"
  ],
```

Everywhere else?

kmdrGroch:
@nicojs I think it can be closed and removed from ``backlog``

nicojs:
Yes this one can be closed. What solution did you find for forcing an import order?

kmdrGroch:
@nicojs there is an option in tslint, but honestly i dunno how it works (ASCII i suppose, SOMEHOW)

PR #1102, opened 2018-08-22
feat: solve #1076 , #436

kmdrGroch:


kmdrGroch:
Note: there is no deprecation warning.
Note: there are still places where ``projectType`` is said to be ``default`` not ``custom``

simondel:
Hi @kmdrGroch Could you create a separate PR for #1076? That makes it easier to review since there are 3.000 changed lines in this pull request.

This also makes it easier for us to release the pull request since our changelog and version change is based on the commit messages. So if I would merge this one pull request, then every package would go a minor version up and every package will have a changelog entry saying the `project` property has been renamed.

kmdrGroch:
@simondel I'll be able to do it tonight. If it is possible could you first merge this big change? It is mostly style changing then we will be able to work on #1076 easily

simondel:
@kmdrGroch I think it will take me about one week to review the big change.

PR #1105, opened 2018-08-23
Configurable ng test arguments

blabno:
This is a proposition to allow customization of arguments passed to `ng test`. Below is sample conf file:
```
  config.set({
    files: ['*', 'src/**/*'],
    mutate: [
      'src/app/**/*.ts',
      '!src/**/*.unit-spec.ts',
      '!src/**/*.integration-spec.ts',
      '!src/test.ts',
      '!src/environments/*.ts'
    ],
    testRunner: 'karma',
    mutator: 'typescript',
    port: 9336,
    karma: {
      configFile: 'src/karma.conf.js',
      project: 'angular-cli',
      config: {
        browsers: (process.env.KARMA_BROWSERS || 'ChromeHeadless').split(',')
      },
      ngConfig: {
        ngTestArguments: ['--main', 'src/test-unit.ts']
      }
    },
    reporters: ['progress', 'html'],
    coverageAnalysis: 'off',
    maxConcurrentTestRunners: 1,
    fileLogLevel: 'trace'
  });
};
```

### DIFF
--- a/packages/stryker-karma-runner/src/KarmaTestRunner.ts
+++ b/packages/stryker-karma-runner/src/KarmaTestRunner.ts
@@ -26,7 +26,7 @@ export default class KarmaTestRunner implements TestRunner {
       setup.projectType = setup.project;
       this.log.warn('DEPRECATED: `karma.project` is renamed to `karma.projectType`. Please change it in your stryker configuration.');
     }
-    this.starter = new ProjectStarter(setup.projectType);
+    this.starter = new ProjectStarter(setup);
     this.setGlobals(setup, options.port);
     this.cleanRun();
     this.listenToRunComplete();
--- a/packages/stryker-karma-runner/src/StrykerKarmaSetup.ts
+++ b/packages/stryker-karma-runner/src/StrykerKarmaSetup.ts
@@ -6,10 +6,15 @@ export const KARMA_CONFIG_KEY = 'karma';
 
 export type ProjectKind = 'custom' | 'angular-cli';
 
+export interface NgConfigOptions {
+  testArguments?: string[];
+}
+
 export default interface StrykerKarmaSetup {
   // Deprecrated
   project?: ProjectKind;
   projectType: ProjectKind;
   configFile?: string;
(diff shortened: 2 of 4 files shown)


simondel:
@nicojs What are your thoughts on this?

nicojs:
I'm open to the idea of this. However with the current implementation it is possible to break Stryker. For example: I could use ` ngTestArguments: ['--karma-config', 'src/my-karma-config']`. This would be a problem, as Stryker needs it to be `'./stryker-karma.conf'` so it can hooking to Karma. 

@blabno Could we make `ngTestArguments` an object, rather than an array? This: `ngTestArguments: ['--main', 'src/test-unit.ts']` would become `ngTestOptions: { '--main':  'src/test-unit.ts' }`. That way it would be easier to not allow for `--karma-config` using `object.assign`. Pseudo code:

```ts
const FORCED_OPTIONS = Object.freeze({ '--progress': false, '--karma-config': require.resolve('./stryker-karma.conf') });
const options = Object.assign({}, ngConfig.ngTestOptions, FORCED_OPTIONS);
const cliArgs = ['test'].push(... Object.keys(options).map(option => `${option}=${options[option]}`));
```

I'm open for discussion.

blabno:
Of course we can make it an object. I was thinking about it but wanted to get feedback as fast as possible.

Do we want to convert dash case to camel case for the keys?

nicojs:
> Do we want to convert dash case to camel case for the keys?

That would probably be nice, but would also add complexity. What happens if someone uses this?

```ts
ngTestOptions: { '--main': 'foo.ts', main: 'bar.ts' }. 
```

blabno:
It seems that the initial run hangs

nicojs:
@blabno any update on this PR? You want us to help? I don't know why the build is failing, maybe merging master back to your branch will fix it.

nicojs:
@blabno , @YoeriVD finished your work in PR #1273 . Thanks for your work, it is released in stryker-karma-runner v0.22.0.

PR #1107, opened 2018-08-23
feat(stryker config): rename config setting `project` to `projectType` ( #1076)

kmdrGroch:
Please ignore these bigger changes - it is just alphabetic sorting


kmdrGroch:
@nicojs, as i was talking with @simondel, I dunno how to make deprecation warning for this and test for it. I leave it to you, or if you can tell me how to do it, I can make it.

nicojs:
Ok, I'll do one, then you can do the other 👍

nicojs:
As is a habit of mine, I've refactored some things while implementing the deprecation warning in the jest runner (not you're code, mind you). This is what I did:

1. I've added the `logMock` file in the test helpers to have easy access to the current log mock object. This is my preferred way as it keeps things dry. This is not needed for karma, it already has a log mock in place.
1. I've removed the sinon sandboxes all over the place and replace it with the default sinon sandbox (`import * as sinon from 'sinon'` ). Since sinon 5, sinon itself is a sandbox. This is not strictly needed in karma, as we use a custom global `sandbox` object that is available in tests (but you may also refactor that if you want).
1. I've renamed the `'default'` jest project type to `'custom'` (in line with karma's default project type). This was not strictly part of this PR, but I don't want people to have to refactor their stryker configuration twice, so I've decided to also add that to your PR. Hope you don't mind. I've also implemented a deprecation warning for that.

So basically, to add a deprecation warning just use `this.log.warn`. To test that deprecation warning, use `expect(logMock.warn).calledWith('...');` 👍

nicojs:
Thanks! 👍

kmdrGroch:
@nicojs I will PR TSlint changes soon so you can see them :)

PR #1126, opened 2018-09-05
docs: fix and lint-info

kmdrGroch:
- you don't need to run ``npm run build`` before ``npm test`` since ``npm test`` lints, builds, and then tests whole app.
- added ``npm run lint-info`` command which runs tslint

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ We use [Lerna](https://lernajs.io/) to manage the packages in this repository. Y
 
 1. Clone the repository
 1. Install dependencies using `npm install` in the root of the project
-1. Run `npm test` in the root of the project or in one of the package folders
+1. Run `npm test` in the root of the project or in one of the package folders (note: `before` running `npm run test:stryker` and `npm run test:stryker-api` you have to run `npm run build`)
 
 ## VSCode environment configuration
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -19,7 +19,6 @@ We use [Lerna](https://lernajs.io/) to manage the packages in this repository. Y
 
 1. Clone the repository
 1. Install dependencies using `npm install` in the root of the project
-1. Run `npm run build` in the root of the project once.
```
nicojs: True, but the step below specifies to run `npm test` on the root **or in individual packages**. In that last case, you do need to build it once before. I think that deserves an explanation if we're removing that line.

kmdrGroch: Ok, I'll add info that before running `npm run test:stryker` you have to run `npm run build`

kmdrGroch:
@nicojs done :)

CONTRIBUTING.md
```diff
@@ -19,8 +19,7 @@ We use [Lerna](https://lernajs.io/) to manage the packages in this repository. Y
...
-1. Run `npm test` in the root of the project or in one of the package folders
+1. Run `npm test` in the root of the project or in one of the package folders (note: `before` running `npm run test:stryker` and `npm run test:stryker-api` you have to run `npm run build`)
```
nicojs: @kmdrGroch can we remove the `test:stryker` and `test:stryker-api` scripts? Those 2 are maintained in the root package.json, but those are the only once (added once for convenience). Suggestion:

> 1. Run `npm test` in the root of the project or in one of the package folders. Running `npm test` in the root will also build the packages. However, running `npm test` in a child package directory _will not build_ the package, so you probably need to run `npm run build` before doing that.

kmdrGroch: well, sure, but since i have removed that branch i can't edit it :c

nicojs: It seems that you have also removed that repository: `unknown repository` 😅

nicojs:
Ok, I'll close this issue and create a new PR for this.

Issue #1078, opened 2018-08-11
New mutators


kmdrGroch:
### Summary
maybe good mutators could be made of:
- ``>`` -> ``<``
- ``>`` -> ``===`` (etc)
- ``==`` -> ``===`` -> ``==``
- ``!=`` -> ``==`` -> ``!=``
- ``{ some object }`` -> ``{}``
- remove all ``cases`` from ``switch`` and leave only ``default``
-
- Based on https://stryker-mutator.io/mutators.html

simondel:
Thanks for opening an issue for this!
We already support:
- ``>`` -> ``<``
- ``==`` -> ``===`` 
- ``===`` -> ``==``
- ``!=`` -> ``==`` 
- ``==`` -> ``!=``

Mutating case statements and switch statements sounds interesting! I don't really understand what you mean with
- ``>`` -> ``===`` (etc)
Could you be more specific on that one? Why is this a construction that you think should be changed?

kmdrGroch:
@simondel ``etc`` v
- ``<`` -> ``===``
- ``>`` -> ``===``
- ``>=`` -> ``===``
- ``<=`` -> ``===``
It would test another cases.

kmdrGroch:
Also i would like to notice that on your official page -> ``>`` -> ``<``; ``==`` -> ``===`` are not included; https://stryker-mutator.io/mutators.html

simondel:
My bad! We actually don't have these mutators that I claimed we have.
- ``>`` -> ``<`` is already covered by ``>`` -> ``<=``
- ``==`` -> ``===``  will almost never cause an issue and the use of ``==`` is not recommended
- ``===`` -> ``==`` should never cause any issues except for linting

We could add the above mutators (same goes for the ``>`` -> ``===`` etc.) but since they are will almost never cause an issue that is not found using other mutators, I believe it's not the smartest move to add them. They will increase the time it takes to run stryker while not providing a lot of value.

What are your thoughts on this?
ps I still really like the object and switch mutators!

kmdrGroch:
@simondel since ``>`` is changed to ``>=`` or ``<=`` you don't need ``===``. ``==`` -> ``===`` is really dumb (i was writing that at night so i was stupid then xD). You could also try emptying each object like ``Int8Array`` etc but i dunno if it won't cause some issue (or can be hard to implement)

kmdrGroch:
@simondel i think this object replacer would look sth like
```ts
[...]

PR #1127, opened 2018-09-05
feat(mutator): Switch statement mutator

tommilligan:
As suggested by @kmdrGroch in #1078, a mutator which removes all non-`default` cases from a `switch` statement might be interesting.

My first pass is attached - feedback welcome! Working implementation for JS + TS, with tests.

We'd definitely find more mutants disabling `case`s one at a time, but the number of test runs needed would be enormous.

I've only tested with unit tests so far, will look into integration tests next.

### DIFF
--- a/packages/stryker-javascript-mutator/src/mutators/SwitchCaseMutator.ts
+++ b/packages/stryker-javascript-mutator/src/mutators/SwitchCaseMutator.ts
@@ -19,11 +19,11 @@ const sequentialSplices = <T>(collection: T[]): T[][] => {
 /**
  * Represents a mutator which can remove individual clauses from switch statements.
  */
-export default class SwitchStatementMutator implements NodeMutator {
-  public name = 'SwitchStatement';
+export default class SwitchCaseMutator implements NodeMutator {
+  public name = 'SwitchCase';
 
   public mutate(node: types.Node, copy: <T extends types.Node>(obj: T, deep?: boolean) => T): types.Node[] | void {
-    if (types.isSwitchStatement(node)) {
+    if (types.isSwitchCase(node)) {
       // Generate possible case arrays
       const caseSplices = sequentialSplices(node.cases);
       // Map onto cloned nodes
--- a/packages/stryker-javascript-mutator/src/mutators/SwitchCaseMutator.ts
+++ b/packages/stryker-javascript-mutator/src/mutators/SwitchCaseMutator.ts
@@ -3,35 +3,16 @@ import { types } from 'babel-core';
 import NodeMutator from './NodeMutator';
 
 /**
- * Given an array of length n, return an array length n of arrays length n-1
- * where each item has been removed in sequence
- *
- * e.g. [0, 1, 2] -> [[1, 2], [0, 2], [0, 1]]
- */
-const sequentialSplices = <T>(collection: T[]): T[][] => {
-  return collection.map((_, i) => {
-    return collection.filter((_, j) => {
-      return i !== j;
-    });
-  });
-};
-
-/**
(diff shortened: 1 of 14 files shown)


tommilligan:
@leonormes fyi

kmdrGroch:
I don't quite understand it @tommilligan. In 1 test you have
```javascript
it('should mutate the body of a case statement if defined as a block', () => {
      expectMutation('switch (v) { case 42: { a = "spam"; break; } }', 'switch (v) { case 42: {} }');
});
```
and in second
```
it('should remove non-default case', () => {
      expectMutation('switch (v) { case 42: a = "spam";break; }', 'switch (v) {}');
});
```

kmdrGroch:
I also request changes for ``packages/stryker-mutator-specification/src/SwitchStatementMutatorSpec.ts`` to standardize code style.
You wrote
``expectMutation('switch (v) { case 42: a = "spam";break; }', 'switch (v) {}');``
then
``expectMutation('switch (v) {default: a = "spam";}');``
then
``'switch (v) { case 0: a = "foo"; case 1: a = "qux"; break; default: a = "spam";}',``
``'switch (v) {default: a = "spam";}'``

simondel:
Hi @tommilligan thanks for the pull request! I'll have a look at it today :)

kmdrGroch:
After he changes it @simondel if you like this mutator it's green to merge :)

tommilligan:
I've made the spacing as uniform as I can while not breaking the tests - the `js` tests rely on string comparison which is whitespace sensitive. In my opinion the spacing is reasonable, readable and passes linting rules.

If you have a **specific** suggestion of which spacing to change, please do send me a patch.

kmdrGroch:
@tommilligan I've made PR. Now all files of your are in same convention :)

kmdrGroch:
ok, since it has to be spaced like this. @simondel ready to merge

simondel:
My preference would go out to removing the switch cases one at a time. 

Yes, this will create more mutants so it will take longer to run stryker, but the bennefit is also larger.

There is quite a big chance that you'll test at least one of the switch. If you remove them all at once, a test will fail but you won't know if you actually tested the cases. You only know what happens when you replace your switch statement with the code in the default statement.

I'd also be a fan of removing the `default` case as well in one test. That way you also know if your tests pass if you remove your default case.

kmdrGroch:
sounds nice <3

tommilligan:
Updated implementation - creates an array of mutants from a `switch` statement, by removing each `case` (including `default`) sequentially.

tommilligan:
@simondel this is ready for you to review again

kmdrGroch:
@tommilligan it usually takes them a while -> 1day - week. Be patient :)

simondel:
@tommilligan thanks! I'll see if I can make some time this weekend :)

kmdrGroch:
@simondel there is also [stryker-cli PR](https://github.com/stryker-mutator/stryker-cli/pull/31) waiting for you :)

simondel:
The SwitchStatementMutator was not yet registered in the javascript mutator package. I fixed that with the latest commit.  The code looks good, but fixing survived mutants is a bit difficult since the change is made on the entire switch case:
image

How do you think we could fix this?

tommilligan:
@simondel thanks for fixing the registration issue.

Updated to edit case nodes individually - the report output is a lot clearer. Tested on the `stryker` package itself:

switchcase

Any further work needed before master?

simondel:
@tommilligan thanks! I tested the TypeScript implementation yesterday and that worked like a charm 😻 I'll test the JavaScript implementation today.

simondel:
Looks great! Thanks for the hard work! 🥇

PR #1132, opened 2018-09-10
Add ^7.0.0-bridge.0 to babel-core range

hutchy2570:
We're in the process of upgrading to Babel 7 and have hit an issue with stryker-babel-transpiler. It loads babel-core 6.

This PR adds ^7.0.0-bridge.0 to the babel-core peer range, as recommended by Babel https://github.com/babel/babel-bridge

### DIFF
--- a/packages/stryker-babel-transpiler/README.md
+++ b/packages/stryker-babel-transpiler/README.md
@@ -39,7 +39,8 @@ $ stryker run
 ## Peer dependencies
 The `stryker-babel-transpiler` plugin requires the following packages to be installed in order to work: 
 * `stryker-api`
-* `babel-core`
+* `babel-core` (When using babel 6)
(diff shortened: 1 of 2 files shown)


nicojs:
Hmm interesting. @simondel, you were busy with babel7 support. Should we use babel bridge?

simondel:
@nicojs I'm not sure. I haven't heard about babel bridge. I'm not sure if we can merge this PR yet because the integration tests still use Babel 6 so we don't actually know if our code works with Babel 7.

@hutchy2570 Did you have any issues with Stryker running with Babel 7 code?

hutchy2570:
@simondel no issues.  We've been using babel-bridge with other projects internally which have been migrated to babel-7 with other dependencies which have the bridge version in their ranges.

nicojs:
@simondel what are we doing? Updating to babel 7? Or implementing the bridge?

PR #1134, opened 2018-09-10
refactor: stricter types

kmdrGroch:
I have removed a lot of ``any`` types as I fixed one function: ``isNumber`` where ``NaN`` could be an edge case. (I see Travis doesn't work, but i have tested it locally and no issue has been found [even on windows])


kmdrGroch:
@simondel I think it didn't work ._.

simondel:
Looks good to me! Thanks :)

PR #1145, opened 2018-09-14
fix: log MODULE_NOT_FOUND when the karma.conf.js does not exist

### LINKED ISSUE #1098: stryker-karma-runner logs MODULE_NOT_FOUND when the karma.conf.js does not exist
When I set an incorrect path to my karma.conf.js file or if I don't copy the karma.conf.js file to my sandbox. I  see the following logging:

```
09:34:03 (84321) INFO stryker-karma.conf.js Importing config from "C:\dev\myapp\.stryker-tmp\sandbox5403836\karma.conf.js"
09:34:03 (84321) ERROR stryker-karma.conf.js Could not read karma configuration from karma.conf.js. { code: 'MODULE_NOT_FOUND' }
```

It would be good to check if the file exists in the sandbox before trying to require it.

We could use fs.access before trying to require the file (only then with a promise):
```js
fs.access(file, fs.constants.F_OK, (error) => {
  if (error) {
    log.error(`Unable to find karma config at "${configFileName}". Please check your stryker config.`);
  }
});
```

kmdrGroch:
Can fix #1098

### DIFF
--- a/packages/stryker-karma-runner/src/starters/stryker-karma.conf.ts
+++ b/packages/stryker-karma-runner/src/starters/stryker-karma.conf.ts
@@ -15,13 +15,17 @@ function setDefaultOptions(config: Config) {
 function setUserKarmaConfigFile(config: Config, log: Logger) {
   if (globalSettings.karmaConfigFile && typeof globalSettings.karmaConfigFile === 'string') {
     const configFileName = path.resolve(globalSettings.karmaConfigFile);
-    log.debug('Importing config from "%s"', configFileName);
     try {
       const userConfig = requireModule(configFileName);
+      log.debug('Importing config from "%s"', configFileName);
       userConfig(config);
       config.configFile = configFileName; // override config to ensure karma is as user-like as possible
     } catch (error) {
-      log.error(`Could not read karma configuration from ${globalSettings.karmaConfigFile}.`, error);
+      if (error.code === 'MODULE_NOT_FOUND') {
+        log.error(`Unable to find karma config at "${globalSettings.karmaConfigFile}". Please check your stryker config.`);
+      } else {
+        log.error(`Could not read karma configuration from ${globalSettings.karmaConfigFile}.`, error);
+      }
     }
   }
 }


simondel:
Thanks for making a pull request for this!

packages/stryker-karma-runner/src/starters/stryker-karma.conf.ts
```diff
@@ -15,13 +15,17 @@ function setDefaultOptions(config: Config) {
 function setUserKarmaConfigFile(config: Config, log: Logger) {
   if (globalSettings.karmaConfigFile && typeof globalSettings.karmaConfigFile === 'string') {
     const configFileName = path.resolve(globalSettings.karmaConfigFile);
-    log.debug('Importing config from "%s"', configFileName);
     try {
       const userConfig = requireModule(configFileName);
+      log.debug('Importing config from "%s"', configFileName);
```
simondel: Is there a reason why you moved this logging statement?

kmdrGroch: I wanted to test it, but I didn't manage to do it :(

kmdrGroch:
I wanted to test this code too, but I didn't manage to do it :(

packages/stryker-karma-runner/src/starters/stryker-karma.conf.ts
simondel: Yes, I think it would be smart to still log that we're going to try to read the config file

kmdrGroch: Sure. Changing it back right now

packages/stryker-karma-runner/src/starters/stryker-karma.conf.ts
```diff
@@ -21,7 +21,11 @@ function setUserKarmaConfigFile(config: Config, log: Logger) {
       userConfig(config);
       config.configFile = configFileName; // override config to ensure karma is as user-like as possible
     } catch (error) {
-      log.error(`Could not read karma configuration from ${globalSettings.karmaConfigFile}.`, error);
+      if (error.code === 'MODULE_NOT_FOUND') {
+        log.error(`Unable to find karma config at "${globalSettings.karmaConfigFile}". Please check your stryker config.`);
```
nicojs: Could we add `configFileName` to the mix?

`Unable to find karma config at "${globalSettings.karmaConfigFile}" (tried to load from ${configFileName}}. Please check your stryker config. You might need to make sure the file is included in the sandbox directory.`

nicojs:
@kmdrGroch 

> I wanted to test this code too, but I didn't manage to do it :(

Shall I help with that? That way you can see how to do it in the future 👍

kmdrGroch:
@nicojs sure. I know a bit how to write tests (even with mocha), but i couldn't write working one ._. Or I tested bad thing

nicojs:
I've added the test. Can you make sense of it? 

Also fixed a new compile error.

kmdrGroch:
I would never think it should have been make like that O.o 
basically i made sth like that @nicojs 
```javascript
const expectedKarmaConfigFile = 'foobar.conf.js';
    sut.setGlobals({ karmaConfigFile: expectedKarmaConfigFile });
    sut(config);
    expect(logMock.error).calledWith(`Unable to find karma config at "foobar.conf.js" (tried to load from ${path.resolve(expectedKarmaConfigFile)})`);
```
and it didnt work. I need to read more about mocks i suppose. Or practice makes master. :)

Issue #1156, opened 2018-09-22
SwitchCase mutation mishandles empty fallthrough cases


CheshireSwift:
### Summary

When handling an empty fallthrough case, e.g.:

```js
switch (value) {
  case 'A':
  case 'B':
    ...
}
```

SwitchCase will raise an unchanged (thus unkillable) mutant which simply removes + readds the `case 'A'` (maybe changing the formatting slightly, but not the semantics).

### Stryker config

```js
module.exports = function(config) {
  config.set({
    mutator: "javascript",
    packageManager: "npm",
    reporters: ["clear-text", "progress"],
    testRunner: "jest",
    transpilers: [],
    coverageAnalysis: "off",
    mutate: ["src/**/*.js"]
  });
};
```

### Stryker environment

```
stryker@0.29.2
stryker-api@0.21.1
stryker-javascript-mutator@0.10.0
stryker-jest-runner@1.2.1
```

```
jest@23.6.0
```

### Your Environment

| software         | version(s)
| ---------------- | -------  
| node             | multiple
| npm              | multiple
| Operating System | multiple

### Add stryker.log

[stryker.log](https://github.com/stryker-mutator/stryker/files/2408130/stryker.log)

Executable repro repo [here](https://github.com/CheshireSwift/stryker-switchcase-repro).

nicojs:
Thanks for reporting. This is most definitely a bug. It is recently introduced as this is a new mutator. @tommilligan would you mind taking a look? Maybe if you have some time, fixing this issue? Thanks for any feedback you might have.

tommilligan:
No problem, will put in a fix this evening. Thanks for letting me know

kmdrGroch:
@nicojs I think it is ready to close

Archcry:
Closing this issue, when you encounter further problems feel free to reopen.

PR #1159, opened 2018-09-25
fix(mutator): Fix empty case statement unkillable mutant

tommilligan:
Hotfix for #1156: if a case clause contains no statements, do not mutate it.

---

I think a better fix (technically) would be to always remove the case clause altogether. However this was discounted as part of #1127 due to it producing unclear reporter output - stryker then considers the entire `switch` block mutated, rather than the single `case` clause.

One possible alternative is ensuring the case is never matched, rather than removing the statments:
```
case "A":
case "B":
  console.log("Do some actions");
...

// becomes
case "stryker/expression-that-is-never-matched/40ed2198-a801-40fe-8edd-0531783bc069":
case "B":
  console.log("Do some actions");
```

This would pick up mutants for:
- the case clause never being matched
- the case clause resulting action not being tested

Downside is, I don't think this is immediately readable/clear to the user. Unlike the `IfStatement` mutator, we we can't simply replace it with `case false:` or `case null:`, as these are valid expressions that could be easily matched.

### DIFF
--- a/packages/stryker-javascript-mutator/src/mutators/SwitchCaseMutator.ts
+++ b/packages/stryker-javascript-mutator/src/mutators/SwitchCaseMutator.ts
@@ -10,9 +10,12 @@ export default class SwitchCaseMutator implements NodeMutator {
 
   public mutate(node: types.Node, copy: <T extends types.Node>(obj: T, deep?: boolean) => T): types.Node[] | void {
     if (types.isSwitchCase(node)) {
-      const mutatedNode = copy(node);
-      mutatedNode.consequent = [];
-      return [mutatedNode];
+      // if not a fallthrough case
+      if (node.consequent.length > 0) {
+        const mutatedNode = copy(node);
+        mutatedNode.consequent = [];
+        return [mutatedNode];
+      }
     }
   }
 }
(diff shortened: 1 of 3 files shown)


kmdrGroch:
@tommilligan in ``packages/stryker-javascript-mutator/src/mutators/SwitchCaseMutator.ts`` can't you make
```javascript
if (types.isSwitchCase(node) && node.consequent.length > 0) {
        const mutatedNode = copy(node);
        mutatedNode.consequent = [];
        return [mutatedNode];
}
```
instead of
```javascript
if (types.isSwitchCase(node)) {
      // if not a fallthrough case
      if (node.consequent.length > 0) {
        const mutatedNode = copy(node);
        mutatedNode.consequent = [];
        return [mutatedNode];
      }
}
```
?

and in ``packages/stryker-typescript/src/mutator/SwitchCaseMutator.ts``
you don't need ``else``. You can just ``return []`` like so:
```javascript
protected identifyReplacements(node: ts.CaseOrDefaultClause, sourceFile: ts.SourceFile): NodeReplacement[] {
    // if not a fallthrough case
    if (node.statements.length > 0) {
      const clause = isDefaultClause(node)
        ? ts.createDefaultClause([])
        : ts.createCaseClause(node.expression, []);
      const replacement = printNode(clause, sourceFile);
      return [{ node, replacement }];
    }
    return [];
  }
```

tommilligan:
@kmdrGroch indeed I could. However this is a stylistic choice and not covered by the test or lint suite.

Do you have any comments on the functionality of the code?

nicojs:
Thanks for fixing 🥇 , I'll take a look (and probably merge) tomorrow.

Issue #1140, opened 2018-09-12
Filenames not showing up correctly in report


Archcry:
### Summary
In one of my recent projects I used Stryker to mutation test. The report does not seem to be correct though because some wierd glitch is showing up in the filenames.
![screenshot_20180912_065312](https://user-images.githubusercontent.com/1557855/45402940-89330680-b658-11e8-8ff5-435cc22ee124.png)

`modules/helloController.ts` is not working correctly here. My directtory structure looks like:
![screenshot_20180912_065654](https://user-images.githubusercontent.com/1557855/45403095-10807a00-b659-11e8-923d-d7ad85c3b194.png)

### Stryker config

```js
module.exports = function(config) {
  config.set({
    testRunner: "mocha",
    mutator: "typescript",
    transpilers: ["typescript"],
    reporters: ["html", "clear-text", "progress"],
    packageManager: "npm",
    testFramework: "mocha",
    coverageAnalysis: "perTest",
    tsconfigFile: "tsconfig.json",
    mutate: ["src/**/*.ts"],
    coverageAnalysis: "off"
  });
};

```

### Stryker environment

```
+-- stryker@0.29.1
+-- stryker-api@0.21.0
+-- stryker-html-reporter@0.16.1
+-- stryker-mocha-framework@0.12.1
+-- stryker-mocha-runner@0.14.1
+-- stryker-typescript@0.13.2
```

```
+-- mocha@5.2.0
+-- typescript@3.0.3
```

### Reproduction steps
```bash
git clone git@github.com:Archcry/inversify.git
git checkout feat-upgrade-inversify # Might merge into master soon
cd inversify
npm i
npm test
```

### Your Environment

| software         | version(s)
| ---------------- | -------  
| node             | v10.9.0
| npm              | 6.4.1
| Operating System | Linux

Archcry:
Basically the same, and it cannot sure the contents of the file.

simondel:
Does the path exist in the reports folder or is that one also incorrect?

Archcry:
`reports/mutation/html/modules/helloController.ts/` exists in the reports folder. But I think the `html-reporter` is only reporting what Stryker told it to report. I assume something is going wrong when creating the data for all reports because all reporters seem to have this problem.

nicojs:
[...]

PR #1165, opened 2018-10-01
fix(ScoreResultCalculator): fix faulty filenames in stryker score result

Archcry:
The filter method used causes a side-effect that creates a corrupt path, I solved this by short circuiting the filter method so it only returns the elements up untill the element that does no longer match.

solves https://github.com/stryker-mutator/stryker/issues/1140


simondel:
Thanks! I'll look at this tomorrow morning

simondel:
Looks good to me!

Issue #1057, opened 2018-07-31
Inconsistency with object keys


andy128k:
Semantically `{ "key": "value" }` and `{ key: "value" }` are the same, but Stryker creates 2 mutants for former and only one for latter.

So it is possible to hide uncovered code (bug?) from Stryker just by removing quotes.

nicojs:
Hmm good catch. I think this is just an oversight. @simondel, I think this mutant is a false positive, right?

simondel:
What we need to check here is if the parent of the `StringLiteral` node is a `PropertyAssignment` and the current node is the `name` property of that parent. (Names of types/properties may differ between Babylon and TypeScript.

We should add this to the mutator specification so both mutators have to comply with it.

@andy128k Would you be interested in making this change? We can help to guide you to the right places if you want to pick it up.

nicojs:
Sounds good. Don't forget:

```js
{ ["key"]: "value" }
```

Let's make sure the following keeps working though:

```
var a = 3;
{ ["key" + a]: "othervalue" }
{ [`test${a}`]: "othervalue" }
```

roelandvanbatenburg:
This is resolved and can be closed.

nicojs:
Thanks @roelandvanbatenburg

PR #1169, opened 2018-10-02
feat(mutator): Object literal mutator

tommilligan:
As suggested by @kmdrGroch in #1078, a mutator which removes all properties from an object literal during initialization.

Working TS + JS implementation, please let me know any additional cases/requirements you'd like me to test.

### DIFF
--- a/packages/stryker-mutator-specification/src/ObjectMutatorSpec.ts
+++ b/packages/stryker-mutator-specification/src/ObjectMutatorSpec.ts
@@ -1,0 +1,21 @@
+import { expect } from 'chai';
+import ExpectMutation from './ExpectMutation';
+
+export default function ObjectMutatorSpec(
+  name: string,
+  expectMutation: ExpectMutation
+) {
+  describe('ObjectMutator', () => {
+    it('should have name "Object"', () => {
+      expect(name).eq('Object');
+    });
+
+    it('should empty an object declaration', () => {
+      expectMutation('const o = { foo: "bar" }', 'const o = {}');
+    });
+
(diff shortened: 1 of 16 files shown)


kmdrGroch:
@tommilligan is it removing all props at the same time or removes one by one?

kmdrGroch:
And covers objects like shown in https://github.com/stryker-mutator/stryker/issues/1057

tommilligan:
All props simultaneously. I would expect other mutators to mutate the values if necessary (e.g. if a property is a string, for `StringMutator` to pick it up).

packages/stryker-mutator-specification/src/ObjectMutatorSpec.ts
```diff
@@ -0,0 +1,21 @@
+import { expect } from 'chai';
+import ExpectMutation from './ExpectMutation';
+
+export default function ObjectMutatorSpec(
+  name: string,
+  expectMutation: ExpectMutation
+) {
+  describe('ObjectMutator', () => {
+    it('should have name "Object"', () => {
```
nicojs: `"Object"` mutator sounds pretty generic. I think `ObjectLiteral` would be a better fit.

nicojs: Could we also add a test for:

```js
{ ['key']: 'value' } // mutated to { }
```

What do you think?

tommilligan: @nicojs good point on the name - have amended both points

nicojs:
Thank @tommilligan 👍

PR #1170, opened 2018-10-02
fix(Sandbox): Timeout sandbox creation

### LINKED ISSUE #621: stryker-karma-runner hangs when Karma can't connect to the browser
### Summary

When running stryker with karma, stryker will hang forever during the initial test run if karma fails to connect to the browser. This is most easily tested by running stryker in an environment that proxies all HTTP calls.

This isn't a huge issue, if you turn on log level trace, you can see karma launcher say "ChromeHeadless failed x times (timeout). Giving up.", and then it hangs. The solution to the karma launcher not connecting is to make sure you `export NO_PROXY='xxx'`. However, stryker-karma-runner should exit on a launcher failure instead of waiting forever.

### Stryker config

Taken from the Angular Stryker Example project found [here](https://github.com/nicojs/angular-stryker-example)

```js
const getAppFromConfig = require('@angular/cli/utilities/app-utils').getAppFromConfig;
const appConfig = getAppFromConfig();

  config.set({
    files: [
      // Exclude files so we can include only particular files later on
      '!./**/*.ts',
      '!./**/*.js',
      `!./${appConfig.root}/**/*.d.ts`,

      // Include source files and flag them for mutation
      { pattern: `./${appConfig.root}/**/*.ts`, included: false, mutated: true },

      // Exclude test files to later include them again without the mutated flag set to true
      `!./${appConfig.root}/**/*.spec.ts`,
      `!./${appConfig.root}/test.ts`,
      { pattern: `./${appConfig.root}/**/*.spec.ts`, included: false, mutated: false },
      { pattern: `./${appConfig.root}/test.ts`, included: false, mutated: false },

      // Ignore environment configuration files
      `!./${appConfig.root}/environments/*.ts`,

      // Other assets (HTML and CSS for the browser)
      `./${appConfig.root}/**/*.html`,
      `./${appConfig.root}/**/*.css`,
    ],
    testRunner: 'karma',
    mutator: 'typescript',
    transpilers: ['webpack'],
    reporter: ['html', 'clear-text', 'progress'],
    testFramework: 'jasmine',
    coverageAnalysis: 'off',
    karmaConfig: {
      frameworks: ['jasmine'],
      browsers: ['ChromeHeadless'],
      plugins: [
        require.resolve('karma-jasmine'),
        require.resolve('karma-chrome-launcher')
      ]
    },
    tsconfigFile: 'tsconfig.json',
    webpack: {
      configFile: 'webpack-stryker.conf.js'
    },
    logLevel: 'info',
    timeoutMs: 10000
  });
};
```

### Stryker environment

Same as the example project.

```
stryker@0.19.2
stryker-api@0.13.0
stryker-html-reporter@0.12.1
stryker-karma-runner@0.12.2
stryker-typescript@0.9.1
stryker-webpack-transpiler@0.1.4
```

### Your Environment
| ---------------- | -------  
| node             | 8.9.1
| npm              | 5.5.1
| Ubuntu | 16.04

tommilligan:
Currently we set a timeout for test runs inside the sandbox, but do not set a timeout for sandbox creation. This can lead to stryker hanging forever (e.g. with a bad `karma` configuration).

This PR implements a naive timeout on sandbox creation. If preferred I can use an external module/factor out to a util function for reuse.

Closes #621

### DIFF
--- a/packages/stryker/src/Sandbox.ts
+++ b/packages/stryker/src/Sandbox.ts
@@ -5,7 +5,7 @@ import * as mkdirp from 'mkdirp';
 import { RunResult, RunnerOptions } from 'stryker-api/test_runner';
 import { File } from 'stryker-api/core';
 import { TestFramework } from 'stryker-api/test_framework';
-import { wrapInClosure, normalizeWhiteSpaces } from './utils/objectUtils';
+import { wrapInClosure, normalizeWhiteSpaces, timeout, unwrapTimeout } from './utils/objectUtils';
 import TestRunnerDecorator from './test-runner/TestRunnerDecorator';
 import ResilientTestRunnerFactory from './test-runner/ResilientTestRunnerFactory';
 import { TempFolder } from './utils/TempFolder';
@@ -46,14 +46,13 @@ export default class Sandbox {
   public static create(options: Config, index: number, files: ReadonlyArray<File>, testFramework: TestFramework | null, timeoutOverheadMS: number, loggingContext: LoggingClientContext)
     : Promise<Sandbox> {
     const sandbox = new Sandbox(options, index, files, testFramework, timeoutOverheadMS, loggingContext);
-    return new Promise((resolve, reject) => {
-      // Wait for our sandbox to set up
-      sandbox.initialize().then(() => resolve(sandbox));
-
-      // and throw an error if it never does (e.g. karma server hang)
-      // https://github.com/stryker-mutator/stryker/issues/621
-      setTimeout(() => reject(new Error('Sandbox creation timed out')), SANDBOX_CREATE_TIMEOUT);
-    });
+    // throw an error if our sandbox is never set up (e.g. karma server hang)
+    // https://github.com/stryker-mutator/stryker/issues/621
+    return timeout<Sandbox>(
+      sandbox.initialize().then(() => sandbox),
+      SANDBOX_CREATE_TIMEOUT
+    )
+      .then(result => unwrapTimeout(result, 'Sandbox creation timed out'));
   }
 
   public run(timeout: number, testHooks: string | undefined): Promise<RunResult> {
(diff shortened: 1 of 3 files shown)


simondel:
Thanks for creating a fix for this! Would it be possible for you to add some tests so we know if the code works as expected? (And will keep working)

tommilligan:
I've refactored to use the exiting `timeout` implementation in `utils`, and factored out a generic error thrower for timeouts as well. All new code is covered :slightly_smiling_face:

nicojs:
I'm actually not sure about this feature.

I think that the creation of a sandbox should always work or fail. It should never _hang_. If karma config is bad, the `KarmaTestRunner` should fail. If it results in a timeout, people might think there is something wrong with Stryker, while in fact their karma config is corrupt. Do you have another example of a test runner that should result in a timeout?

Can we create reproduction steps for your karma config problem? That way we can start to find ways of fixing it.

tommilligan:
@nicojs I see what you're saying, but the current behaviour *is* for Stryker to hang forever (as a result of karma hanging). I thought the current behaviour *was* a bug in stryker.

There are a couple of cases listed in #621 with reproduction details.

I think the best case might be to:
- timeout
- throw a message like `Sandbox creation failed. Check your test runner [karma] config.`
- expose captured logging from `ChildProcessProxy`, which should actually be helpful

nicojs:
Yeah, the biggest problem with timeouts is that you never know how much is enough. For example, bootstrapping an angular project might actually take a minute or 2. However, a nodejs mocha test runner is instantly done.

> I think the best case might be to:
> 
> * timeout
> * throw a message like Sandbox creation failed. Check your test runner [karma] config.
> * expose captured logging from ChildProcessProxy, which should actually be helpful

Well.. maybe. But the stryker-karma-runner would be the place to do it.

simondel:
We've decided not to merge this for now.

PR #1175, opened 2018-10-06
feat(babel): upgrade babel 6 ~> 7

### LINKED ISSUE #1172: Stryker and BigInt notation
### Summary

Stryker crashes while parsing BigInt notation, such as `4n`.

It looks like stryker-javascript-mutator uses Babel 6, while Babel 7 is required to support such BigInt notation (through @babel/plugin-syntax-bigint).

My project does contain the proper version of Babel, and the proper .babelrc file, but these are ignored. I also tried stryker-babel-transpiler as it lets me explicitly specify Babel as the transpiler, including linking towards my .babelrc file. However even this package, while it uses my .babelrc file, does not use my Babel installation, meaning it's trying to run Babel v6 with my Babel v7 config (which obviously fails).

This points to a bigger problem where Stryker can't apply its JS mutations without Babel, yet doesn't use the Project's own Babel and config, and doesn't provide the required hooks to properly configure Babel either. Essentially, any code that requires a different Babel version than is being used by the aforementioned packages, will simply crash Stryker.

### Stryker config

```js
module.exports = (config) => {
    config.set({
        mutate: ['src/**/solution.js'],
        testRunner: 'command',
        coverageAnalysis: 'off',
        mutator: 'javascript',
        transpilers: [],
        reporters: ['clear-text', 'progress'],
        packageManager: 'npm',
    });
};
```

### Stryker environment

```
+-- stryker@0.29.5
+-- stryker-api@0.21.2
+-- stryker-javascript-mutator@0.10.2
+-- stryker-jest-runner@1.2.2
```

```
+-- jest@23.5.0
```

### Your Environment

| ---------------- | -------  
| node             | 10.9.0
| npm              | 6.4.1
| Operating System | Windows 10

### Add stryker.log

[stryker.log](https://github.com/stryker-mutator/stryker/files/2447102/stryker.log)

nicojs:
Upgrade babel 6 ~> 7. This also adds the the [bigInt](https://babeljs.io/docs/en/babel-parser#ecmascript-proposals-https-githubcom-babel-proposals) plugin so it is now possible to mutate code with the BigN notation (for example `4n`).

Fixes #1172


simondel:
Could you add a test for the BigN notation? I don't think that works yet because we didn't alter the `plugins` for the parser.

nicojs:
Awesome remark @simondel, I was a bit fast in declaring that it should work 😅  

Wat is the idea for the babel plugins? Are we going to make that configurable? I know there was some talk about that a while back, but it wasn't needed in the end?

Next up is support for babel 7 in the stryker-babel-transpiler. Let's do that in a different PR

simondel:
There are no plans to make the plugin system link to your Babel config unfortunately :( (this is from Babel's perspective, not ours). 

I think we should simply add the plugins when people run into an issue.

nicojs:
So this can be merged?

simondel:
Looking good!

PR #1179, opened 2018-10-09
feat(karma-runner): Support angular public api

nicojs:
Add support for angular's public API. It is save to merge, the public api was already there since angular 6.0, but never documented. The private API we're now using will be **removed** when https://github.com/angular/angular-cli/pull/12499 is merged.


simondel:
Nice find!

Issue #975, opened 2018-07-12
Use prettier colors on the CLI reporter?


JoshuaKGoldberg:
### Summary

https://twitter.com/_nicojs/status/1017154960834945024

Referring to colored output in terminals:

> We should use that in #StrykerMutator's clear text console reporter 😎

nicojs:
I was actually referring to the deep linking. It would be awesome if you could click to the exact location where the mutant is located. Pretty colors are nice as well, we already have some colors I think, but as I said in the twitter chat, it might all use a make over 💄. You may work your magic by all means 👍 

The code is already prepared to write to file as well (it uses an output stream). Not sure if anyone is waiting for that feature, but if you want, you can add it as well. When writing to file we should remove color codes I guess.

The console report is also pretty verbose, which bothers me a little. It prints several lines per mutant. Not sure if others agree @stryker-mutator/contributors ?

JoshuaKGoldberg:
Is there a dedicated emoji for scope creep? 🐍 perhaps?

What is the total feature set discussed here?

💄 Color make over for the CLI if in tty mode?
🔗 Links to original file lines in output?
👽 Emoji decorations in output if in a terminal that supports them?

nicojs:
> Is there a dedicated emoji for scope creep?

I think you just coined the emoji: 🐍 it is! 🥇

>  💄 Color make over for the CLI if in tty mode?

Yes. Well... I'm not sure if tty is the correct check to be making. Non-tty can support colors and emoj.

For example:

In vscode debugger terminal:
![image](https://user-images.githubusercontent.com/1828233/42642057-43dc7fa4-85f6-11e8-9bf8-9a364276ca82.png)

In [conemu](https://conemu.github.io/):

![image](https://user-images.githubusercontent.com/1828233/42642097-5ca7d506-85f6-11e8-8c3a-01552df27a06.png)
 
In command prompt:

![image](https://user-images.githubusercontent.com/1828233/42642117-6a802c8c-85f6-11e8-8e61-2be7f9a8f9e1.png)

>🔗 Links to original file lines in output?
> 👽 Emoji decorations in output if in a terminal that supports them?

If that is feasible for you :)

nicojs:
There is an issue for emoji support in Windows console: https://github.com/Microsoft/console/issues/190

It seems that there already is _some_ support. So no need to actively check for support I think

PR #1185, opened 2018-10-15
Added colors to file names and line+col numbers in clear-text reporter

JoshuaKGoldberg:
Same cyan & yellow as TypeScript. It seems to work well there. Removes the extra lines of output as an opt-out behavior to make output more concise.

Starts on #975 _(fancy emojis can come next!)_

Before:

image

After:

image


JoshuaKGoldberg:
I'm a little lost trying to get these unit tests to work with `chalk` 😕... if someone could help, I'd appreciate it!

nicojs:
Hmm looks better I think. I'll have a look at the tests.

@simondel what do you think? It removes the names of the tests that ran, but do we really need them? It's a nice feature, but it also takes up a lot of space.

JoshuaKGoldberg:
I also wonder if we really need the `Mutant survived!` complaint on top of each file. How about a `12/345 mutants survived!` once before the list?

nicojs:
It is a lot of repeated text. It would be awesome if could at a remark in front of the mutator:

```
1. [Survived] StringLiteral
/tmp/cmp/success.component.js:5:5
-    controllerAs: 'cpt'
+    controllerAs: ''
2. [No coverage] BinaryOperator
/tmp/cmp/success.component.js:10:5
-    1+1
+    1-1
```

I would just remove the `Mutant survived!` entirely. I'm pretty sure it is remnant from the original code base, when @simondel was still a student :). A blast from the past.

JoshuaKGoldberg:
If you don't mind, I think we could also remove the `Mutator: *Mutator` message too. Doesn't seem too useful to end users, and we can generally figure it out from the change. Removed in the latest commit.

simondel:
RIP *Mutant Survived!*
2015 - 2018

nicojs:
> If you don't mind, I think we could also remove the `Mutator: *Mutator` message too. 

This is actually useful info. If you want to disable a mutator, you need to know the name.

nicojs:
I'm thinking of adding a global option for `allowColor`. Instead of one specific for the clear text reporter. Might be useful for the log4js formatting as well. @JoshuaKGoldberg is it OK if I move it to a separate issue?

nicojs:
How about this?

image

JoshuaKGoldberg:
> global option for `allowColor`

That sounds nice... though what if we want to have one reporter in color and another without? Would the `allowColor` set what the default is, as overridden by reporters?

I'll update this PR for the `105. [NoCoverage] Block`, and if your separate issue / branch is different and this PR gets closed, I won't be offended in the slightest. 😊

nicojs:
@JoshuaKGoldberg I've created a separate issue for it. Could you explain it further there? #1218

Issue #1133, opened 2018-09-10
strange part of code


kmdrGroch:
I have been looking through the code and I noticed some strange code:
```typescript
function isPrimitive(value: any) {
  return value !== 'object' && !Array.isArray(value);
}
```
shouldn't it be sth like:
```typescript
function isPrimitive(value: any) {
  return typeof value !== 'object' && !Array.isArray(value);
}
```
+ value can also be ``typeof function`` so all cases haven't been checked. (if you don't count NaN as a primitive then also NaN should be checked)

path: stryker\packages\stryker\test\helpers\producers.ts

so final code i suppose should be is like:
```typescript
function isPrimitive(value: any) {
  if(typeof value === 'number' && isNaN(value)) return false;
  if(typeof value !== 'object' && typeof value !== 'function') return true;
  return false;
}
```

nicojs:
This is indeed a problem. Good proposal. Looking forward to a PR 👍 😎

kmdrGroch:
@nicojs should i PR it? Or you will do it on hacktoberfest?

nicojs:
Nope, this one isn't claimed yet. You're picking it up?

If you do + want a free T-shirt, make sure you're signing up for hacktoberfest: https://hacktoberfest.digitalocean.com/

kmdrGroch:
I can pick it up :slightly_smiling_face:

kmdrGroch:
@nicojs this can be closed.

nicojs:
Yes it can. Thanks @kmdrGroch 

If you add `fixes #1133` in the description it will close this issue automatically when the PR is accepted.

kmdrGroch:
thanks :D

PR #1190, opened 2018-10-15
fix: isPrimitive function repair

kmdrGroch:
should solve #1133

### DIFF
--- a/packages/stryker/test/helpers/producers.ts
+++ b/packages/stryker/test/helpers/producers.ts
@@ -19,8 +19,20 @@ export function mock<T>(constructorFn: sinon.StubbableType<T>): Mock<T> {
   return sandbox.createStubInstance(constructorFn);
 }
 
-function isPrimitive(value: any) {
-  return value !== 'object' && !Array.isArray(value);
+/**
+ * @description Checks if variable is a primitive
+ * @param value
+ */
+function isPrimitive(value: any): boolean {
+  const type: string = typeof value;
+  if (
+    ['string', 'undefined', 'symbol', 'boolean'].indexOf(type) > -1
+    || (type === 'number' && !isNaN(value)
+    || value === null)
+    ) {
+    return true;
+  }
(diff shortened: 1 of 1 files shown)


kmdrGroch:
It was really hard @nicojs ;D I didn't expect there could be an issue in your function and that "null" is primitive. But now all should work and be clear :)

kmdrGroch:
Waiting for some easy task to do :)

packages/stryker/test/helpers/producers.ts
```diff
@@ -19,8 +19,20 @@ export function mock<T>(constructorFn: sinon.StubbableType<T>): Mock<T> {
...
 }
 
-function isPrimitive(value: any) {
-  return value !== 'object' && !Array.isArray(value);
+/**
+ * @description Checks if variable is a primitive
+ * @param value
+ */
+function isPrimitive(value: any): boolean {
+  const type: string = typeof value;
```
nicojs: Explicit type here is not needed.

packages/stryker/test/helpers/producers.ts
```diff
@@ -19,8 +19,20 @@ export function mock<T>(constructorFn: sinon.StubbableType<T>): Mock<T> {
...
+  if (
```
nicojs: This implementation can be simplified: 

```ts
return ['string', 'undefined', 'symbol', 'boolean'].indexOf(type) > -1
    || (type === 'number' && !isNaN(value)
    || value === null)
```

nicojs:
> Waiting for some easy task to do :)

I've added some hacktoberfest labels. Feel free to look around 👍

kmdrGroch:
@nicojs and right now? (I have checked these issues, but I don't find anything easy enough at this moment.)

nicojs:
LGTM. Merge when green

kmdrGroch:
@nicojs GREEN!!! :D

PR #1192, opened 2018-10-16
fix(config): Warn uses when no reporters are configured

kmdrGroch:


nicojs:
Great PR! Will make Stryker adoption easier.

Could you update the name of the PR to reflect the change better? We use squash merge, so the name of the PR would preferably be the commit message.  We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#specification)

Something like:

> fix(config): Warn uses when no reporters are configured
"fix(config):

PR #1215, opened 2018-10-26
Enable no implicit this in tsconfig #1210

### LINKED ISSUE #1210: Enable `noImplicitThis` tsconfig option
Enable `noImplicitThis` tsconfig option (in the root tsconfig.json file) and fix any build errors that come along.

DanielSauve:
Ended up having to use ts-ignore on the one issue that came up because it is using 'this' inside an arrow function.

Fixes #1210


nicojs:
I had some time and wanted to see if it was possible. So I went ahead and implemented the fix with an explicit `this`. Will merge when green. Thanks for picking this up!

kmdrGroch:
@nicojs green

nicojs:
Thanks! Merged :)

Issue #1209, opened 2018-10-26
Enable `noFallthroughCasesInSwitch` tsconfig option


nicojs:
Enable `noFallthroughCasesInSwitch` tsconfig option (in the root tsconfig.json file) and fix any build errors that come along.

kmdrGroch:
it was solved in #1217

nicojs:
True! Thanks!

PR #1217, opened 2018-10-27
refactor(tsconfig): Enable `noFallthroughCasesInSwitch` compiler opti…

HVish:
…on (#1209)

### DIFF
--- a/packages/stryker/src/SourceFile.ts
+++ b/packages/stryker/src/SourceFile.ts
@@ -107,7 +107,9 @@ export default class SourceFile {
           if (this.file.textContent.charCodeAt(pos) === CharacterCodes.lineFeed) {
             pos++;
           }
-        // falls through
+          result.push(lineStart);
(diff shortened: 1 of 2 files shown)


packages/stryker/src/SourceFile.ts
```diff
@@ -107,7 +107,9 @@ export default class SourceFile {
           if (this.file.textContent.charCodeAt(pos) === CharacterCodes.lineFeed) {
             pos++;
           }
-        // falls through
+          result.push(lineStart);
```
nicojs: Could we extract these 2 lines into a method? Something like `markLineStart`? It's fine if the function is local to `computeLineStarts`. That way, you don't have to pass along the variables.

HVish: Sure! :+1:

nicojs:
Great job. Thanks!

PR #1235, opened 2018-11-08
fix(JestTestRunner): run jest with --findRelatedTests

### LINKED ISSUE #1226: Running Jest with --findRelatedTests
When running Stryker with Jest, it runs your entire test suite for every mutation. If you have a large test suite this can take quite some time.

Jest has an awesome flag called `--findRelatedTests`, which finds tests related to the file you pass as an arguments. E.g. `jest --findRelatedTests index.js` would only run the tests related to `index.js`. (Often used with [lint-staged](https://github.com/okonet/lint-staged).

If it would be possible to pass the current mutation to the Test Runner, this would save a ton of time!

RobertBroersma:
Pass sandbox fileName to the test runner.

It's probably not the best code, but it works. It speeds up my stryker runs from around 2h45m to around 45m (70% faster!).

It would be great if someone could assist me in writing tests and giving me some pointers for cleaning up/refactoring my code, as the project is quite a lot to take in for me atm.

Cheers, let me know what you think!

PS. As I'm not entirely clear on how to run the project locally on my own code, I had to replace ```importModule(name);``` with ```importModule(__dirname + '../../../' + name);``` in `PluginLoader.ts` in order for stryker to be able to find its plugins. Not sure why.

Fixes #1226


kmdrGroch:
@RobertBroersma in `packages/stryker-jest-runner/src/jestTestAdapters/JestPromiseTestAdapter.ts` in line 22 there is unnecessary comma. You have to remove it in order to tests can pass.
And remember to use `npm test` or `npm run lint-info` before PR so other tests can pass (not blocked by linter :) )

RobertBroersma:
@kmdrGroch You are right! 

I had some trouble running tests and checks locally, but I think I got it now. Will push some fixes soon.

packages/stryker-api/src/test_runner/RunOptions.ts
```diff
@@ -12,6 +12,8 @@ interface RunOptions {
    * It should be loaded right after the test framework but right before any tests can run.
    */
   testHooks?: string;
+
+  fileName?: string;
```
nicojs: Could we change this to `mutatedFileName`? This will further differentiate it from `fileNames` in the `RunnerOptions` interface.

packages/stryker-jest-runner/src/jestTestAdapters/JestPromiseTestAdapter.ts
```diff
@@ -5,12 +5,17 @@ import { Configuration, runCLI, RunResult } from 'jest';
...
     jestConfig.reporters = [];
     const config = JSON.stringify(jestConfig);
     this.log.trace(`Invoking Jest with config ${config}`);
+    if (fileName) {
+      this.log.trace(`Only running tests related to ${fileName}`);
+    }
 
     return runCLI({
+      ...(fileName && { _: [fileName]}),
+      ...(fileName && { findRelatedTests: true }),
```
nicojs: Couldn't these 2 be merged together? 

```ts
return runCLI({
      ...(fileName && { _: [fileName],  findRelatedTests: true }),
```

RobertBroersma:
@nicojs Thanks for the feedback! I've processed your comments and expanded the test for `Sandbox.ts`.

Should we consider adding a configuration option for running with or without `--findRelatedTests`? I don't see any reason why you would run it without the flag, but maybe someone else does.

nicojs:
> Should we consider adding a configuration option for running with or without `--findRelatedTests`? I don't see any reason why you would run it without the flag, but maybe someone else does.

Can you explain to me how the filtering works? If I have a fooTest.js file that tests foo.js, but `foo.js` requires `bar.js` and tests it as well implicitly. What would happen if I run `--findRelatedTests` on `bar.js`?

RobertBroersma:
> Can you explain to me how the filtering works? If I have a fooTest.js file that tests foo.js, but `foo.js` requires `bar.js` and tests it as well implicitly. What would happen if I run `--findRelatedTests` on `bar.js`?
> 
> I have high doubts that it would work in 100% of the cases, because of the dynamic nature of JavaScript.

I've never questioned the working of it tbh, but I've went and looked it up. From a curious stackoverflow user: https://stackoverflow.com/questions/44066996/how-does-jest-findrelatedtests-work-under-the-hood

> When running --findRelatedTests path/to/src-code.js, the first thing that happens is Jest creates an instance of an internal package, jest-resolve-dependencies. It's a pretty straightforward class with two methods: resolve and resolveInverse.
> 
> findRelatedTests calls resolveInverse on the paths you provided, looking up every source and test file that requires your file, in our example path/to/src-code.js. This lookup relies directly on some Jest configuration, specifically roots and/or rootDir, to help resolve paths.
> 
> If the found file is a test file, Jest runs it, simple enough. If the found file is a source file, call it found-file.js, then any test files that import found-file.js and the test files that import any of the source files that themselves import found-file.js will be run.
> 
> It's a clever implementation of, as the maintainers put it, a resolver of "transitive inverse dependencies". You can see for yourself in this while loop.

However there have been reports of it not working optimally, which have been recognized by the Jest team, e.g. https://github.com/facebook/jest/issues/6062

nicojs:
Thanks for figuring this out @RobertBroersma. With all side effects in mind, let's make it configurable. Would you mind making the change?

The default can be `enabled` in my opinion. That way, it is a nice performance boost out-of-the box. Can we call it `jest.enableFindRelatedTests`? Or do you have another proposal?

RobertBroersma:
> Thanks for figuring this out @RobertBroersma. With all side effects in mind, let's make it configurable. Would you mind making the change?
> 
> The default can be `enabled` in my opinion. That way, it is a nice performance boost out-of-the box. Can we call it `jest.enableFindRelatedTests`? Or do you have another proposal?

No problem! I agree making it configurable would be good. I'll make the change. `jest.enableFindRelatedTests` sounds good to me.

RobertBroersma:
@nicojs I've made the flag configurable. Not sure if this is the best way or if and how I should add any tests for it. Let me know what you think!

packages/stryker-jest-runner/src/JestTestRunner.ts
```diff
@@ -15,19 +16,25 @@ export default class JestTestRunner implements TestRunner {
     // Get jest configuration from stryker options and assign it to jestConfig
     this.jestConfig = options.strykerOptions.jest.config;
 
+    // Get enableFindRelatedTests from stryker jest options or default to true
+    this.enableFindRelatedTests = options.strykerOptions.jest.enableFindRelatedTests;
```
nicojs: Maybe we can log on debug what we are doing with this setting?

packages/stryker/src/Sandbox.ts
```diff
@@ -44,8 +44,8 @@ export default class Sandbox {
     return sandbox.initialize().then(() => sandbox);
   }
 
-  public run(timeout: number, testHooks: string | undefined): Promise<RunResult> {
-    return this.testRunner.run({ timeout, testHooks });
+  public run(timeout: number, testHooks: string | undefined, mutant?: TestableMutant): Promise<RunResult> {
+    return this.testRunner.run({ timeout, testHooks, ...(mutant && { mutatedFileName: this.fileMap[mutant.fileName]}) });
```
nicojs: Could we move the locating of the mutated file to the `runMutant` call? It is specific to the case of running when targeting a mutant. I feel like the separation of concerns is intermingled now.

packages/stryker-jest-runner/test/unit/jestTestAdapters/JestPromiseTestAdapterSpec.ts
```diff
@@ -38,7 +38,7 @@ describe('JestPromiseTestAdapter', () => {
   });
 
   it('should call the runCLI method and return the test result', async () => {
-    const result = await jestPromiseTestAdapter.run(jestConfig, projectRoot);
+    const result = await jestPromiseTestAdapter.run(jestConfig, projectRoot, true);
 
     expect(result).to.deep.equal({
       config: {
```
nicojs: I'm missing an assertion for the mutated file here, right?

packages/stryker-jest-runner/src/jestTestAdapters/JestTestAdapter.ts
```diff
@@ -1,5 +1,5 @@
 import { RunResult } from 'jest';
 
 export default interface JestTestAdapter {
-  run(config: object, projectRoot: string): Promise<RunResult>;
+  run(config: object, projectRoot: string, enableFindRelatedTests: boolean, mutatedFileName?: string): Promise<RunResult>;
```
nicojs: Could we leave the name `mutatedFileName` out of this interface. It is not a clean separation of concerns. The JestTestAdapter shouldn't care about mutants, only about running tests. I think these 2 options can be combined into 1: `fileNameUnderTest` or something. If it is provided, it can add `{findRelatedTests: true, _: [fileNameUnderTest]`.

packages/stryker-jest-runner/src/jestTestAdapters/JestPromiseTestAdapter.ts
```diff
@@ -5,12 +5,16 @@ import { Configuration, runCLI, RunResult } from 'jest';
 export default class JestPromiseTestAdapter implements JestTestAdapter {
   private readonly log = getLogger(JestPromiseTestAdapter.name);
 
-  public run(jestConfig: Configuration, projectRoot: string): Promise<RunResult> {
+  public run(jestConfig: Configuration, projectRoot: string, enableFindRelatedTests: boolean, mutatedFileName?: string): Promise<RunResult> {
     jestConfig.reporters = [];
     const config = JSON.stringify(jestConfig);
     this.log.trace(`Invoking Jest with config ${config}`);
+    if (enableFindRelatedTests && mutatedFileName) {
+      this.log.trace(`Only running tests related to ${mutatedFileName}`);
```
nicojs: I like this log message.

RobertBroersma:
I'll be working on your feedback hopefully soon. Got a bit busy at work the past days!

packages/stryker-jest-runner/src/jestTestAdapters/JestTestAdapter.ts
RobertBroersma: I'm not sure I understand. Do you mean to only pass the `fileNameUnderTest` when running related to a specific file? I.e. when the flag is true and we're testing a mutant?

nicojs:
@RobertBroersma just let us know if you need some more help.

RobertBroersma:
@nicojs I'm hoping to find some time this evening. I'll let you know if I run into any issues.

packages/stryker-jest-runner/src/JestTestRunner.ts
```diff
@@ -15,19 +16,31 @@ export default class JestTestRunner implements TestRunner {
...
+
     // basePath will be used in future releases of Stryker as a way to define the project root
     // Default to process.cwd when basePath is not set for now, should be removed when issue is solved
     // https://github.com/stryker-mutator/stryker/issues/650
     this.jestConfig.rootDir = options.strykerOptions.basePath || process.cwd();
     this.log.debug(`Project root is ${this.jestConfig.rootDir}`);
   }
 
-  public async run(): Promise<RunResult> {
+  public async run(options?: RunOptions): Promise<RunResult> {
```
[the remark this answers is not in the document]
RobertBroersma: I think I initially did that because tests started failing and I was unsure whether or not running without `RunOptions` was a case. I've taken a look at the other `*TestRunnerSpec`s and I've added some `RunOptions` to the Jest ones. Is this correct?

nicojs:
Thanks for the quick response! I first want to merge #1260. In there we have an integration test that tests the score of an actual stryker run with the jest runner. After that, I'll merge master in this branch and see if it works locally as expected (also will keep an eye on performance to see it improved). 

After that, I'll merge. Should still be today.

Awesome job 👍

nicojs:
Worked great, although no performance improvement for the integration test (takes 1 minute and 1 second for both cases locally on my machine).

Will merge when green :+1:

PR #1236, opened 2018-11-09
refactor(mocha-timeout): Configure mocha timeout in package.json

### LINKED ISSUE #1157: Configure mocha timeout in package.json
| Operating System | win10

So i managed to find that issue which is like always.

1) A babel project with the `only` option enabled
       should be able to transpile the input files:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\Users\user\Desktop\stryker\packages\stryker-babel-transpiler\test\integration\BabelProjects.it.js)

It happens all the time in first test. After second one, it passes *WTF*

Honestly i dunno how to "recreate" that, since as i told, it happens only once, and then it works (i think after restarting PC it should happen again(?) [maybe some caching or coz I test it in VScode or not enough RAM(? 8GB)]

I found second one <3

1) CoverageInstrumenterTranspiler
       when coverage analysis is "all"
         should instrument code of mutated files:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\Users\user\Desktop\stryker\packages\stryker\test\unit\transpiler\CoverageInstrumenterTranspilerSpec.js)

bharaninb:
Closes #1157

### DIFF
--- a/packages/stryker-api/package.json
+++ b/packages/stryker-api/package.json
@@ -8,7 +8,8 @@
     "prebuild": "npm run clean",
     "build": "tsc -p .",
     "postbuild": "tslint -p tsconfig.json",
-    "test": "nyc --reporter=html --report-dir=reports/coverage --check-coverage --lines 90 --functions 68 --branches 63 mocha \"test/**/*.js\""
+    "test": "nyc --reporter=html --report-dir=reports/coverage --check-coverage --lines 90 --functions 68 --branches 63 npm run mocha",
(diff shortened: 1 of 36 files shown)


bharaninb:
- Refactored test script with existing timeouts.
- Updated timeout for `stryker-webpack-transpiler` since the test fails sometimes.

Let me know, if we need to change timeout for other packages

simondel:
Thanks!

PR #1237, opened 2018-11-12
feat(mocha-runner): improve Mocha support

### LINKED ISSUE #1046: Improve Mocha support
### Summary

Hello all, I was hoping somebody could let me know what I'm doing wrong here. I keep getting a syntax error on CSS files. We're using Stryker version 0.14.1. We are also using react/redux with the mocha test framework. We are not using webpack. I'm unsure if it is some configuration error, version issue, or something completely different. The error is as follows:

[2018-07-26 13:02:05.874] [ERROR] InitialTestExecutor - One or more tests resulted in an error:
        SyntaxError: Unexpected token .
C:\code_base\poi\node_modules\form\lib\util\block.css:1
(function (exports, require, module, __filename, __dirname) { .a-form-v3-10-0-block {

### Stryker config

```js
module.exports = function(config){
  config.set({
   files: [
        "src/**/*.js",
        "test/**/*.js"
    ],
    mutate:[
      "src/pages/poi/*.js"
    ],
    testRunner: "mocha",
    mutator: "javascript",
    transpilers: ["babel"],
    reporter: ["html", "clear-text", "progress"],
    testFramework: "mocha",
    coverageAnalysis: "off",
    babelrcFile: ".babelrc"
  });
}
```

### Stryker environment

```
+-- stryker@x0.14.1
+-- stryker-api@x0.11.0
+--stryker-babel-transpiler@0.2.2
+--stryker-html-reporter@0.11.1
+--stryker-javascript-mutator@0.2.3
+--stryker-mocha-framework@0.7.0
+==stryker-mocha-runner@0.10.0
```

### Your Environment
| software                | version(s)
| -------------------- | -------  
| node                      | 6.1.0
| npm                       | 3.8.6
| Operating System | Windows 7 Enterprise

bharaninb:
closes #1046

### DIFF
--- a/packages/stryker-mocha-framework/src/MochaTestFramework.ts
+++ b/packages/stryker-mocha-framework/src/MochaTestFramework.ts
@@ -17,6 +17,7 @@ export default class MochaTestFramework implements TestFramework {
   public filter(testSelections: TestSelection[]) {
     const selectedTestNames = testSelections.map(selection => selection.name);
     return `var Mocha = window.Mocha || require('mocha');
+      var describe = Mocha.describe;
       var selectedTestNames = ${JSON.stringify(selectedTestNames)};
       if (window.____mochaAddTest) {
         Mocha.Suite.prototype.addTest = window.____mochaAddTest;
--- a/packages/stryker-mocha-runner/src/MochaOptionsLoader.ts
+++ b/packages/stryker-mocha-runner/src/MochaOptionsLoader.ts
@@ -13,7 +13,7 @@ export default class MochaOptionsLoader {
     const mochaOptions = Object.assign({}, config[mochaOptionsKey]) as MochaRunnerOptions;
     let optsFileName = path.resolve(this.DEFAULT_MOCHA_OPTS);
 
-    if (mochaOptions && mochaOptions.opts) {
+    if (mochaOptions.opts) {
       optsFileName = path.resolve(mochaOptions.opts);
     }
 
@@ -21,10 +21,8 @@ export default class MochaOptionsLoader {
       this.log.info(`Loading mochaOpts from "${optsFileName}"`);
       const options = fs.readFileSync(optsFileName, 'utf8');
       return Object.assign(this.parseOptsFile(options), mochaOptions);
-    }
-
-    if (mochaOptions && mochaOptions.opts) {
-      throw new Error(`Could not load opts from "${optsFileName}". Please make sure opts file exists.`);
+    } else if (mochaOptions.opts) {
+      this.log.error(`Could not load opts from "${optsFileName}". Please make sure opts file exists.`);
     } else {
       this.log.debug('No mocha opts file found, not loading additional mocha options (%s.opts was not defined).', mochaOptionsKey);
     }
--- a/packages/stryker-mocha-runner/src/MochaOptionsLoader.ts
+++ b/packages/stryker-mocha-runner/src/MochaOptionsLoader.ts
@@ -7,18 +7,28 @@ import MochaRunnerOptions, { mochaOptionsKey } from './MochaRunnerOptions';
 export default class MochaOptionsLoader {
 
   private readonly log = getLogger(MochaOptionsLoader.name);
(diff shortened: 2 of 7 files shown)


bharaninb:
* [x]use your mocha.opts file by default (if exists)
* [ ]  Make sure relative imports when using `--require` (same as mocha does it)
* [ ]  Make sure `const { describe } = require('mocha');` works as well.

packages/stryker-mocha-framework/src/MochaTestFramework.ts
```diff
@@ -17,6 +17,7 @@ export default class MochaTestFramework implements TestFramework {
   public filter(testSelections: TestSelection[]) {
     const selectedTestNames = testSelections.map(selection => selection.name);
     return `var Mocha = window.Mocha || require('mocha');
+      var describe = Mocha.describe;
```
simondel: Why do we need to save this?

packages/stryker-mocha-runner/src/MochaOptionsLoader.ts
```diff
@@ -7,18 +7,28 @@ import MochaRunnerOptions, { mochaOptionsKey } from './MochaRunnerOptions';
 export default class MochaOptionsLoader {
 
   private readonly log = getLogger(MochaOptionsLoader.name);
+  private readonly DEFAULT_MOCHA_OPTS = 'test/mocha.opts';
 
   public load(config: StrykerOptions): MochaRunnerOptions {
     const mochaOptions = Object.assign({}, config[mochaOptionsKey]) as MochaRunnerOptions;
+    let optsFileName = path.resolve(this.DEFAULT_MOCHA_OPTS);
+
     if (mochaOptions && mochaOptions.opts) {
```
simondel: I notice that we always check if we have `mochaOptions`, but there should always be an object, right? We define it on line 13.

packages/stryker-mocha-runner/src/MochaOptionsLoader.ts
```diff
@@ -7,18 +7,28 @@ import MochaRunnerOptions, { mochaOptionsKey } from './MochaRunnerOptions';
...
+    }
+
+    if (fs.existsSync(optsFileName)) {
       this.log.info(`Loading mochaOpts from "${optsFileName}"`);
       const options = fs.readFileSync(optsFileName, 'utf8');
       return Object.assign(this.parseOptsFile(options), mochaOptions);
+    }
+
+    if (mochaOptions && mochaOptions.opts) {
+      throw new Error(`Could not load opts from "${optsFileName}". Please make sure opts file exists.`);
```
simondel: This looks really strange to me. Because we return on line 23, we throw an error here.

If we get an error while reading or parsing the `Opts` file, we will not log this error because we have no try-catch. 

Could you change this to:
```
if (fs.existsSync(optsFileName)) {
   ...
} else if (mochaOptions.opts) {
   this.log.error(`Could not load opts from "${optsFileName}". Please make sure opts file exists.`);
}
```
(The stryker-karma-runner handles an error like this by logging it instead of throwing an error)

packages/stryker-mocha-runner/src/MochaTestRunner.ts
```diff
@@ -116,7 +116,11 @@ export default class MochaTestRunner implements TestRunner {
 
   private additionalRequires() {
     if (this.mochaRunnerOptions.require) {
-      this.mochaRunnerOptions.require.forEach(LibWrapper.require);
+      const stuffToRequire = this.mochaRunnerOptions.require
```
simondel: Could you rename this to `modulesToRequire`?

packages/stryker-mocha-runner/src/MochaTestRunner.ts
```diff
@@ -116,7 +116,11 @@ export default class MochaTestRunner implements TestRunner {
...
+        .map(theModule => {
```
simondel: Could you rename this to `module`?

simondel:
Looks great! Thanks again!

Issue #809, opened 2018-05-29
Align test runner config options


nicojs:
Right now, each test runner has his own setup when it comes to custom configuration.

* The stryker-jest-runner uses a `jest` key as a container for its config.
* The stryker-karma-runner uses the `karmaConfig` and `karmaConfigFile` keys.
* The stryker-mocha-runner uses the `mochaOptions` config key
* The stryker-jasmine-runner uses the `jasmineConfigFile` key (has no other way of configuring as of yet).

It's kind of a mess. I feel like we should align on a standard way of providing test runner options.

I've been toying with the following idea:

```js
{
  testRunner: {
    name: 'karma',
    settings: { 
        /* custom settings go here */
        configFile: 'karma.conf.js',
        config: {
           browsers: [ 'ChromeHeadless' ]
        }
    }
  }
}
```

This would make the way of providing settings unified across test runners. It will also make it really clear that the settings are in fact to configure the test runner, keeping it nicely together.

@stryker-mutator/contributors what do you think?

mthmulders:
I like the idea to have all runners that we maintain work in a more-or-less similar way. Makes it easy for us, makes it easy for Stryker users.

Having said that, my personal preference would be to clearly separate the *Stryker* configuration from that of the *test framework* (Jest, Karma, ...). Sure, they can live in the same file, but the naming should make it clear that whatever you provide there applies to the *test framework* that you use. I think that your naming suggestion addresses that.

Also, we might take the opportunity to standardise on prevalence. In your example, imagine `karma.conf.js` also has a setting `browsers`. What will the final configuration value be? I wouldn't be surprised if the different Stryker runners currently have different rules for that.

nicojs:
> Having said that, my personal preference would be to clearly separate the Stryker configuration from that of the *test framework* (Jest, Karma, ...). 

The problem is that separating it might make it confusing. For example, the `stryker-jasmine-runner` would use the `jasmine` key in your example. But this is to configure the *jasmine test runner*, not the *jasmine test framework*. You even made the mistake on your comment. Moving it to the `testRunner` settings would eliminate this problem.

My idea would be to inject the settings under `testRunner.settings` directly into the test runner, instead of the entire stryker options. Simular approaches could also be taken for `testFramework` and others if we need it.

[...]

PR #1239, opened 2018-11-14
feat(karma-runner): Support automatic port selection

### LINKED ISSUE #740: Support automatic port selection in stryker-karma-runner
At this moment whenever karma (automatically) selects the next open port, stryker will report it as an error. We should support this feature.

When we support this feature, we can also remove the `port` config in stryker-api, making the api much cleaner (remove test runner specific configuration)

bharaninb:
closes #740

### DIFF
--- a/packages/stryker-karma-runner/test/integration/KarmaTestRunner.it.ts
+++ b/packages/stryker-karma-runner/test/integration/KarmaTestRunner.it.ts
@@ -241,7 +241,7 @@ describe('KarmaTestRunner', () => {
     }));
   });
 
-  describe.only('when specified port is not available', () => {
+  describe('when specified port is not available', () => {
 
     let dummyServer: DummyServer;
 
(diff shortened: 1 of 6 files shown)


bharaninb:
https://github.com/stryker-mutator/stryker/issues/740#issue-322192634

Removing the `port` config in stryker-api can be done in feature #809

packages/stryker-karma-runner/test/integration/KarmaTestRunner.it.ts
```diff
@@ -238,4 +238,33 @@ describe('KarmaTestRunner', () => {
       return true;
     }));
   });
+
+  describe('when specified port is not available', () => {
+
+    before(() => {
+      const testRunnerOptions: RunnerOptions = {
+        fileNames: [],
+        port: 9882,
```
nicojs: I can see this test actually works, because it picks the same port as previous integration test. However, i would like to keep the integration tests independent. Could open a server ourselves and verify that a different port is picked?

bharaninb: Yeah sure

nicojs:
Thanks! I've tested your implementation on windows, but it gave some problems. Namely, the port binding is done on host "0.0.0.0", which is different from the `http.Server.listen` default of "localhost". I've also introduced some promises, awaits and an assertion to make really sure the port is actually a different one. 

Will merge when green 💚

PR #1250, opened 2018-11-22
fix(mutator): String literal mutator

### LINKED ISSUE #465: StringLiteral mutator should not empty strings that are exported directly.
Examples that shouldn't be modified:

```
export { foo } from "foo";

export const foo = "bar";
export const bar = `baz`;
```

Examples that should be modified:

```
export const baz = `${foo}{$bar}`;
export const quux = () => "foo";
```

This last example is still implementing some small piece of functionality. The static strings however can't usefully be **unit tested** without simply duplicating the string itself into the unit test.

I'll probably pick this ticket up, but I'm leaving a note for myself here.

bharaninb:
Fixes #465


nicojs:
Another spectacular PR @bharaninb, thanks a lot 🥇

PR #1251, opened 2018-11-23
feat(console-colors): Add a global config option to enable/disable colors in console

### LINKED ISSUE #1218: Add a global config option to enable/disable colors in console
We should allow people to disable console colors with one setting. Right now, both the clear text reporter and the log4js logger log with colors. Build servers generally don't support colors. 

We should add a global config option `allowColors` or `allowConsoleColors`. If that is false, we should not use colors.

I think we should also disable the clear text specific setting for this, or i would like to hear a good use case where colors are desired in the clear text reporter, but not in the log4js output.

bharaninb:
closes #1218

### DIFF
--- a/packages/stryker-api/src/config/Config.ts
+++ b/packages/stryker-api/src/config/Config.ts
@@ -27,6 +27,7 @@ export default class Config implements StrykerOptions {
     high: 80,
     low: 60
   };
+  public allowConsoleColors: boolean = true;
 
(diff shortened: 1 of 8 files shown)


nicojs:
Awesome. ❤️ this PR! I've changed one thing: allow for the value to be passed through the command line. As this argument has the most value on build servers, you probably want the value to be passed via command line (so you can share local configuration with build server configuration as much as possible).

PR #1264, opened 2018-11-30
WIP: zero config runs

### LINKED ISSUE #1051: Zero config test runs
With the command test runner (#1047) and the new sane default for the `files` array (#665), I think we have everything in place to support zero configuration test runs. I would suggest the following defaults to be added:

* Test runner: `'command'`
* Mutator: ~~The installed mutator (javascript/typescript)~~ `"javascript"`
* Files to mutator: `['src/**/*.js', 'lib/**/*.js']`
* Coverage analysis: `'off'` (it is `'perTest'` now, but the CommandTestRunner does not support it)

With this issue we can also improve our documentation of the main stryker package, to use the [jest](https://jestjs.io/docs/en/configuration.html) style configuration.

We should document the zero config somewhere. I think the quickstart on the website would be a good place.

We should maybe log the stryker configuration that is used for the run _somewhere_

@stryker-mutator/contributors do you agree?

nicojs:
Fixes #1051


nicojs:
@stryker-mutator/contributors what do you guys think about leaving the mutator to `"es5"` for now? Update it when we release Stryker 1.0? To reduce the amount of breaking changes introduced by this PR?

mthmulders:
I would opt for keeping the number of breaking changes in this feature as low as possible.

nicojs:
This PR is ready for review. @mthmulders @simondel @Archcry, welcome to take this.

kmdrGroch:
@nicojs I understand this is `zero tests` for JS only? (no TS support I see)

simondel:
Awesome :)

On Dec 19, 2018 6:48 PM, "Bartosz Leoniak" <notifications@github.com> wrote:

@nicojs <https://github.com/nicojs> I understand this is zero tests for JS
only? (no TS support I see)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/stryker-mutator/stryker/pull/1264#issuecomment-448684962>,
or mute the thread

packages/stryker/src/input/InputFileResolver.ts
```diff
@@ -50,34 +53,77 @@ export default class InputFileResolver {
...
+        files.forEach(fileName => fileSet.add(fileName));
+      }
+    }
+    return Array.from(fileSet);
+  }
+
+  private async expandPattern(globbingExpression: string, logAboutUselessPatterns: boolean): Promise<string[]> {
+    const fileNames = (await glob(globbingExpression)).map(relativeFile => path.resolve(relativeFile));
+    if (!fileNames.length && logAboutUselessPatterns) {
+      this.log.warn(
```
mthmulders: Nice feature :-)

packages/stryker/src/input/InputFileResolver.ts
```diff
@@ -50,34 +53,77 @@ export default class InputFileResolver {
...
-        throw new StrykerError(
-          `Cannot determine input files. Either specify a \`files\` array in your stryker configuration, or make sure "${process.cwd()}" is located inside a git repository`,
-          error
-        );
-      });
+      );
+      const fileNames = stdout.toString()
+        .split('\n')
+        .map(line => line.trim())
+        .filter(line => line)
```
mthmulders: Can you explain why this is necessary? It looks superfluous, but knowing you it probably isn't.

packages/stryker/src/input/InputFileResolver.ts
```diff
@@ -50,34 +53,77 @@ export default class InputFileResolver {
...
-    return childProcessAsPromised
-      .exec(
+  private resolveMutateFiles() {
+    return this.expand(this.mutatePatterns, !shallowEquals(this.mutatePatterns, new Config().mutate));
+
+    function shallowEquals<T>(arr1: ReadonlyArray<T>, arr2: ReadonlyArray<T>): boolean {
+      if (arr1.length !== arr2.length) {
+        return false;
+      } else {
+        for (let i = 0; i < arr1.length; i++) {
```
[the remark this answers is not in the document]
nicojs: I've changed it up, however still have a variable. I don't like the fact that I can't enforce a recursive function to be tail recursive in JavaScript:

```ts
const iter1 = arr1.values();
        const iter2 = arr2.values();
        let item1: IteratorResult<T>;
        while (!(item1 = iter1.next()).done) {
          if (item1.value !== iter2.next().value) {
            return false;
          }
        }
        return true;
```

packages/stryker/src/input/InputFileResolver.ts
nicojs: Yeah, i was testing the new zero config out and didn't like a lot of "warnings" that weren't really warnings. However, I _do_ like the warning when a user configured it.

packages/stryker/src/input/InputFileCollection.ts
```diff
@@ -14,13 +14,14 @@ export default class InputFileCollection {
   public logFiles(log: Logger) {
     if (!this.files.length) {
       log.warn(normalizeWhiteSpaces(`
-        No files selected. Please make sure you either run stryker a git repository context (and don't specify \`files\` in your stryker.conf.js file),
-        or specify the \`files\` property in your stryker config.`));
+        No files selected. Please make sure you either (1) run Stryker a git repository context
```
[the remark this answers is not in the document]
nicojs: True, that's better.

packages/stryker/src/input/InputFileResolver.ts
nicojs: This better?

```ts
        .filter(line => line) // remove empty lines
```

It filters out empty lines, as empty string is a falsy value.

By the way, I'm touched to hear that you think I'm not a suspicious person 🎅

packages/stryker/src/input/InputFileResolver.ts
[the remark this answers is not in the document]
nicojs: EDIT: Even better:

```ts
function shallowEquals<T>(arr1: ReadonlyArray<T>, arr2: ReadonlyArray<T>): boolean {
      if (arr1.length !== arr2.length) {
        return false;
      } else {
        const iter = arr2.values();
        for (const value of arr1) {
          if (value !== iter.next().value) {
            return false;
          }
        }
        return true;
      }
    }
```

nicojs:
Thanks for your review @mthmulders! I've implemented your review comments. Did you have a look at everything? If you're happy, I think we can merge it.

packages/stryker/src/input/InputFileResolver.ts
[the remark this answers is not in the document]
mthmulders: Nice find!

packages/stryker/src/input/InputFileResolver.ts
mthmulders: Yes, absolutely better IMHO.

nicojs:
Haha, sorry @mthmulders. We might need to need to revert back to my beautiful `for` statement. 😂 

```
InputFileResolver
       when a globbing expression does not result in a result
         should not log a warning if the globbing expression was the default logging expression:
     TypeError: arr2.values is not a function
```

Node 6 doesn't support this magic. We're dropping node 6 support April 30th, so be patient.

PR #1277, opened 2018-12-13
stryker-mocha-runner : Add support for --grep option

xurei:
Add support for the --grep option.

**Use case** 
I'm testing a REST service with a database. Testing the write operations makes the tests fail, which leads to false negative with mutation testing.
Example : 
- Test1: GET /something should only contain N elements.
- Test2: POST /something should be accepted.
If Test1 is runned again, it will fail, even with the original code (it contains N+1 elements now).
The alternative would be to reset the database before every test, but it is time consuming and will make mutation testing much longer.

The --grep option in mocha allows to test read-only operations exclusively, so I'm sure that there is no failed test due to a side effects.

### DIFF
--- a/packages/stryker-mocha-runner/README.md
+++ b/packages/stryker-mocha-runner/README.md
@@ -42,7 +42,8 @@ module.exports = function (config) {
             ui: 'bdd',
             timeout: 3000,
             require: [ /*'babel-register' */],
-            asyncOnly: false
+            asyncOnly: false,
+            grep: /.*/
         }
     });
 }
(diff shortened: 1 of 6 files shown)


nicojs:
Hi @xurei , thanks for this PR. Did you test it on your own use case with [install-local](https://github.com/nicojs/node-install-local) to make sure it has the desired effects?

Not sure why the build failed. Will try to merge master to see if that fixes it.

kmdrGroch:
@nicojs there is an error (at least travis says so):
```
ERROR: /home/travis/build/stryker-mutator/stryker/packages/stryker-api/src/core/Factory.ts:14:50 - Type literal has only a call signature — use `new (settings: TSettings) => T` instead.
ERROR: /home/travis/build/stryker-mutator/stryker/packages/stryker-api/src/core/Factory.ts:31:48 - Type literal has only a call signature — use `new (settings: TSettings) => T` instead.
```

nicojs:
Yeah, appearently a tslint update or something. Will fix tomorrow (or else PR's welcome)

xurei:
@nicojs Yes tested it and working. I'm using is actively on a project.
I believe the build failure is from the master branch. However, if it comes from my code, I'll be glad to fix that :-)

nicojs:
Awesome! I fixed the build on master and merged it back into this branch. 

Could you document this feature in the readme of stryker-mocha-runner? Will merge it after that. Thanks!

nicojs:
I've added the documentation. Will merge when green. Thanks!

PR #1281, opened 2018-12-18
refactor(tslint): Fix tslint warnings

nicojs:


nicojs:
@kmdrGroch this should fix the build

Issue #1294, opened 2018-12-24
ERROR StrykerCli an error occurred TypeError: Cannot read property 'length' of undefined


kmdrGroch:
New version of Stryker throws an error (previous wasn't)

ERROR StrykerCli an error occurred TypeError: Cannot read property 'length' of undefined
at shallowEquals (C:\Users\19gro\Desktop\BigMath\node_modules\stryker\src\input\InputFileResolver.js:63:38)
    at InputFileResolver.resolveMutateFiles (C:\Users\19gro\Desktop\BigMath\node_modules\stryker\src\input\InputFileResolver.js:61:50)
    at InputFileResolver.<anonymous> (C:\Users\19gro\Desktop\BigMath\node_modules\stryker\src\input\InputFileResolver.js:37:34)
    at step (C:\Users\19gro\Desktop\BigMath\node_modules\tslib\tslib.js:133:27)
    at Object.next (C:\Users\19gro\Desktop\BigMath\node_modules\tslib\tslib.js:114:57)
    at C:\Users\19gro\Desktop\BigMath\node_modules\tslib\tslib.js:107:75
    at new Promise (<anonymous>)
    at Object.__awaiter (C:\Users\19gro\Desktop\BigMath\node_modules\tslib\tslib.js:103:16)
    at InputFileResolver.resolve (C:\Users\19gro\Desktop\BigMath\node_modules\stryker\src\input\InputFileResolver.js:31:24)
    at Stryker.<anonymous> (C:\Users\19gro\Desktop\BigMath\node_modules\stryker\src\Stryker.js:51:132)

Using typescript config
```js
module.exports = function(config) {
  config.set({
    mutator: "typescript",
    packageManager: "npm",
    reporters: ["html", "clear-text", "progress"],
    testRunner: "mocha",
    transpilers: ["typescript"],
    testFramework: "mocha",
    coverageAnalysis: "off",
    tsconfigFile: "tsconfig.json",
    mutate: ["src/**/*.ts"]
  });
};
```

kmdrGroch:
It was my fault, since dependabot didn't update all packages and so there was an issue. It works now though. Sorry

PR #1295, opened 2018-12-24
fix:  cannot read property 'length' of undefined

kmdrGroch:
fixes: #1294


kmdrGroch:
somehow "undefined" is passed to that function and so it throws error. This should fix it, but still this is strange :/

PR #1300, opened 2018-12-24
maintain: Rename 'integrationTest' folder to 'e2e'

### LINKED ISSUE #1298: Rename "integrationTest" to "e2e"
Right now, we have 2 kinds of integration tests:

1. **Integration tests for a specific package.**
  These tests are located inside a specific package 'test' directory
2. **Overall integration tests.**
  These tests are located inside the "integrationTests" directory.

We should rename one of them, because this is quit confusing. I think renaming "integrationTest" to "e2e" will be the best solution, as that is what they are: we're testing Stryker end to end with different plugin configurations.

kmdrGroch:
Fixes #1298


kmdrGroch:
@nicojsI hope that's what you meant. Also, I have removed some unused script

nicojs:
Thanks 👍

PR #1306, opened 2018-12-27
style: improve readability

kmdrGroch:
I have added:
- trailing comma in objects
- order in imports
- order in objects
- order in classes / namespaces etc
- newline before return (can significantly improve readability in some cases

- other small changes

// Since then we will be using `tslint:all` config so everything that is implemented is `enabled` and we can disable sth if we don't like it. Now we can try enabling new rules - my suggestions:
- "deprecation" - there are 6 of them I guess
- "no-duplicate-imports" - 6-7? No idea how to fix it though
- "no-empty" - a bunch of them
- "no-empty-interface" - bunch of them
- "no-implicit-dependencies" - didn't check
- "no-non-null-assertion" (we have two of them, honestly I don't even know what it is and how to deal with it)
- "no-unnecessary-class" - 1 i guess
- "no-unnecessary-qualifier" - also 1
- "prefer-template"
- "strict-type-predicates" <- this can find some issues (there are some type issues in stryker) We can talk about it in the separate issue too


kmdrGroch:
Also I'm so sorry @nicojs that this PR is so big... I didn't know that changing order will take that much :c
But honestly if it passes, and you like how it looks, you can freely merge it, since there are only style changes, there should be no breaking change or anything

nicojs:
@kmdrGroch this PR is far too big. It's virtually impossible for me to review.

Could you open an issue explaining which rules you want to enable and why? 

As a rule, I don't want to enforce alphabetic order of properties. I know we have it for objects, we should rethink this. For example: 

```ts
interface TestFramework {
  afterEach(): string;
  beforeEach(): string;
}
```

This doesn't make sense to me. 
We should also make it easy for developers to comply. I'll create a PR for this. My idea is to run the tslint plugin when a file is saved.

PR #1309, opened 2018-12-27
fix: remove unused property - port

### LINKED ISSUE #1304: Remove "port" property from Stryker configuration
The "port" property in Stryker options is a remnant of the past. It is specific to the Karma test runner. Since then Karma supports automatic free port selection. This is also supported by Stryker since #1239 . I think it is time to remove the setting from the config options and `TestRunnerSettings`.

kmdrGroch:
Fixes #1304


packages/stryker/README.md
```diff
@@ -254,14 +254,7 @@ you can consult [npm](https://www.npmjs.com/search?q=%40stryker-plugin) or
...
 
-Default: `9234`  
-Command line: `--port 9234`  
-Config file: `port: 9234`    
- 
-  
-With `port` you specify the first port to pass on to the test runner to use. Any additional test runners will be spawned using ports n+1, n+2, etc.
-For example, when you set to use port 9234 and Stryker decides to start four test runner processes, ports 9234, 9235, 9236 and 9237 will be passed to the test runner.  
-If the test runner decides to use the port it should be available for use.
+Deprecated
```
nicojs: We should explain a little bit what happened and how to configure the port in karma options. How about:

> Removed. With this setting you could configure the `port` used by the `karma` test runner. It can now be configured directly with `karma: { config: { port: 7654 } }` (but shouldn't be needed as karma supports its own port selection)

packages/stryker-karma-runner/test/unit/KarmaTestRunnerSpec.ts
```diff
@@ -176,11 +170,6 @@ describe('KarmaTestRunner', () => {
       karmaRunStub.callsArgOn(1, 0);
     });
 
-    it('should execute "karma run"', async () => {
-      await sut.run({});
-      expect(karmaRunStub).calledWith({ port: settings.port });
```
nicojs: This test is still valid, just no longer with the port configuration

packages/stryker-karma-runner/test/unit/KarmaTestRunnerSpec.ts
```diff
@@ -206,12 +195,6 @@ describe('KarmaTestRunner', () => {
       expect(TestHooksMiddleware.instance.currentTestHooks).eq('foobar');
     });
 
-    it('should update port when karma chooses a different port', async () => {
-      reporterMock.emit('server_start', 1984);
-      await sut.run({});
-      expect(karmaRunStub).calledWith({ port: 1984 });
```
nicojs: This is also still a valid test. The stryker-karma-runner remains interested in the port that karma chooses to claim.

packages/stryker-karma-runner/test/unit/starters/stryker-karma.confSpec.ts
```diff
@@ -87,12 +87,6 @@ describe('stryker-karma.conf.js', () => {
     });
   });
 
-  it('should set the port', () => {
-    sut.setGlobals({ port: 1337 });
-    sut(config);
-    expect(config).include({ port: 1337 });
-  });
```
nicojs: True, this is no longer needed 👍

packages/stryker/src/Sandbox.ts
```diff
@@ -115,10 +115,9 @@ export default class Sandbox {
   private initializeTestRunner(): Promise<void> {
     const settings: RunnerOptions = {
       fileNames: Object.keys(this.fileMap).map(sourceFileName => this.fileMap[sourceFileName]),
-      port: this.options.port + this.index,
       strykerOptions: this.options,
     };
-    this.log.debug(`Creating test runner %s using settings {port: %s}`, this.index, settings.port);
+    this.log.debug(`Creating test runner %s using settings {port: %s}`, this.index);
```
nicojs: This string template needs to be updated as well.

packages/stryker-karma-runner/test/unit/KarmaTestRunnerSpec.ts
kmdrGroch: it says "Property 'port' does not exist on type 'RunnerOptions'" in `settings.port`

kmdrGroch: + I think it is the same as that one from below

kmdrGroch:
@nicojs we've got a problem >.<

nicojs:
I found the issue. You removed too much. The stryker-karma-runner is still in charge of figuring out the exact port karma is running on. I've restored that part of the code. Should be green now 🤞

kmdrGroch:
@nicojs it's green! <3

PR #1310, opened 2018-12-27
tslint: use tslint:all template to be aware of all options

kmdrGroch:
Here is what I think it could be done.
By comments I show what I guess would improve quality, readable and consistency of code


simondel:
We've decided not to merge this for now.

PR #1314, opened 2019-01-09
feat(html-reporter): Remove side effects from html reporter

nicojs:
Note: #1313 needs to be merged to master first


packages/stryker-html-reporter/package.json
```diff
@@ -41,21 +41,19 @@
     "mkdirp": "~0.5.1",
     "mz": "~2.7.0",
     "rimraf": "~2.6.1",
+    "stryker-api": "~0.23.0",
```
simondel: Why pin it on exactly `0.23.0`? (`~` pins on the minor version which is the last `0` here.

packages/stryker-html-reporter/src/HtmlReporter.ts
```diff
@@ -1,26 +1,28 @@
...
 export default class HtmlReporter implements Reporter {
-  private readonly log = getLogger(HtmlReporter.name);
-  private _baseDir: string;
-  private mainPromise: Promise<void>;
-  private mutantResults: MutantResult[];
-  private files: SourceFile[];
-  private scoreResult: ScoreResult;
-
-  constructor(private readonly options: Config) {
+  private _baseDir!: string;
```
simondel: What does this `!` do?

packages/stryker-html-reporter/package.json
nicojs: Because it now is a _dependency_ instead of a peerDependency. It will automatically be updated by lerna (at least, as far as I understand it).

packages/stryker-html-reporter/src/HtmlReporter.ts
nicojs: It's the [definite assignment operator](https://blogs.msdn.microsoft.com/typescript/2018/01/31/announcing-typescript-2-7/#definite-assignment-assertions). This allows us to use `--strictPropertyInitialization`, while not assigning them in the constructor.

simondel:
@nicojs Feel free to merge

PR #1317, opened 2019-01-11
feat(config-editors): Remove side effects from all config editor plugins

nicojs:
_Note: #1313 Needs to be merged first_

Removes side effects from registering all `ConfigEditor` plugins:

- [x] `JestConfigEditor`
- [x] `MochaRunnerConfigEditor`
- [x] `TypeScriptConfigEditor`


packages/stryker/src/Stryker.ts
```diff
@@ -51,16 +52,19 @@ export default class Stryker {
...
-    this.freezeConfig();
-    this.injector = rootInjector
+    const configEditorInjector = rootInjector
       .provideValue(commonTokens.getLogger, getLogger)
       .provideFactory(commonTokens.logger, loggerFactory, Scope.Transient)
-      .provideValue(commonTokens.pluginResolver, pluginLoader as PluginResolver)
+      .provideValue(commonTokens.pluginResolver, pluginLoader as PluginResolver);
+    configEditorInjector
+      .provideFactory(coreTokens.pluginCreatorConfigEditor, PluginCreator.createFactory(PluginKind.ConfigEditor))
+      .injectClass(ConfigEditorApplier).edit(this.config);
```
simondel: The way the config editor is used here is quite sneaky. I would prefer if this is done in a separate method so we can name this operation like we had previously.

nicojs: I agree. I've changed it in the transpiler branch.

See https://github.com/stryker-mutator/stryker/blob/4caf63d214046cd6396521c4dee54361fb6a40dc/packages/stryker/src/Stryker.ts#L45

and https://github.com/stryker-mutator/stryker/blob/4caf63d214046cd6396521c4dee54361fb6a40dc/packages/stryker/src/config/configFactory.ts#L13-L17

If we merge it here, I'll fix it with that one. I'm planning on working on Stryker full time from tomorrow for about 2 weeks, so if you agree I'll merge them in quick succession.

PR #1328, opened 2019-01-22
Add decorators to babel parser

wiseolddj:
We have a project that is using decorators. We would like to add the option to the babel parser so that Stryker will work on our project


nicojs:
@wiseolddj thanks for the PR. Unfortunately, we didn't have time to look at it before.  

I'm closing this PR in favor of #1474. Is that implementation also good for you?

PR #1344, opened 2019-01-30
refactor(typed-inject): Move typed inject to it's own repo

nicojs:
Move typed-inject to https://github.com/nicojs/typed-inject/


nicojs:
I don't think we can do a release without this, as typed-inject is already published with the version that lerna would choose

PR #1370, opened 2019-02-11
feat(ES5 support): remove ES5 mutator

simondel:
BREAKING CHANGE: Removed the ES5 mutator. The 'javascript' mutator is now the default mutator. Users without a mutator plugin should install `@stryker-mutator/javascript-mutator`.


simondel:
I have set the destination for this PR to the v1 branch.

simondel:
@nicojs ⚠️ This PR **did not** trigger a build because it's to the v1 branch ⚠️

simondel:

PR #1371, opened 2019-02-11
feat(reporter config): remove deprecated reporter config option

simondel:
BREAKING CHANGE: Removes the 'reporter' config option. Please use the 'reporters' (plural) config option instead.


simondel:

PR #1373, opened 2019-02-11
build: Include v1 branch in the branches to build

simondel:


simondel:
@nicojs should we push this to master or only the v1 branch? Currently, it's linked to the master branch.

PR #1374, opened 2019-02-11
feat(jest project config): Remove deprecated project setting and 'default' project name

simondel:
BREAKING CHANGE: removes deprecated 'jest.project' config key and project value 'default'. Please use the 'jest.projectType' config key and the 'custom' project value.


simondel:
This PR is linked to the `master` branch instead of v1

PR #1393, opened 2019-02-12
build(lerna versioning): stop using independent versioning

simondel:


simondel:
perhaps the version should still be 0.0.1 because we will release a breaking change.

nicojs:
It needs to be `0.x.x`.

image

PR #1407, opened 2019-02-14
chore(lerna): enable --force-publish to lock all packages together

nicojs:


simondel:
Why does Lerna do this 😭

PR #1419, opened 2019-02-25
feat(jest config): set notify to false in JEST_OVERRIDE_OPTIONS

ollelauribostrom:
I propose turning of jest notifications by default. Having test notifications turned on in a project can ofc be a great thing but when stryker runs the tests this adds a lot of noise (one notification each time stryker runs the tests against a mutant). 

### Summary of changes
- Adding `notify: false` to jestOverrideOptions.ts
- Updating tests to verify that this property is overridden

### DIFF
--- a/packages/jest-runner/src/jestOverrideOptions.ts
+++ b/packages/jest-runner/src/jestOverrideOptions.ts
@@ -11,6 +11,10 @@ const JEST_OVERRIDE_OPTIONS = Object.freeze({
 
     // Disable bail so the jest process does not quit with a non-zero exit code
     bail: false,
+
+    // Disable notifications for test results, this will otherwise show a notification about
+    // the results each time Stryker runs the tests
+    notify: false
 });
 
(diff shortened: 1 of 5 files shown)


simondel:
What does this look like without explicitly setting it to `off`?

And what does it look like with the setting explicitly set to `off` but Stryker's logging set to `'TRACE'`?

ollelauribostrom:
Looks like this without explicitly turning this off:
https://media.giphy.com/media/8UGFvYxcwWWDrblA77/giphy.gif

Trace:
https://gist.github.com/ollelauribostrom/27f9e8dc8afd2c8abc3a1391cd0fd0ad
(Note that im running Stryker programatically from a custom script if that could have an impact on the trace 🙂)

nicojs:
Lol @ notifications :)

Turning it off sounds like the best approach.

simondel:
Wow that's something you don't want to have enabled when doing mutation testing 😂

PR #1420, opened 2019-02-25
fix(jest-runner): fix stryker error on todo tests

ollelauribostrom:
Stryker currently treats todo tests `it.todo('some description')` as failed tests. Having these in your test suite will result in `There were failed tests in the initial test run`, and the mutation test run fails. This PR fixes this by treating tests marked todo as skipped instead.

### Summary of changes
- Checking for tests with 'todo' status when test result status is being determined
- Adding unit test to verify that tests with todo status is marked as skipped


simondel:
Nice find! Thanks for fixing this ❤️

nicojs:
Especially like the fact that you added tests 🥇

PR #1421, opened 2019-02-25
chore: add .DS_Store files to gitignore

ollelauribostrom:
A convenience for OSX users 🙂


simondel:
Nice catch! I thought we already had this ❤️

Issue #1249, opened 2018-11-22
Support for babel.config.js


Djaler:
### Summary

At the current moment I can only specify babelrc location, but I don't use it in my application

### Add stryker.log
```
17:17:52 (7962) INFO BabelConfigReader Reading .babelrc file from path "/home/kromanov/Projects/sow/frontend/babel.config.js"
17:17:52 (7962) ERROR BabelConfigReader Error while reading .babelrc file: SyntaxError: Unexpected token m in JSON at position 0
```

nicojs:
Thanks to you both. We're definitely aware of this and will add support in the near future. Stay tuned.

simondel:
We have Babel 7 support now, so we can start to do this 🎉

Mrna1:
@nicojs @simondel : Good day, Is this functionality already released? I updated stryker to last release versions and still gets:

> ERROR BabelConfigReader Error while reading "C:\...\.babelrc.js" file: SyntaxError: Unexpected token c in JSON at position 0

> "stryker": "^0.35.1",
> "stryker-api": "^0.24.1",
> "stryker-babel-transpiler": "^0.10.1",
> "stryker-html-reporter": "^0.18.1",
> "stryker-mocha-framework": "^0.15.1",
> "stryker-mocha-runner": "^0.17.1",
> "stryker-typescript": "^0.18.1",

nicojs:
Hi @Mrna1 could you please show me your babelrc.js file? It should contain JSON content (not javascript).

Mrna1:
@nicojs: So you do not support js files. Is that right? Will it be in the future?

nicojs:
Could you please give me an example of a babel config file that is a js file? I didn't know babel allows it. We want to reach feature parity between the stryker babel-transpiler with the babel cli, so a link to the official docs would help. 

It also helps if this issue gets upvoted because our time is limited and we're focussing on the features with the most impact.

nicojs:
Also, you can use arbitrary javascript with a stryker.conf.js file by `require`-ing your babel configuration:

```js
babel: {
    options: require('./babelrc.js')
     // Add extensions here
    extensions: [/*'.ts'*/]
}
transpilers: [
    'babel' // Enable the babel transpiler
],
```

Mrna1:
@nicojs: Thank you. I gonna try that. :)

jayands:
[Link to the official docs, you say?](https://babeljs.io/docs/en/configuration)

[...]

PR #1422, opened 2019-02-25
fix(babel-transpiler): accept .js babel config files

ollelauribostrom:
Babel supports a few different config file options as described [here](https://babeljs.io/docs/en/next/config-files#configuration-file-types). Stryker currently expects a `.babelrc`-type file (in other words a json file that can be read and parsed). This PR adds support for `.babelrc.js` and `babel.config.js` files as well so that there is no longer any need for duplicated configurations 🙂 

### Summary of changes
- Check if the provided optionsFile is a `.babelrc.js`-file. If so, require and return it.
- Check if the provided optionsFile is a `babel.config.js`. If so, require it, call it with an [optionsApi object](https://babeljs.io/docs/en/next/config-files#config-function-api) provided by the user and return the result
- Add integration tests for verify that this is working as expected


nicojs:
Thanks a lot! Does this issue close #1249?

PR #1447, opened 2019-03-13
test(wct): improve test stability

nicojs:


nicojs:
This should fix the build

Issue #656, opened 2018-03-20
Don't mutate `declare` types


JoshuaKGoldberg:
### Summary

If you have a `.d.ts` file that includes a `declare module "foo" { }`, the module name (`"foo"`) will be mutated during tests.

```diff
Mutant survived!
C:\Code\Project\src\typings\b2a.d.ts:3:15
Mutator: StringLiteral
-   declare module "b2a" {
+   declare module "" {
```

### Stryker config

```js
module.exports = function(config) {
  config.set({
    coverageAnalysis: "off",
    files: [
      "tsconfig.test.json",
      {
        included: false,
        pattern: "test/unit/jest.config.js",
        transpiled: false,
      },
      "src/**/*.ts",
      "src/**/*.tsx",
    ],
    jest: {
      config: require("./test/unit/jest.config.js"),
    },
    mutate: [
      "src/**/*.ts",
      "!src/**/*.test.ts",
    ],
    mutator: "typescript",
    reporter: ["clear-text", "progress"],
    testRunner: "jest",
    transpilers: ["typescript"],
    tsconfigFile: "tsconfig.json",
  });
};
```

### Stryker environment

```
+-- stryker@0.19.3
+-- stryker-api@0.13.0
```

### Your Environment
| software         | version(s)
| ---------------- | -------  
| node             |  9.8.0
| npm              |  5.6.0
| Operating System |  Windows 10

nicojs:
@JoshuaKGoldberg thanks for reporting this issue.

Just to be sure: you do want to mutate `.d.ts` files? If not, you can just ignore that by adding `"!src/**/*.d.ts"` in your `mutate` array.

JoshuaKGoldberg:
That's a great point - I should have done that.

nicojs:
Yeah, but you're point still is valid. It doesn't make sense to mutate declare statements 👍

ankhzet:
Possibly related: stryker mutates boolean constants in type decls, which produces false-negatives:

```
class A {
    methodB(): SomeType | false { // 'false' here is mutated to 'true'

     }
}
```

Haven't tested, but have suspicion, that string literal typedefs would be also mutated (`const foo: Type | 'enum1' | 'enum2' = ...`).
Also haven't tested boolean literals in interfaces.

config:
```
module.exports = function(config) {
  config.set({
    files: [
      '**/*.json',
      'src/**/*.ts?(x)',
      'tests/**/*',
    ],
[...]

PR #1458, opened 2019-03-21
Stopped mutating nodes with a `declare` modifier

JoshuaKGoldberg:
Starts on #656. If this is the right approach, I can also send a PR to remove type definition nodes.

### DIFF
--- a/packages/typescript/src/TypescriptMutator.ts
+++ b/packages/typescript/src/TypescriptMutator.ts
@@ -30,6 +30,10 @@ export class TypescriptMutator {
   }
 
   private mutateForNode<T extends ts.Node>(node: T, sourceFile: ts.SourceFile): Mutant[] {
+    if (shouldNodeBeSkipped(node)) {
+      return [];
+    }
+
     const targetMutators = this.mutators.filter(mutator => mutator.guard(node));
     const mutants = flatMap(targetMutators, mutator => mutator.mutate(node, sourceFile));
     node.forEachChild(child => {
@@ -39,3 +43,7 @@ export class TypescriptMutator {
     return mutants;
   }
 }
+
+const shouldNodeBeSkipped = (node: ts.Node): boolean => {
+  return node.modifiers !== undefined && node.modifiers.some(modifier => modifier.kind === ts.SyntaxKind.DeclareKeyword);
+};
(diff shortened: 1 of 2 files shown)


packages/typescript/src/TypescriptMutator.ts
```diff
@@ -30,6 +30,10 @@ export class TypescriptMutator {
   }
 
   private mutateForNode<T extends ts.Node>(node: T, sourceFile: ts.SourceFile): Mutant[] {
+    if (shouldNodeBeSkipped(node)) {
+      return [];
+    }
```
nicojs: Nitpicking: Could we put the rest of the method in an `else` expression for clarity?

packages/typescript/test/unit/TypescriptMutator.spec.ts
```diff
@@ -90,5 +90,25 @@ describe('TypescriptMutator', () => {
...
+
+          function mutated(a: number) { }`);
+      });
+
+      it('should skip nodes with declare keywords', () => {
+        const mutants = sut.mutate([
+          file1,
+        ]);
+        expect(mutants.filter(mutant => mutant.mutatorName === 'SourceFileForTest')).lengthOf(1);
+        expect(mutants.filter(mutant => mutant.mutatorName !== 'SourceFileForTest')).lengthOf(2);
```
nicojs: Maybe we can check the _exact_ state here since the mutators are controlled for this unit test. I realize that we didn't do that for other tests, but in this case, it would help clarify what you're doing IMHO.

packages/typescript/src/TypescriptMutator.ts
JoshuaKGoldberg: Funny, I prefer the [opposite](https://github.com/palantir/tslint/pull/4502). 😄

packages/typescript/test/unit/TypescriptMutator.spec.ts
```diff
@@ -105,8 +105,11 @@ describe('TypescriptMutator', () => {
         const mutants = sut.mutate([
           file1,
         ]);
-        expect(mutants.filter(mutant => mutant.mutatorName === 'SourceFileForTest')).lengthOf(1);
-        expect(mutants.filter(mutant => mutant.mutatorName !== 'SourceFileForTest')).lengthOf(2);
+        expect(mutants.map(mutant => mutant.mutatorName)).to.deep.equal([
+          'SourceFileForTest',
+          'FunctionDeclarationForTest',
+          'FunctionDeclarationForTest'
```
nicojs: I would also expect the location of the mutant to be tested, right? Since that's an important detail. We could also split it into 2 tests, one for a `declare function` and one for a `function`

JoshuaKGoldberg: Oh sure, that sounds good. +1

nicojs:
Awesome! Thanks 👍

PR #1474, opened 2019-03-29
feat(javascript-mutator): allow decorators

nicojs:


nicojs:
@wiseolddj is this helping you? This enables the `decorators` babel plugin.

PR #1476, opened 2019-04-01
feat(mutation testing elements): Implement new mutation testing elements html reporter

nicojs:


nicojs:
I think we should "Rebase and merge" rather than "Squash and merge", better for the changelog.

packages/core/src/reporters/MutationTestReportCalculator.ts
```diff
@@ -0,0 +1,117 @@
...
+  private mutationTestReport(results: ReadonlyArray<MutantResult>): mutationTestReportSchema.MutationTestResult {
+    return {
+      files: this.toFileResults(results),
+      schemaVersion: '1.0',
+      thresholds: this.options.thresholds
+    };
+  }
+
+  private toFileResults(results: ReadonlyArray<MutantResult>): mutationTestReportSchema.FileResultDictionary {
+    const files: mutationTestReportSchema.FileResultDictionary = Object.create(null);
```
simondel: This is quite sneaky. You're creating a variable with the same name as an instance property

packages/core/src/reporters/MutationTestReportCalculator.ts
```diff
@@ -0,0 +1,117 @@
...
+    results.forEach(mutantResult => {
+      const fileResult = files[mutantResult.sourceFilePath];
+      if (fileResult) {
+        fileResult.mutants.push(this.toMutantResult(mutantResult));
+      } else {
+        const sourceFile = this.files.files.find(file => file.name === mutantResult.sourceFilePath);
```
simondel: You're inside of an anonymous function, referencing `this.files.files` while you have a variable in the method above called `files` and you have an instance variable called `files` as well. It's quite confusing which files is which and why you need to do `files.files`

packages/html-reporter/src/templates/customElements.ts
```diff
@@ -0,0 +1,8 @@
+interface MutationTestReportAppHTMLElement {
+  titlePostfix: string;
+}
+declare namespace JSX {
+  interface IntrinsicElements {
```
simondel: Where is this used?

nicojs: It is how to add custom elements to typed-html: https://www.npmjs.com/package/typed-html#custom-elements

nicojs:
@simondel  I've fixed/commented on all review comments, is this better?

PR #1505, opened 2019-04-18
build(deps-dev): update nyc requirement from ^13.0.1 to ^14.0.0

simondel:
Set `--exclude-after-remap` to `false` due to it now being [enabled by default](https://github.com/istanbuljs/nyc/blob/master/CHANGELOG.md#breaking-changes).


packages/core/package.json
```diff
@@ -6,9 +6,9 @@
   "typings": "src/Stryker.d.ts",
   "scripts": {
     "start": "tsc -w",
-    "test": "nyc --check-coverage --reporter=html --report-dir=reports/coverage --lines 80 --functions 80 --branches 75 npm run mocha",
+    "test": "nyc --exclude-after-remap=false --check-coverage --reporter=html --report-dir=reports/coverage --lines 80 --functions 80 --branches 75 npm run mocha",
     "mocha": "mocha \"test/helpers/**/*.js\" \"test/unit/**/*.js\" && mocha --timeout 15000 \"test/helpers/**/*.js\" \"test/integration/**/*.js\"",
-    "@stryker-mutator/core": "node bin/stryker run"
```
nicojs: What's wrong with `@stryker-mutator/core` 😢

PR #1512, opened 2019-04-19
 feat(files): use mocha 6's way of resolving files

### LINKED ISSUE #1428: Support `--files` in mocha.opts
I'm trying to run Stryker on a Vue project with webpack and after fixing some other issues -mainly npm package errors- I'm stuck with this error:

`Error: [vue-test-utils]: window is undefined, vue-test-utils needs to be run in a browser environment. 
You can run the tests in node using jsdom`

I've created a quick-and-dirty example at [https://github.com/glezalex/vue-tdd](https://github.com/glezalex/vue-tdd) to show an example of this error, in case somebody could help me understand where this issue comes from :).

nicojs:
Use mocha 6's implementation to resolving files.
This adds support for `exclude`, `spec`, `files` as well as positional arguments.

Fixes #1428 

Note: #1511 should be merged first


nicojs:
This is now ready for review. 

@simondel I'm using https://github.com/mochajs/mocha/blob/master/lib/cli/run-helpers.js#L132 This is a yet undocumented API. Do you think we should work around this?

packages/mocha-runner/src/MochaOptionsLoader.ts
```diff
@@ -5,41 +5,61 @@ import { tokens, commonTokens } from '@stryker-mutator/api/plugin';
...
-      const args = serializeArguments(overrides);
-      const rawConfig = LibWrapper.loadOptions(args) || {};
-      if (this.log.isTraceEnabled()) {
-        this.log.trace(`Mocha: ${LibWrapper.loadOptions.name}([${args.map(arg => `'${arg}'`).join(',')}]) => ${JSON.stringify(rawConfig)}`);
-      }
-      return filterConfig(rawConfig);
+      this.log.debug('Mocha >= 6 detected. Using mocha\'s `%s` to load mocha options', LibWrapper.loadOptions.name);
+      return this.loadMocha6Options(overrides);
     } else {
       this.log.debug('Mocha < 6 detected. Using custom logic to parse mocha options');
```
simondel: Didn't we want to log this as a deprecation warning? Or is that somewhere else?

packages/mocha-runner/src/MochaOptionsLoader.ts
```diff
@@ -5,41 +5,61 @@ import { tokens, commonTokens } from '@stryker-mutator/api/plugin';
...
+
+  private loadMocha6Options(overrides: MochaOptions) {
+    const args = serializeArguments(overrides);
+    const loadOptions = LibWrapper.loadOptions || (() => ({}));
+    const rawConfig = loadOptions(args) || {};
+    if (this.log.isTraceEnabled()) {
+      this.log.trace(`Mocha: ${loadOptions.name}([${args.map(arg => `'${arg}'`).join(',')}]) => ${JSON.stringify(rawConfig)}`);
+    }
+    const options = filterConfig(rawConfig);
+    if (this.log.isDebugEnabled()) {
```
simondel: Why do you check if debug is enabled? in the `loadMochaOptions` above, you don't check this.

packages/mocha-runner/src/MochaOptionsLoader.ts
nicojs: Let's do that in a separate PR, before we release 2.0

packages/mocha-runner/src/MochaOptionsLoader.ts
nicojs: Because we do `JSON.stringify`. An action that's only needed if debug is enabled.

simondel: Check, so it's just a performance improvement.

packages/mocha-runner/src/MochaOptionsLoader.ts
simondel: Sure 👍

PR #1513, opened 2019-04-19
 feat(onScoreCalculated): deprecate onScoreCalculated reporter event

### LINKED ISSUE #1509: Deprecate onScoreCalculated event
We should deprecate the `onScoreCalculated` event because we now have `onMutationTestReportReady`.

We should also use http://npmjs.com/package/mutation-testing-metrics

simondel:
The onScoreCalculated reporter event is deprecated. Please use 'onMutationTestReportReady' and calculate the score as an alternative. The EventRecorderReporter still implements this event to prevent breaking changes.

Fixes #1509


simondel:
You can see the warning in action here btw:
https://travis-ci.org/stryker-mutator/stryker/jobs/522083900#L2393

image

packages/core/src/reporters/EventRecorderReporter.ts
```diff
@@ -70,14 +70,14 @@ export default class EventRecorderReporter implements StrictReporter {
...
 
-  public onScoreCalculated(score: ScoreResult): void {
-    this.work('onScoreCalculated', score);
-  }
-
   public onMutationTestReportReady(report: mutationTestReportSchema.MutationTestResult): void {
     this.work('onMutationTestReportReady', report);
   }
 
+  public onScoreCalculated(score: ScoreResult): void {
```
nicojs: Does this mean that you'll get the deprecation warning all the time when this reporter is enabled?

packages/core/test/helpers/producers.ts
```diff
@@ -156,6 +157,38 @@ export const scoreResult = factoryMethod<ScoreResult>(() => ({
   transpileErrors: 0
 }));
 
+export const metrics = factoryMethod<Metrics>(() => ({
```
nicojs: Can we move this to the `@stryker-mutator/test-helpers`? It might already be there, because it is also used in the html reporter

packages/core/src/reporters/EventRecorderReporter.ts
simondel: Yes, you'll get a warning at the end of every run if you have this reporter enabled

simondel:
The build is red right now because of mutation-testing-metrics. It doesn't support Node 6 (it uses Object.values)

packages/core/src/reporters/EventRecorderReporter.ts
nicojs: Yeah, I guess it's not a big deal

nicojs:
Fixed it by using this polyfill: https://github.com/tc39/proposal-object-values-entries/blob/master/polyfill.js

We can remove it later once we've dropped support for node 6

PR #1518, opened 2019-04-19
build: output es2017 code

simondel:


nicojs:
I think this PR changes too many tsconfig.json files. Should only be the once governing Stryker IMHO.

kmdrGroch:
I don't know if it is good that you are removing Libs. I don't think ES2017 contains DOM lib etc. It should all be added as it was in my opinion.

nicojs:
@kmdrGroch `es2017` indeed doesn't include DOM, but it shouldn't as Stryker runs in NodeJS, right?

simondel:
I have removed the other tsconfig changes.

kmdrGroch:
@nicojs, @simondel change target to "target": "es2017" and just add to libs es2017. Do not remove older libs like es2015, etc

kmdrGroch:
and @nicojs all these libs are nothing more as the sets of interfaces. It is the same as you are using @types/mocha for example. It is like using @types/ecmascript5, @types/DOM.

nicojs:
@kmdrGroch 
> change target to "target": "es2017" and just add to libs es2017. Do not remove older libs like es2015, etc

I'm pretty sure that `es2017` includes 2015 etc. See https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es2017.d.ts

> and @nicojs all these libs are nothing more as the sets of types / interfaces. It is the same as you are using @types/mocha for example. It is like using @types/ecmascript5, @types/DOM.

I know. However, I _don't want those types_. For example, I don't want to accidentally add `document.getElementById` somewhere in our code.

kmdrGroch:
yea, you are right. They have reflections to previous versions, my fault.

nicojs:
Yeah, we'll have 16th and 17th this month dedicated to Stryker development. We'll release Stryker 2.0 then.

PR #1529, opened 2019-04-28
feat(mocha): deprecate mocha version 5 and below

simondel:


packages/mocha-runner/src/MochaTestRunner.ts
```diff
@@ -27,6 +27,7 @@ export default class MochaTestRunner implements TestRunner {
       this.log.debug('Mocha >= 6 detected. Using mocha\'s `handleFiles` to load files');
       this.testFileNames = this.mocha6DiscoverFiles(LibWrapper.handleFiles);
     } else {
+      this.log.warn('DEPRECATED: Mocha < 6 detected. Please upgrade to at least Mocha version 6.');
```
nicojs: Probably better to do this in the `MochaConfigEditor`. That way, it is only logged once, instead of once per test runner

simondel: I have moved it to the MochaOptionsLoader. That logic is called from the MochaConfigEditor and there was already a check for a 'legacy' Mocha version there.

nicojs: LGTM

PR #1558, opened 2019-05-24
adapt config for manomano need

trollepierre:
DO NOT MERGE IT.

Here is the small adjustment that I needed to do to make my ejected React project (so not based on CRA) work with stryker.

I believe if you find a way to add an option to configure optional JestConfig with stryker, then using stryker with react will be very easy!

### DIFF
--- a/packages/jest-runner/src/configLoaders/ReactScriptsJestConfigLoader.ts
+++ b/packages/jest-runner/src/configLoaders/ReactScriptsJestConfigLoader.ts
@@ -14,7 +14,7 @@ export default class ReactScriptsJestConfigLoader implements JestConfigLoader {
 
   public loadConfig(): Configuration {
     // Get the location of react script, this is later used to generate the Jest configuration used for React projects.
-    const reactScriptsLocation = path.join(this.loader.resolve('react-scripts/package.json'), '..');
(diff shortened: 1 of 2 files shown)


nicojs:
I'm not familiar with `manomano`. Is it an alternative to CRA?

trollepierre:
Explanation:
In this project, we ejected from CRA + we export the configuration inside another library that is used by other projects.

My point is:
 every place where I made changes, should be configured in stryker.conf.js . 

By default, stryker.conf.js uses react-scripts to get jest.config. However for other specific projects, that have ejected, stryker should let the ability to configure jest.config path (as it is custom).

nicojs:
Is this what you're looking for?

```
{
    jest: {
        projectType: 'custom',
        config: require('path/to/your/custom/jestConfig.js'),
        enableFindRelatedTests: true,
    }
}
```

Documented here: https://github.com/stryker-mutator/stryker/tree/master/packages/jest-runner#configuring-stryker

trollepierre:
Looks like what I expect. I will test next week I will let you know:-)

trollepierre:
Maybe
Documented here: https://github.com/stryker-mutator/stryker/tree/master/packages/jest-runner#configuring-stryker
should be documented/linked in
 https://github.com/stryker-mutator/stryker 
or
https://stryker-mutator.io/stryker/quickstart

PR #1603, opened 2019-06-26
ci: increase timeout for the perf test to 20 minutes

DanielRuf:
This PR should increase the maximum timeout to 20 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,5 +33,3 @@ stages:
   - test
   - name: release
     if: branch = release AND fork = false
-  - name: "Performance test"
-    if: type = cron # Only run in nightly build
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,15 @@ jobs:
       script: npm run build && npm run e2e
     - stage: Performance test
       name: "npm run perf"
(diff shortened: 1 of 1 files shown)


nicojs:
Awesome! Didn't know this was possible. I'll merge it in when green.

DanielRuf:
I've temporarily enabled the performance test to see the actual result.

See https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out for common solutions.

DanielRuf:
Yep, like I've expected it. Let's see if the other recommended solution works using a background process.

Reference: https://github.com/travis-ci/dpl/issues/568#issuecomment-272465349

DanielRuf:
I'll inspect execa and the underlying spawned child process. There is probably another part which has a timeout.

nicojs:
Hi @DanielRuf I see you've deleted the branch. Does that mean you've given up on this? 😢

What was the problem you ran into?

DanielRuf:
It's fixed by removing the execa timeout (my other PR which is already merged).

nicojs:
Ah yes, realized that later. Thanks for figuring out what's wrong!

PR #1604, opened 2019-06-26
ci: disable execa timeout to prevent failing perf job

DanielRuf:
https://github.com/sindresorhus/execa/blob/master/readme.md#timeout

> If timeout is greater than `0`, the parent will send the signal identified by the `killSignal` property (the default is `SIGTERM`) if the child runs longer than timeout milliseconds.

Let's see if this improves the current CI issue with the performance tests.

Otherwise I'll have to install the complete project with all dependencies when I can use the other laptop (this one overheats too much - which is not great at around 30°C here).

### DIFF
--- a/perf/tasks/run-perf-tests.ts
+++ b/perf/tasks/run-perf-tests.ts
@@ -36,7 +36,7 @@ function execNpm(args: string[], testDir: string): Observable<string> {
   console.log(`Exec ${testDir} npm ${args.join(' ')}`);
 
   return new Observable(observer => {
-    const testProcess = execa('npm', args, { timeout: 500000, cwd: currentTestDir, stdio: 'pipe' });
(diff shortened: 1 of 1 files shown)


DanielRuf:
Which would perfectly match with https://travis-ci.org/stryker-mutator/stryker/jobs/550797307#L889-L892.

DanielRuf:
Do not yet merge. I'll remove the commit which removes the job condition.

DanielRuf:
This PR can be merged now. The perf test was green and no timeout occurred.

DanielRuf:
See https://travis-ci.org/stryker-mutator/stryker/jobs/550850365

nicojs:
Awesome! Thanks a lot for this PR!

Issue #1588, opened 2019-06-17
consistency in files naming


kmdrGroch:
small issue related to folders and files.
I have noticed that in most of the cases you are using "registerChaiPlugins", where all lines are the same. But you have one file, in "HTML-reporter" - "chaiHelpers" with the same code, but another name. It should be the same name I guess.

nicojs:
Yeah should be, feel free to fix that in a small PR. 👍

PR #1605, opened 2019-06-26
maintenance: provide one naming convention

kmdrGroch:
fixes: #1588


kmdrGroch:
I found these names the most accurate since you are using a lot of 'initSomething'. And as you can see (mostly on chai's) you have 3-4 different names for the same 1 file :D

nicojs:
Awesome, thanks! 😅

Issue #1589, opened 2019-06-17
Update dependencies + fix issues related with that


kmdrGroch:
By mistake, I have updated some of the files (it was typescript), and after some bug fixing related to that, I have noticed that you are missing a lot of @types and most of them are obsolete!
for example:
currently installed
```
"@types/glob": "^7.1.0",
"@types/istanbul": "^0.4.29",
"@types/karma": "^3.0.0",
"@types/lodash": "^4.14.110",
"@types/mkdirp": "^0.5.2",
"@types/mocha": "^2.2.44",
"@types/node": "^10.11.5",
"@types/rimraf": "2.0.2",
"@types/sinon": "5.0.5",
"@types/sinon-chai": "^3.2.0"
```

current:
```
"@types/glob": "^7.1.1",
"@types/istanbul": "^0.4.30",
"@types/karma": "^3.0.3",
"@types/lodash": "^4.14.134",
"@types/mkdirp": "^0.5.2",
"@types/mocha": "^5.2.7",
"@types/node": "^10.11.5",
"@types/rimraf": "2.0.2",
"@types/sinon": "7.0.13",
"@types/sinon-chai": "^3.2.2",
```
some of them don't have super changes but other like sinon / mocha are quite obsolete
\+ you in one package i have noticed you don't use @types for jest (it can cause some type error in the future)

nicojs:
With regards to TypeScript 3.5, I'm waiting for the fix for https://github.com/microsoft/TypeScript/issues/31676 to be released (will hopefully be 3.5.2). After that I we will be able to update TypeScript itself.

TL;DR;
I'm happy to accept a PR that changes all `^` (minor version) ranges to `~` (patch version) ranges. For more info on `^` and `~` see https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept. 
I'm also happy to accept a PR for mocha types.

Long version:

With regards to the `@types` dependencies, they're not exactly "obsolete". We're using [dependabot](https://dependabot.com/) to automagically update our dependencies. The thing is, it will only create a PR if the new version is outside of the semver range. For example `"@types/lodash": "^4.14.134",` is still in ranche of `"@types/lodash": "^4.14.110",` and so it will not create a PR for it. We're also happy with not following patch releases as it would create a lot of PR's. 

I would agree that we should never use `^`, instead use `~`. That way we will get a PR for minor releases.

[...]

PR #1608, opened 2019-06-26
maintenance: change ^ to ~ in packages

kmdrGroch:
Partially solves #1589.
Still not updated packages:
stryker\packages\core\package.json - @stryker-mutator/api
stryker\e2e\package.json - @types/semver, semver, ts-node, typescript, webpack
stryker\package.json - @types/mocha, @types/sinon, execa, tslint, typescript
(i still might be missing some)

Important to change:
stryker\packages\core\package.json - @stryker-mutator/api
stryker\package.json - @types/mocha, @types/sinon
But it will cause a lot of issues (It might be a PR more than 1 person should work)

less important but still: tslint, typescript
will cause more issues :/

Note:
Stryker supports node >= 8.
But we are using types for node 10.x.x
should we change it to 8.x.x for now, until support for 8 is dropped?


kmdrGroch:
@nicojs little guy is waiting for approval :x

Archcry:
We just merged some dependabot stuff, could you update the pull-request :see_no_evil:

kmdrGroch:
@nicojs, @simondel, @Archcry  could we please wait with updating dependencies until we fix this issue? :)

nicojs:
> Stryker supports node >= 8.
> But we are using types for node 10.x.x
> should we change it to 8.x.x for now, until support for 8 is dropped?

No, we shouldn't change it to 8.x.x. We did this practice in the past, however as the typings for 8.x.x are effectively frozen we don't get all the awesome typing improvements the good guys over at definitely typed provide. That's why we use the latest typings and run indeed the risk of using something that we shouldn't. We trust our tests enough.

Will merge after conflicts are solved

Issue #1594, opened 2019-06-20
maintenance: rename Types / Interfaces / Enums


kmdrGroch:
Lurking through the code I can see that all types, interfaces, enums, are just ex 'strykerOptions'. It doesn't give a lot of info about what it is (especially in imports). Since we are using a lot of types, some may be the same in different packages (Range is already in Typescript).

My idea is to rename all of them, so it is easier to know, what exactly we are dealing with (without going to declaration etc.)

Types: starting with T - TRange
Interfaces: starting with I - IStrykerConfig
Enums: starting with E - ELogLevel

nicojs:
A kind of [hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) for naming types, interfaces and enums is not recommended for typescript. Using an `I` for interfaces is something that you do a lot in C#, but it is also not accepted in the larger typescript community.  

I'm of the opinion that the code should speak for itself and be clear, no matter what the exact type of things is. Your code editor should help you to determine the exact types of things if you need it.

When we started with Stryker, there were no style guides. Now there are. I think we're mostly in line with this one: https://basarat.gitbooks.io/typescript/docs/styleguide/styleguide.html#interface

Maybe we should document this in the contributing.md? I'll be happy to accept a PR for it :)

kmdrGroch:
we should create our own personalized tslint config so everything will be documented there (we can take all available properties and set them accordingly to our needs) [i can prepare prepare PR for both tslint config and contributing.md)

nicojs:
That makes sense. But we first should try to look for ones that implement basarat's style guide (or ESlint plugins that do that).

kmdrGroch:
@nicojs I could lurk a bit. If it already exists I can extend it, if not, I can make one based on basarat's style guide and 'my personal preferences'

kmdrGroch:
@nicojs I have some problem with linting, since in your styleGuide there is 'use camelCase' and you are using in dozen of files 'PascalCase'. I'll leave it for now. You will decide what to do with that later

PR #1615, opened 2019-07-01
tslint: match basarat styleGuide (#1594)

kmdrGroch:
Now you can choose whatever naming convention for variables and files do you want


nicojs:
Great! Fixes #1594

Issue #1614, opened 2019-07-01
StrykerCli an error occurred SyntaxError: Unexpected character '#'


simivar:
### Summary

I am using Stryker with Babel as transpiller and in one of my classess I've got private property `#hash`. When I try to run Stryker on that class it fails.

Class:
```js
const something = require('something');
const something2 = require('something2');
const something3 = require('something3');

class Example {
    #hash = '';

    constructor(hash) {
        this.#hash = hash;
    }
}
```

Error I am getting:
```
16:45:00 (12996) ERROR StrykerCli an error occurred SyntaxError: Unexpected character '#' (6:4)
    at Object.raise (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:6344:17)
    at Object.readToken_numberSign (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:6671:12)
    at Object.getTokenFromCode (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:7017:14)
    at Object.getTokenFromCode (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:3631:18)
    at Object.getTokenFromCode (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:2306:20)
    at Object.nextToken (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:6542:12)
    at Object.next (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:6482:10)
    at Object.eat (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:6487:12)
    at Object.expect (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:7645:10)
    at Object.parseClassBody (C:\Users\my_username\PhpstormProjects\my-project-name\node_modules\@babel\parser\lib\index.js:10562:10) {
  pos: 166,
  loc: Position { line: 6, column: 4 }
}
```
### Stryker config
```js
module.exports = function (config) {
    config.set({
        mutator: 'javascript',
        packageManager: 'npm',
        reporters: ['html', 'clear-text', 'progress'],
        testRunner: 'mocha',
[...]

PR #1623, opened 2019-07-07
feat: add option to provide babelPlugins

kmdrGroch:
fixes: #1614


kmdrGroch:
It should give a possibility to add plugins into javascriptMutator like so:
`stryker.config.js`
```
mutator: {
  name: 'javascript';
  excludedMutations: [];
  babelPlugins: ['classPrivateProperties', 'classProperties', 'classPrivateMethods'];
}
```

Note:  if you like this change - we need to provide tests to it (i guess using #1614 as an example) and update docs

packages/api/src/mutant/Mutator.ts
```diff
@@ -1,6 +1,7 @@
 import { File } from '../../core';
 import Mutant from './Mutant';
+import { ParserPlugin } from '@babel/parser';
 
 export default interface Mutator {
-  mutate(inputFiles: ReadonlyArray<File>): ReadonlyArray<Mutant>;
+  mutate(inputFiles: ReadonlyArray<File>, options?: ParserPlugin[]): ReadonlyArray<Mutant>;
```
nicojs: Instead of changing the api of a mutator, please let the `StrykerOptions` be injected in the `JavaScriptMutator` here: https://github.com/stryker-mutator/stryker/blob/master/packages/javascript-mutator/src/JavaScriptMutator.ts#L13-L16. If you want, I can add this for you, so you can just use it there.

packages/core/src/mutants/MutatorFacade.ts
```diff
@@ -15,7 +15,10 @@ export class MutatorFacade implements Mutator {
 
   public mutate(inputFiles: ReadonlyArray<File>): ReadonlyArray<Mutant> {
     const allMutants = this.pluginCreator.create(this.getMutatorName(this.options.mutator))
-      .mutate(inputFiles);
+      .mutate(
+        inputFiles,
+        typeof this.options.mutator !== 'string' ? this.options.mutator.babelPlugins : []
```
nicojs: This should not be done in the `Facade`, but in the `JavaScriptMutator` itself.

kmdrGroch: I had to pass options somehow. Honestly no idea how can i do it other way

kmdrGroch:
@Archcry There is some issue with `mutatorDescriptor` in ` packages/javascript-mutator/src/JavaScriptMutator.ts`. 
It seems it can be `undefined` in some cases.
Changing line `22` to 
```js
const babelPlugins = (this.mutatorDescriptor || {}).babelPlugins || [];
```
solve that but tests still aren't passing.
Could you please take a look on that / fix it / provide some explanation about it so i can fix it?

kmdrGroch:
@nicojs I think it finally is done, let's just wait for the green :)

kmdrGroch:
@nicojs You might want to take a look since this is a quite important feature 🏭 🚀

PR #1660, opened 2019-07-25
feat(#1472): add excluding specific mutations

### LINKED ISSUE #1472: Ignore specific mutations
In some projects, there are mutations that users want to exclude. An example of this is logging. We've had multiple questions of users that do not want to test (some) logging.

## Why?
1. Not all code is worth testing (using unit tests)
1. Wanting to maintain a 100% mutation score and be alerted of any new surviving mutants

## Why not?
1. Excluding mutations gives a false sense of quality: "I've tested 100% of my code, except for the code I haven't tested"
1. If it's not important to test a piece code, you may not find that piece of code important at all, so removing it would be preferred.
1. Adds complexity to the usage and development of Stryker

## Proposed implementation
We can use inline comments to disable certain mutators or disable Stryker entirely. For example:

* `// stryker-mutator disable next` or `/* stryker-mutator disable next */`: this would disable all mutators Stryker on the next AST node (statement)
* `// stryker-mutator disable next ['StringLiteral']` or `/*stryker-mutator disable next ['StringLiteral']*/`: this would only disable the `StringLiteral` mutator for the next AST node (statement).
* We *might* also want to allow people to exclude some patterns via stryker configuration. First, we would need to implement basic ignoring.

During the mutating process, we could keep track of the comments. We should _still_ generate the mutants, but output them as `"MutantState.Ignored"` (new state) so it can be reported accordingly by different reporters.

## Timeline
We want to introduce a **parser API** to support an easy and safe way to make the exclusions. This issue will lay the groundwork for it: #1893. Before this is done, we will not support this option.

_This main post will be kept up to date with the latest info_

## Related issues:

#1174, #1470, #1464, #1653 

## Related pull requests:
#1810, #1912, #2084

kmdrGroch:
fixes #1472


kmdrGroch:
for now only
```
stryker:on
stryker:off
stryker:on MutatorName...
stryker:off MutatorName...
```
are working

```
stryker:next-on
stryker:next-off
```
are not implemented yet, since there were several problems with it

kmdrGroch:
`next-on` and `next-off` are working now!
Next thing is to do the same for `typescript` mutator :)

kmdrGroch:
Help needed. I can't extract a comment from TS file :/ any help would be nice

Djaler:
@kmdrGroch https://github.com/Microsoft/TypeScript/issues/21049#issuecomment-355756511

kmdrGroch:
@Djaler I saw it, but it didn't really help me :/

kmdrGroch:
@Djaler so with help of that comment, I somehow managed to get these comments.
But now I need some help with tests. Some of them aren't passing. @nicojs, @simondel, @Djaler any idea? :<

nicojs:
I would be happy to accept a merge request that only implements the babel-side of things. We will need to make clear remarks in the readme's

kmdrGroch:
I'd like https://github.com/stryker-mutator/stryker/pull/1718 this PR to be accepted, so I can start this over again.

Issue #1668, opened 2019-08-08
add `maintenance` 'issue type'


kmdrGroch:
could we make a template for `maintenance`?

Things like adding docs, new styles, things to refactor etc. All these issues could be marked as `maintenance`

PR #1670, opened 2019-08-08
feat: add maintenance issue_template

kmdrGroch:
fixes: #1668


kmdrGroch:
If you want I could be assigned by default, since I am doing a lot of work with TSLint, smaller refactors etc (+ I would like to learn / do even more in that direction :) )

nicojs:
Sure! Assigned by default if you want :). Do you want to make the change before I merge?

kmdrGroch:
@nicojs that should do :)

PR #1685, opened 2019-08-18
build(deps-dev): Update @types/sinon-chai from ^3.2.0 to 3.2.2

apalaniuk:
Builds started failing at #1678 due to pulling in a patch version for sinon-chai (3.2.3) which causes builds to fail due to mismatched chai typings.


kmdrGroch:
@apalaniuk You might want to contribute to #1608? I am trying to change all ^ to ~ or *nothing* to fix all deps and typings

simondel:
Thanks for fixing this <3

PR #1706, opened 2019-09-07
maintenance: change deps to ~version

kmdrGroch:


kmdrGroch:
that should be less invasive than previous one so it should pass

PR #1707, opened 2019-09-07
fix: jest typings

kmdrGroch:


kmdrGroch:
there were some issues with typescript about `import jest from 'jest'` since our typings are not `jest` at all, changed the namings so there should be no conflict now

kmdrGroch:
it is not working at all so I close this

PR #1718, opened 2019-09-11
maintenance: cleanup code

kmdrGroch:
small cleanup on javascript-mutator package.
Removing of `void` type from the `mutate` function and then just checking the `length` in `JavaScriptMutator`.


packages/javascript-mutator/src/mutators/ArrayNewExpressionMutator.ts
```diff
@@ -7,15 +7,13 @@ import { NodeMutator } from './NodeMutator';
...
 
-  public mutate(node: types.Node, copy: <T extends types.Node>(obj: T, deep?: boolean) => T): void | types.Node[] {
-    const nodes: types.Node[] = [];
-
+  public mutate(node: types.Node, copy: <T extends types.Node>(obj: T, deep?: boolean) => T): types.Node[] {
     if ((types.isCallExpression(node) || types.isNewExpression(node)) && types.isIdentifier(node.callee) && node.callee.name === 'Array' && node.arguments.length > 0) {
       const mutatedNode = copy(node);
       mutatedNode.arguments = [];
-      nodes.push(mutatedNode);
+      return [mutatedNode];
```
nicojs: Why is this better?

packages/javascript-mutator/src/JavaScriptMutator.ts
```diff
@@ -25,7 +25,7 @@ export class JavaScriptMutator implements Mutator {
         this.mutators.forEach(mutator => {
           const mutatedNodes = mutator.mutate(node, copy);
 
-          if (mutatedNodes) {
+          if (mutatedNodes.length) {
```
nicojs: With this improvement, we don't need to check on anything, right? We can just remove this if all together ❤

kmdrGroch: even if this is an array, we should check if the length is > 0 to ensure we won't run unnecessarily some part of code

packages/javascript-mutator/src/mutators/ArrayNewExpressionMutator.ts
kmdrGroch: previously we had 3 steps in order to get value:
create an array,
push element,
return array.

Right now we are doing these 3 steps in 1 step.

kmdrGroch: about performance: after small check:
```js
function func() {
 	const k = [];   
    k.push({"hello": "hi"});
    return k;
}
```js
per = performance.now(); for(let i = 0; i < 1000000; i++) {
    func();
} console.log(performance.now() - per);
```
and
```js
function func2() {
    return [{"hello": "hi"}];
}
```
```js
per = performance.now(); for(let i = 0; i < 10000000; i++) {
    func2();    
} console.log(performance.now() - per);
```

it seems, my solution is 20 times faster

packages/javascript-mutator/src/JavaScriptMutator.ts
kmdrGroch: take note that array.push is quite an expensive function so we shouldn't overdo it. Also, the spread operator is quite a performance killer (even if both are just taking small parts of ms, we ms to ms and you get s)

nicojs: Readability is more important than performance. Especially here, it won't make a difference. The push function will be called without arguments, which should be fast to execute. 

In any case, we shouldn't care about it. Generating mutants in memory takes less than ~0.01% of the total mutation testing time, right? If we can make it more readable and maintainable by removing this code branch, we should IMHO.

kmdrGroch: @nicojs and what about leaving these 'push' but at least removing 'void' type from mutators?

nicojs: Do you mean removing the `void` from the return type of `mutate` in the different mutators? Yeah, I'm all for that! Cleans up the code nicely 💐

kmdrGroch: It will be done with easy (soon. I am sick as hell, but probably I'll manage to do it tomorrow)

nicojs: Wow man.. just get well, don't worry about it. @simondel and I carved out some time in our schedules to spent on Stryker next week. We'll be merging all PR then as well as fix some issues ourselves. 🎉

kmdrGroch: @nicojs its done btw XD

kmdrGroch:
@nicojs could you review it after changes / merge it? I want to start #1472 after this is merged ._.

kmdrGroch:
@nicojs :<

kmdrGroch:
@simondel can you tell nico to review it again? Or can you review? :<

Issue #1719, opened 2019-09-11
ESlint migration


kmdrGroch:
Should I start working on migration to ESlint soon?
https://github.com/palantir/tslint/issues/4534
since we are close to the `November 1st, 2019` mark, I think we could be thinking about implementing it sooner or later

kmdrGroch:
I'll slowly start migrating our config since next week. It would be nice if you assign me so I don't forget about it ;D Also I have already known about that tool, although it still has some issues I guess I can handle it more or less :)

nicojs:
Probably this will be a good time to introduce prettier as well (#1271) using the prettier plugin for eslint. Instead of migrating our config, we could also base our config on Airbnb's one (https://www.npmjs.com/package/eslint-config-airbnb-base), since it would save us a lot of time, as well as being aligned with somewhat of an industry standard. What do you think about this?

I'll publish a small setup (something like "typescript-eslint-rbnb-prettier-starter") to github soon since I'm working on that for my day job at the moment. Probably later today.

kmdrGroch:
well, isn't it better to keep our config like it is right now (I would like it even 'stronger' since some whitespaces are quite annoying in some places, but whatever), I will do it in a few days since I have just finished exams on the university. And sure I can work with prettier. But do you want me to add both eslint and prettier in one PR or should I make 2 separate PRs?

nicojs:
This is an example: https://github.com/nicojs/typescript-eslint-prettier-starter

> well, isn't it better to keep our config like it is right now 

That's fine 2. It will save another dev-dependency we need to maintain, so good decision.

> prettier in one PR or should I make 2 separate PRs

You can make it 1 PR, I don't mind. Let's keep a `printWidth` for prettier of 100 or something (not too small).

> I will do it in a few days since I have just finished exams on the university

Much success my friend! Hope you do well! 🤞

Issue #1271, opened 2018-12-07
Add prettier to development flow


josecolella:
### Summary

Hello @nicojs , I would like to add prettier to the development flow to ensure that all files are formatted correctly.

Please let me know any specifics before starting on this task, such as single quotes and maximum line length

nicojs:
Awesome! Thanks for opening this issue and your offer to help. It is truly appreciated. 

How will the workflow work? The way I see it, it could work 3 ways

1, Run as a "prebuild" (or other)
1. Run as a githook every time you commit (prettier shouldn't change functionality, only styling)
1. Configure it in the editor (and document it very well).

We could also combine more of them. Editor support seems like a logical one.

As for your specifics:

* Single quotes
* Column size. How about 130? How does this work? Will long strings be broken up for example?

simondel:
I would also prefer editor support :)

nicojs:
I agree, but just editor support is not enough IMHO. I want a way to validate it as well. Don't know exactly what is possible and preferred though.

kmdrGroch:
Since we are using TSLint, is it nessesary to use prettier?
If not, we can close this issue :)

nicojs:
Yeah, actually, I'm looking to migrate away from tslint (since it [will be deprecated soon anyway](https://github.com/palantir/tslint/issues/4534)). 

Prettier [easily integrates with linters](https://prettier.io/docs/en/integrating-with-linters.html) especially eslint. So it would still be nice to have it.

A linter and a formatter are similar, but they still do different things.

(it's just a matter of time before someone creates a "prettier-eslint" do-it-all-tool that does both ... 😑 )

PR #1731, opened 2019-09-18
refactor: ESlint migration & prettier

kmdrGroch:
fixes: #1719
fixes: #1271


kmdrGroch:
Nico... I am soo sorry :< and you probably know why. But this prettier was so nicely configured by default and that `array-type` looks much better now >_<

kmdrGroch:
travis... hello? I have removed TSlint -.-

kmdrGroch:
@nicojs help... Travis hates me ;-;

kmdrGroch:
IT'S GREEN *YAAAAAAAAS* 🎉

kmdrGroch:
@nicojs If you could check & possibly merge this it would be nice <3 :3

nicojs:
This looks really good! I'll have a closer look later today and if everything goes well I'll merge it. Thanks a lot! 👍

nicojs:
I've added configure to fix-on-save in vscode. 

I absolutely love some of the rules eslint brings to the table. We enable more rules :). For example `@typescript-eslint/no-floating-promises`. There is no reason to have dangling promises, and they probably all bugs. Enabling the rule will show a couple of bugs. We can enable rules in separate issues IMHO, not in this one.

kmdrGroch:
Love it @nicojs! For now it would be nice if you merge it. Then we can discuss other rules which should be changed! (Take a consideration that for now @typescript-eslint is ~beta version - it doesn't have all features it had before, and so some are missing like `ordered-imports` [unless they changed it in some last versions])

nicojs:
Yeah, I've tried to enable ordered imports. It works, but it doesn't have the autofix feature: https://eslint.org/docs/rules/sort-imports

> The `--fix` option on the command line automatically fixes some problems reported by this rule: multiple members on a single line are automatically sorted (e.g. import { b, a } from 'foo.js' is corrected to import { a, b } from 'foo.js'), but multiple lines are not reordered.

Never mind that for now.  Thanks a lot once again. Logging off for tonight 🍻

Issue #1732, opened 2019-09-20
Cannot find module '@stryker-mutator/util' (VSCode issue)


kmdrGroch:
Basically, in each file I get an error that '@stryker-mutator/util' cannot be found. It is for sure installed, I have checked node_modules twice. Every other stryker package works just fine, only that one has some problems (at least it's one I have noticed). Do you think it is related to VSCode itself, Typescript or maybe it's our fault?

kmdrGroch:
The same issue is with `test-helpers`

PR #1736, opened 2019-09-23
fix typings problem

kmdrGroch:
fixes: #1732


kmdrGroch:
So I have made some research about it since it was really bugging me (of course found nothing relevant in the web), so I just started messing around with files. It seems packages do only consist of TS files, no declaration files at all. And so if you make `typings` in `package.json`, typescript is like `oww you have .d.ts file. Nice. I'll not bother making a new one`, then it says `oww... you don't have a file you told me you have. So f u.`.  After removing `typings` typescript makes it by it's own and fixes it.

kmdrGroch:
@nicojs or @simondel, could you please merge it? :/

kmdrGroch:
Please merge it. It's super annoying bug -.-
image

nicojs:
This is strange. I never had this issue. How are you compiling the typescript code? You should use `npm start` in the root package, or `CTRL+Shift+B` (`⌘+Shift+B` on OSX) to start the typescript compiler with `--build` as command-line option.

The `*.d.ts` files _actually are generated_. However, they are _hidden from vscode file explorer_. The reason to hide them is that otherwise, you'll get mad by all the build output. This is an example from the test-helpers package: 
image

kmdrGroch:
we are not using `typings` setting in all packages tho. + in one of the files we are actually pointing to ts file not .d.ts. It might be working after using `npm start`, but it works always if we remove `typings` at all

kmdrGroch:
This issue works like:
if you declare `typings` in `package.json`, typescript (at least in VSCode), will only look for this file. If this file doesn't exist, the whole package is without types, and then VSCode says that module doesn't exist. If we remove `typings`, VSCode is going to automatically search in files for types (it won't throw an error that module doesn't exist or type doesn't exist)

simondel:
I also sometimes have this, but I'm not sure why/when. I haven't been in the code lately.

kmdrGroch:
@simondel if you dont run `npm start` -> you dont create `.d.ts` files

nicojs:
I just figured out that `typings` is optional. If you specify a `main` in your package.json and point it to, for example, `src/a.js`, TypeScript will automatically look for `src/a.d.ts`, which makes sense when you think about it, with `"moduleResolution": "node"`. 

Lets remove all our `typings` attributes.

Good catch @kmdrGroch !

kmdrGroch:
@nicojs these are all I have found :)

nicojs:
Found 2 more 🙈

kmdrGroch:
*hoooray*

Issue #1697, opened 2019-09-02
Update our e2e dependencies


nicojs:
We should update all our e2e tests. The build reveals a lot of security warnings. Note: we should keep them locked to ensure relyable builds.

```
> e2e-tests@0.0.0 install-local-dependencies /home/travis/build/stryker-mutator/stryker/e2e
> ts-node tasks/install-local-dependencies
starting installation of local dependencies
bootstrap /home/travis/build/stryker-mutator/stryker/e2e
ine-framework)l-reporter)ocha-framework)utator)a-runner)ocha-runner)ine-runner)utator)er-project, vue-javascript
[install-local] polymer-project installed
+ @stryker-mutator/wct-runner@2.0.2
+ @stryker-mutator/api@2.0.2
+ @stryker-mutator/core@2.0.2
+ @stryker-mutator/javascript-mutator@2.0.2
+ @stryker-mutator/util@2.0.2
added 67 packages from 98 contributors and audited 5893 packages in 40.603s
found 63 vulnerabilities (23 low, 1 moderate, 38 high, 1 critical)
  run `npm audit fix` to fix them, or `npm audit` for details
[install-local] e2e-tests installed
> core-js@2.6.9 postinstall /home/travis/build/stryker-mutator/stryker/e2e/node_modules/@babel/runtime-corejs2/node_modules/core-js
> node scripts/postinstall || echo "ignore"
+ @stryker-mutator/core@2.0.2
+ @stryker-mutator/babel-transpiler@2.0.2
+ @stryker-mutator/html-reporter@2.0.2
+ @stryker-mutator/jasmine-framework@2.0.2
+ @stryker-mutator/jasmine-runner@2.0.2
+ @stryker-mutator/karma-runner@2.0.2
+ @stryker-mutator/mocha-framework@2.0.2
+ grunt-stryker@2.0.2
+ @stryker-mutator/javascript-mutator@2.0.2
+ @stryker-mutator/api@2.0.2
+ @stryker-mutator/mocha-runner@2.0.2
+ @stryker-mutator/util@2.0.2
+ @stryker-mutator/vue-mutator@2.0.2
+ @stryker-mutator/webpack-transpiler@2.0.2
+ @stryker-mutator/typescript@2.0.2
added 107 packages from 108 contributors and audited 8508 packages in 42.079s
found 252 vulnerabilities (1 low, 1 moderate, 250 high)
  run `npm audit fix` to fix them, or `npm audit` for details
[...]

PR #1750, opened 2019-09-28
update e2e dependencies

kmdrGroch:
fixes: #1697


simondel:
Does the Angular E2E test work on your machine? The initial test run fails on the build server. This is a different error than the inconsistency error, so I think it is caused by your change. There is a possibility that it's caused by the Typescript upgrade in the Angular project. Angular is quite strict on which versions they allow.

simondel:
Yes, I think Angular 8 only supports TypeScript 3.5.x

simondel:
I find it so strange (and annoying) that the build doesn't trigger D:

kmdrGroch:
I guess changing something which is working (travis) for something new, isnt always good -.-

simondel:
21:59:39 (4055) ERROR InitialTestExecutor One or more tests resulted in an error:
	Module not found: Error: Can't resolve 'core-js/es7/reflect' in '/home/runner/work/stryker/stryker/e2e/test/angular-project/.stryker-tmp/sandbox2803611/src

kmdrGroch:
I'll close it and work independently on each dependency

PR #1757, opened 2019-10-01
Update README to make the build always pass

simondel:


kmdrGroch:
@simondel I think it's a bad practice to do this. Isn't #1765 a better option? :/

simondel:
@kmdrGroch this was a temporary PR to test if the GitHub build would start from a fork PR (which didn't work for you).

I thought I had already closed it 🙀

PR #1762, opened 2019-10-03
feat(babel-transpiler): support object-style babel.config.js

Djaler:
The babel transpiler will now also work with babel.config.js files that export object instead of function.


nicojs:
Awesome! Just merged master inside this branch. Merging when green.

Sorry it took a while to get around to it 🙊

PR #1764, opened 2019-10-05
feat: add option to provide babelPlugins

kmdrGroch:
fixes: #1614 
fixes: #1806


kmdrGroch:
(i was hoping for github CI to remove all previous works ._.)

packages/api/package.json
```diff
@@ -37,6 +37,7 @@
     "tslib": "~1.10.0"
   },
   "devDependencies": {
+    "@babel/parser": "^7.5.5",
```
nicojs: I don't like to add `@babel/parser` to the `api` of Stryker. Stryker core and api are parser agnostic. I would like to keep it that way. Also seem my next remark

packages/api/src/core/MutatorDescriptor.ts
```diff
@@ -1,6 +1,9 @@
+import { ParserPlugin } from '@babel/parser';
+
 interface MutatorDescriptor {
   name: string;
   excludedMutations: string[];
+  babelPlugins: ParserPlugin[];
```
nicojs: Could we rename it to `plugins` (of type `string[]`)? That way we can also reuse it whenever typescript would allow plugins.

packages/core/src/di/factoryMethods.ts
```diff
@@ -36,3 +36,16 @@ optionsFactory.inject = tokens<[typeof coreTokens.configReadFromConfigFile, type
   coreTokens.configReadFromConfigFile,
   coreTokens.configEditorApplier
 );
+
+export function mutatorDescriptorFactory(options: StrykerOptions): MutatorDescriptor {
```
nicojs: I would suggest to change the type of `mutator` on the `StrykerOptions` to be of type `mutator: string | Partial<MutatorDescriptor>`.

After that change, you'll get a compile error here, stating that `Type 'Partial<MutatorDescriptor>' is not assignable to type 'MutatorDescriptor'`. You can fix that by providing default values:
```ts
return {
    plugins: [],
    excludedMutations: [],
    name: '',
    ...options.mutator
  };
```

packages/javascript-mutator/src/JavaScriptMutator.ts
```diff
@@ -8,14 +8,33 @@ import copy from './helpers/copy';
...
+      If it is not and `mutators` isn't as well, `mutatorDescriptor` is undefined
+      If it is not, but `mutators` is, `mutatorDescriptor` is an array
+    */
+    if (typeof mutatorDescriptor !== 'object' || Array.isArray(mutatorDescriptor)) {
+      this.mutatorDescriptor = {
+        babelPlugins: [],
+        excludedMutations: [],
+        name: ''
+      };
+    }
```
nicojs: This constructor implementation can be removed. We can rely on the type safety of TypeScript. When we receive a `MutatorDescriptor` here, we don't need to validate it again.

packages/javascript-mutator/src/JavaScriptMutator.ts
```diff
@@ -8,14 +8,33 @@ import copy from './helpers/copy';
...
+  }
 
   public mutate(inputFiles: File[]): Mutant[] {
     const mutants: Mutant[] = [];
+    const babelPlugins = this.mutatorDescriptor.babelPlugins || [];
 
     inputFiles.forEach(file => {
-      const ast = BabelHelper.parse(file.textContent);
+      const ast = BabelHelper.parse(file.textContent, babelPlugins);
```
nicojs: This will probably need a cast here. From a `string []` to `ParserPlugin[]`. 

We should check what happens when we supply a plugin that does not exist. What is the error message Stryker provides?

packages/test-helpers/src/TestInjector.ts
```diff
@@ -15,15 +15,20 @@ class TestInjector {
...
   public injector: Injector<OptionsContext> = rootInjector
     .provideValue(commonTokens.getLogger, this.provideLogger)
     .provideFactory(commonTokens.logger, this.provideLogger, Scope.Transient)
     .provideFactory(commonTokens.options, this.provideOptions, Scope.Transient)
-    .provideFactory(commonTokens.pluginResolver, this.providePluginResolver, Scope.Transient);
+    .provideFactory(commonTokens.pluginResolver, this.providePluginResolver, Scope.Transient)
+    .provideFactory(commonTokens.mutatorDescriptor, this.provideMutatorDescriptor, Scope.Transient);
 
   public reset() {
     this.options = factory.strykerOptions();
```
nicojs: I would expect that the mutator descriptor is reset here as well.

packages/api/package.json
kmdrGroch: cool

packages/javascript-mutator/src/JavaScriptMutator.ts
kmdrGroch: I think babel throw an error itself when can't find plugin

kmdrGroch:
Cool, looks great, it even passes tests :D

Issue #1667, opened 2019-08-08
repo (mostly docs) cleanup


kmdrGroch:
We should make some clean up repository, and possibly update documentation a bit.

Cleanup:
- `docs` folder should be removed if it is not needed.

Documentation:
there is an issue in https://github.com/stryker-mutator/stryker/tree/master/packages/util
since `download` and `node` badges are not displayed as intended

Updating `main` README / somehow merging with `core` README. Since they are +- 1 to 1.
For example:
```
Getting started
For the quickstart, see quickstart.

Usage
For details on the stryker run command and others, see stryker's readme

Configuration
See the stryker's package readme for all config options.

Supported mutators
See our website for the list of currently supported mutators.
```
This part can be easy replaced by `core` README. Some tweaks might be needed tho

We should also consider making 1 unified README for stryker project instead of making tens of them (or at least reduce the amount).

Alternative option to the previous one is to make `main` README a bunch of links (like it is now) to our website, and making better docs on website (including all READMEs [or at least the most important parts]). This unfortunately will again increase number of unnesessary files.

gugacavalieri:
Hey Guys!

I just took a quick look at the repo and I think all the docs moved to the handbook repo, am I right?
(https://github.com/stryker-mutator/stryker-handbook#readme)

And the `docs` folder in the main repo hasn't been updated in two years.

How are you handling documentation? Maybe I could give a hand with that :smile:

kmdrGroch:
thanks, @gugacavalieri for initiative! I would like @nicojs or @simondel to opinion that too, but for now, I think adding documentation for #1731 and #1623 would be nice :3 (You can PR to my repo so I'll add it). Also it's a bit hard for me to tell where exactly docs are. Since half of them are in handbook like you said, but also a big part of them is on the website https://github.com/stryker-mutator/stryker-mutator.github.io. So it's a bit messy in my opinion, but I can't do anything without their permission :/ (any idea, of course, is good!)

gugacavalieri:
Cool !

I will take a look at the docs then and check where the main part is also looking for some typos or grammar mistakes :)

It's feels kind of messy right now but I think we might be able to organize it!

Let's see what @nicojs and @simondel think as well

kmdrGroch:
@gugacavalieri My main point is to change the amount of docs. We have docs in 3-4 different places and so it's messy and hard to maintain (just look on main repo doc)
```
Introduction
[...]

PR #1768, opened 2019-10-06
docs: fixing typos in core/readme

gugacavalieri:
# Changelog: 
* fixing fixing typos in core/readme

First improvement for #1667


kmdrGroch:
is `whether` enough? I mean, it means the same thing as `if` am I correct?

gugacavalieri:
@kmdrGroch actually a [grammar checker](https://www.online-spellcheck.com/) suggested this change :)

I think it makes sense. Do you think it sounds weird ?

kmdrGroch:
not at all. I mean, it sounds better than `whether or not` since it's quite obvious that "if there is `whether` there is `not`"

Issue #1315, opened 2019-01-09
Use `oneFailurePerSpec` option in the stryker-jasmine-runner


nicojs:
When running our jasmine integration tests we see this deprecation warning:

> Setting throwOnExpectationFailure directly on Env is deprecated, please use the oneFailurePerSpec option in `configure`

We should address this.

enyfour5:
Hi @nicojs is this issue beginner friendly? If so, I would like to help out. Can you please point me in the right direction as to where to find the specific issue and what needs to be done? I'm already reading your contribution guidelines so I can get started. Thanks!

dhuang612:
Hi,

I would like to work on this

simondel:
Hi @dhuang612, welcome! If you need any help, please let me or @nicojs know :)

dhuang612:
created PR #1775

dhuang612:
I updated my PR with some comments

dhuang612:
someone else can take this over, thanks

PR #1775, opened 2019-10-13
updated throwOnExpectionFailure to oneFailurePerSpec

dhuang612:
worked on issue #1315
changed instances of throwOnExpectationFailure to use the newer call of 
oneFailurePerSpec

Please let me know if anything else needs to be changed

### DIFF
--- a/packages/jasmine-runner/src/JasmineTestRunner.ts
+++ b/packages/jasmine-runner/src/JasmineTestRunner.ts
@@ -55,7 +55,7 @@ export default class JasmineTestRunner implements TestRunner {
     const jasmine = new Jasmine({ projectBaseDir: process.cwd() });
     // The `loadConfigFile` will fallback on the default
     jasmine.loadConfigFile(this.jasmineConfigFile);
-    jasmine.stopSpecOnExpectationFailure(true);
+    jasmine.env.oneFailurePerSpec(false);
     jasmine.exit = () => { };
     jasmine.clearReporters();
     jasmine.randomizeTests(false);
--- a/packages/jasmine-runner/src/JasmineTestRunner.ts
+++ b/packages/jasmine-runner/src/JasmineTestRunner.ts
@@ -56,7 +56,6 @@ export default class JasmineTestRunner implements TestRunner {
     // The `loadConfigFile` will fallback on the default
     jasmine.loadConfigFile(this.jasmineConfigFile);
     jasmine.stopSpecOnExpectationFailure(true);
-    jasmine.env.throwOnExpectationFailure(true);
     jasmine.exit = () => { };
     jasmine.clearReporters();
     jasmine.randomizeTests(false);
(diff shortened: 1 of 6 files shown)


dhuang612:
currently getting the error
env.oneFailurePerSpec is not a function
I'll investigate into this more tomorrow

dhuang612:
I think part of the issue might be this?
https://github.com/sinonjs/sinon/issues/1429
```
var sandbox = sinonSandbox.create();
        console.log(sandbox.server) // => undefined
```

since we are using something similar to what is returning undefined

```
beforeEach(() => {
    sandbox = sinon.createSandbox();
    jasmineStub = sandbox.createStubInstance(Jasmine);
    jasmineStub.env =  {
      oneFailurePerSpec: sandbox.stub()
    };
```

kmdrGroch:
```
@stryker-mutator/jasmine-runner:   1) JasmineRunner integration
@stryker-mutator/jasmine-runner:        using the jasmine-init project
@stryker-mutator/jasmine-runner:          should run the specs:
@stryker-mutator/jasmine-runner:      TypeError: jasmine.env.oneFailurePerSpec is not a function
@stryker-mutator/jasmine-runner:       at JasmineTestRunner.createJasmineRunner (src/JasmineTestRunner.ts:2:68)
@stryker-mutator/jasmine-runner:       at JasmineTestRunner.run (src/JasmineTestRunner.js:2:284)
@stryker-mutator/jasmine-runner:       at Context.it (test/integration/JasmineRunner.it.spec.ts:57:35)
```

kmdrGroch:
isn't what we are looking for caled `stopOnSpecFailure`?
https://jasmine.github.io/api/3.5/Env.html

kmdrGroch:
https://jasmine.github.io/api/3.5/Env.html#configure

I guess this note 'Use the `oneFailurePerSpec` option with Env#configure' may be useful(?)

dhuang612:
I'm already doing that and getting an error that oneFailurePerSpec is not a function.
I'll continue reviewing this.

dhuang612:
@kmdrGroch 
Hey updated the code and it passed everything but the e2e which I believe is having issues right now?

simondel:
@nicojs as you were the one who created this issue in the first place, could you review it?

nicojs:
I don't have time to look in the logging of our e2e tests at the moment, but I suspect they will still log the warning, seeing as the root cause is not addressed:

https://github.com/stryker-mutator/stryker/blob/2bb7dd8b6f605eac29ba4af0463aa0cc4f395af2/packages/jasmine-runner/src/JasmineTestRunner.ts#L59

The idea is that our jasmine test runner plugin should set the `throwOnExpectationFailure` flag in the new (non-deprecated) way. We don't want to rely on people configuring it in there jasmine configuration.

dhuang612:
@nicojs 
I followed this from the Jasmine people on getting this removed
https://github.com/jasmine/jasmine-npm/issues/145
changing this line 
```
 jasmine.env.throwOnExpectationFailure(true); 
```

causes the fore mentioned errors of 
```
oneFailurePerSpec is not a function
```

Issue #1795, opened 2019-10-20
Remove dependency on lodash


nicojs:
**Is your maintenance request related to a problem? Please describe.**
We use lodash in a hand full of places, however we only do it for simple things. Like checking if a function is a function (instead of `value instanceof Function`), or checking if something is an array (instead of `Array.isArray(value)`). We should be able to remove the dependency.

**Describe the solution you'd like**
Remove all usages of `import * as _ fro lodash` and instead use current ES features.

**Additional context**
We've used lodash before, because back in the day of es5, it was a valid choice. Since then, the JS ecosystem has evolved. Lodash is a pretty big dependency. Let's remove it.

MagicCurlyHair:
Hi @nicojs 
I can take this issue.

kmdrGroch:
@nicojs I have noticed a bug in this template. It doesn't create a label `🔧 Maintenance` and doesn't assign me by default :<

nicojs:
Maybe because I assigned labels directly from "create issue" screen? Sorry about it, I thought it would still add the default label.

nicojs:
Closed with #1799

PR #1799, opened 2019-10-22
refactor: remove lodash dependency (resolves issue #1795)

D-Ig:
resolves issue #1795
Hi, I've removed lodash dependency with native lang features where considered reasonable, elsewise replaced whole lodash library dependency for dependencies on selective single functions.


nicojs:
Awesome! Thanks!

Issue #1653, opened 2019-07-21
React `propTypes` mutation.


rodoabad:
**Question**

When we're writing  React component, we typically write `propTypes` validations so we know that the props we're passing to a component is of the correct type.

Stryker is mutating these though even though they are for development purposes.
<img width="503" alt="Screen Shot 2019-07-21 at 6 41 29 PM" src="https://user-images.githubusercontent.com/1240667/61598744-8c3b4380-abe7-11e9-8a8c-f6f780d611e4.png">

For those who have actually prevented Stryker from mutating these, what did you do?

kmdrGroch:
For now i guess this is a bug. I don't see workaround. We might need to add some option to exclude given line or sth like this, the way it is done ex. in tslint or eslint

simondel:
Hi! Right now this is not possible to exclude. We have a general feature request for this: #1472 
As there is a general issue open, I'm closing this issue for now.

itsmeraji:
Hi, our team is interested in using Stryker to get the mutation test results, but almost all the files get this propTypes mutated and it brings the mutation test score down. We would like to know is there a possibility to exclude propTypes from getting mutated in Stryker. Is there any progress made on this?

kmdrGroch:
@itsmeraji I'll start working on #1472 very soon. It should be done until the end of this month I hope

itsmeraji:
@kmdrGroch, Thank you for the reply. But will it be possible to add some configuration in Stryker.conf and disable proptypes mutation in all files automatically, based on the configuration ? If we have to add ignore comments, we ve to add ignore comment for proptypes in all the files..

kmdrGroch:
@itsmeraji, for now, we are only planning to add the comments-based exclusion. In the future, we could write some plugin for it, but no hopes right now.

itsmeraji:
@kmdrGroch, is there any update on this feature? I see PR #1810 created for this.

kmdrGroch:
It is still being worked on. Unfortunately this issue is quite complex so it takes time :<

itsmeraji:
[...]

Issue #1470, opened 2019-03-26
Mutants surviving because it's trying to mutate logger.info


shaunswales:
### Summary

Mutants surviving because it's trying to mutate logger.info

When Stryker run's it is mutating the logger.info output which is for traceback, not actually affect the test.

<img width="633" alt="Screenshot 2019-03-26 at 09 42 28" src="https://user-images.githubusercontent.com/17008391/54987019-ae131980-4fab-11e9-9f3d-e94f52b99f97.png">

<img width="693" alt="Screenshot 2019-03-26 at 09 44 19" src="https://user-images.githubusercontent.com/17008391/54987054-bf5c2600-4fab-11e9-91f8-6e3b33c5abb8.png">

Is there a way to make Stryker ignore these?

### Stryker config

```js
module.exports = function (config) {
  config.set({
    files: [
      'src/**/*',
      '!test/**/*',
    ],
    mutate: [
      'src/code1/**/*.js',
      'src/code2/**/*.js',
      '!src/common/**/*',
      '!src/**/*.test.js',
      '!src/**/*.json',
      '!src/**/*.ejs',
      '!src/**/__mocks__/*',
      '!test/**/*',
    ],
    timeoutMs: 10000,
    mutator: 'javascript',
    packageManager: 'npm',
    reporters: ['html', 'clear-text', 'progress', 'dashboard'],
    testRunner: 'jest',
    transpilers: [],
    coverageAnalysis: 'off',
    jest: { config: require('./.jestrc.unit.json'), enableFindRelatedTests: true },
    htmlReporter: { baseDir: 'reports/mutation' },
    thresholds: { high: 80, low: 50, break: 40 },
    allowConsoleColors: true,
  });
};
```

### Stryker environment

```
+-- @stryker-mutator/core@1.1.1
+-- @stryker-mutator/api@1.1.1
+-- @stryker-mutator/util@1.1.1
+-- @stryker-mutator/html-reporter@1.1.
+-- @stryker-mutator/javascript-mutator@1.1.1
+-- @stryker-mutator/jest-runner@1.1.1
```

```
+-- jest@24.5.0
```

### Test runner environment

```
# Test command
npm run stryker
```

```json
{
  "collectCoverage": true,
  "collectCoverageFrom": [
    "src/**/*.test.{js,jsx}",
    "!<rootDir>/node_modules/",
    "!src/**/*.local.{js,jsx}"
  ],
  "coverageDirectory": "coverage",
[...]

PR #1810, opened 2019-10-28
feat(excludedExpressions): typescriptMutator - add an option to exclude mutations by code expressions 

tsachis:
Add the config option `excludedExpressions` (string array) to the `mutator` option, to support filtering out code blocks that include specific code expressions.

stryker.conf.js example:
```
mutator: {
  name: 'typescript',
  excludedExpressions: [
    'logger.debug',
    'LoggerFactory.',
    '.propTypes = '
  ]
}
```

This option can help improve performance and reduce reports noise by allowing more control over the generated mutations.

Related issues: #1470, #1653, #1472

***Need more work on:***
*  reporting the excluded mutations
*  Supporting excludedExpressions for javascript mutator


kmdrGroch:
That looks pretty good! Good job <3

Djaler:
This can handle only one line sub string? What about regexp instead?

kmdrGroch:
@Djaler it's both a good and bad idea of using regexps. + It would make it easier to look for things you want to exclude. - Regexps are super slow, so imagine running regexp hundreds/thousands of times.

packages/api/src/core/MutatorDescriptor.ts
```diff
@@ -1,6 +1,7 @@
 interface MutatorDescriptor {
   name: string;
-  excludedMutations: string[];
+  excludedMutations?: string[];
```
simondel: Could you add comments explaining the difference between the two exclusions?

packages/typescript/src/TypescriptMutator.ts
```diff
@@ -41,9 +45,21 @@ export class TypescriptMutator {
   }
 }
 
-const shouldNodeBeSkipped = (node: ts.Node): boolean => {
+const shouldNodeBeSkipped = (node: ts.Node, sourceFile: ts.SourceFile, excludedExpressions: string[] = []): boolean => {
+  if (excludedExpressions.length > 0) {
+    const nodeFirstLine = getNodeFirstLine(node, sourceFile);
```
simondel: I don't think this setup will work.

Take the following Angular component for example:
```ts
@Component({
  selector: 'app-bank-account',
  inputs: ['bankName', 'id: account-id'],
  template: `
    Bank Name: {{ bankName }}
    Account Id: {{ id }}
  `
})
export class BankAccountComponent {
  bankName: string|null = null;
  id: string|null = null;
  normalizedBankName: string|null = null;
}
```

If I don't want to mutate any `@Component({})` decorators, I could use the following config:
```js
mutator: {
  name: 'typescript',
  excludedExpressions: [
    '@Component({'
  ]
}
```

However, I think this will still mutate every property inside of my `@Component` because they are on a different linenumber.

simondel:

PR #1827, opened 2019-11-06
remove misleading docs

kmdrGroch:
Since people normally look for documentation in `docs` folder, we should remove it because it's misleading.


simondel:
These files seem to have already been moved here: https://github.com/stryker-mutator/stryker-handbook/tree/master/stryker/technical-reference/diagrams

So I'm merging this PR

Issue #1749, opened 2019-09-27
Angular End-to-End tests are unstable


simondel:
The Angular End-to-End tests are unstable. The mutation score differs sometimes because one specific mutant has a different state (survived instead of killed).

The mutation that survives is this Block mutation:
https://github.com/stryker-mutator/stryker/blob/8a20a17d75ee15edacae530cdbd82aafc9f0319c/e2e/test/angular-project/src/app/app.component.ts#L3-L7

Which is **mutated** into:
```ts
@Component({ })
```

This should cause tests in [app.component.spec.ts](https://github.com/stryker-mutator/stryker/blob/master/e2e/test/angular-project/src/app/app.component.spec.ts) to fail because the component cannot be created. For some reason, sometimes this does not happen.

kmdrGroch:
Have you tried putting 
```js
{ 
   selector: 'app-root', 
   styleUrls: ['./app.component.css'], 
   templateUrl: './app.component.html' 
 }
```
into a variable? It could be caused coz of `@Component` (somehow). It should be always changed in variable tho :/

simondel:
@kmdrGroch The mutated code is default Angular code. This shouldn't cause an issue. We can't ask our users to adopt a different syntax. If this issue is due to the syntax (and not the build environment for example), it should be fixed in Stryker.

kmdrGroch:
I know, I am just curious about the origin of that issue. Is it because decorator or some parser can't find it or what

nicojs:
This might be a race condition in the Angular-cli / karma itself.

Writing mutation to disk -> run karma -> see result.

The way compilation errors are reported is by this line of code:


I'm updating the tests to pass in the mean time. This is causing too many red builds

kmdrGroch:
It might also be caused by coz obsolete versions of testing tools / angular. Let's if anything changes after #1832

simondel:
As we've implemented a workaround, I'm closing this issue for now.

PR #1831, opened 2019-11-07
test(e2e): allow flaky angular test for now

nicojs:
See #1749


nicojs:
@simondel Any idea why the grunt test is flaky? I've looked into it, but can't find a reason

nicojs:
I've set the wct tests timeout higher, as that also caused flaky tests in unit+integration.

simondel:
Unfortunately not. Although they almost never seem to fail (compared to the Angular test)

PR #1832, opened 2019-11-07
Update packages (to review)

kmdrGroch:
fixes: #1697


kmdrGroch:
Problem during updating:
- e2e/ts-node  - 'cannot find bind' during linking
- e2e/cross-env - 'cannot find cross-env/dist...' (file is in 'src' folder not 'dist')

kmdrGroch:
@nicojs this should be done soon. One more package left :)

kmdrGroch:
when vue-javascript push passes it is ready to go :)

kmdrGroch:
@nicojs can i change testing tools in vue-javascript to chai?

kmdrGroch:
If this test doesn't pass I will need help

nicojs:
Great! I'll take a look tomorrow. Thanks a lot!

kmdrGroch:
@nicojs I think it's done *after 6 hours of testing*. There is one place you should take a look (a timeout in a router. I guess it is normal behaviour right now, but you should check it by yourself).
After that tell me if everything is alright, I still have to remove logs from stryker (vue-javascript logs generation) and fix these tests if necessary

nicojs:
To anwser you're questions:

> ... angular ...

This e2e/test/angular-project is a simple angular project. Probably it's easiest to [create a new one from scratch using the angular cli ](https://angular.io/guide/setup-local#step-1-install-the-angular-cli) and copy over the files from the `src` directory to this new one, or you can use [https://update.angular.io/](https://update.angular.io/). The e2e tests are not used, it's simply there because an angular project comes with them. Feel free to remove it. Protractor was probably removed as a dependency to get faster installation.

> In jest-react package. Is jest even used? I see only mocha.

Jest is used in the tests of the app here: https://github.com/stryker-mutator/stryker/tree/master/e2e/test/jest-react/src/__tests__. We always use mocha in the `verify` directories. You can run the e2e test by simply running `npm test` in the e2e test directory. It will first run Stryker (which in turn will use jest). As a `posttest` script we run the verify tests, which read from the reports to see if everything worked out.

> In jest-react package. Is webpack used? In scripts, I see some requirements for webpack config file but I don't see any

I don't think so. This project was generated with [create-react-app](https://github.com/facebook/create-react-app) (CRA). Webpack is used by CRA to bundle for production, however that is not needed since we're only running the tests. So feel free to remove these kind of things, it saves maintenance and download time.

kmdrGroch:
@nicojs thanks for the answers. I'll cover them in separate PR tho. For now, I would like to have this one finished. Please take a look at this timeout getting in a router, and so we can polish last things and merge it :)

nicojs:
I'm pretty sure the vue-javascript timeout issue is related to #621. Karma is handling the router issue as an error, rather than a failed test. It started doing that in v2 as far as I can see. I'll update the test expectation, for now, we should fix #621 and update it back later.

nicojs:
Ok, I went on a bug finding spree this weekend (a wonderful weekend indeed 😕) and found out that the `vue-javascript` issue has to do with an issue in karma-webpack as well as an issue in karma. I've opened #1837 to track them both.

PR #1839, opened 2019-11-11
Update linting tools and add import order

kmdrGroch:
closes: #1835


kmdrGroch:
@nicojs Finally I have found order import rule!
And it imports as you wanted to:
```
- built in modules

- external modules

- internal modules
```

simondel:
Hi @kmdrGroch Could you look at the failing build?

PR #1842, opened 2019-11-12
fix(perf/angular-cli): upgrade to latest angular version and fix bugs

simondel:


simondel:
Awesome, now we just have to revert the build change

kmdrGroch:
Simon, could you btw please change all '^' into '~' in package.json?

simondel:
Yeah, I will. The `ng update` command must have done that. Nice catch

simondel:
@kmdrGroch Done!

kmdrGroch:
Thanks @simondel ! But could you please rerun tests, since angular is not passing performance tests(?), It probably is false check but to be sure

PR #1843, opened 2019-11-13
Fix linting2

kmdrGroch:


kmdrGroch:
this `import something = require('something')` syntax is really annoying. It is not sorted by `fix` :< it isn't really hard to solve manually tho, but still it's a pain. I hope everything passes now

kmdrGroch:
But it works now @simondel 😄

packages/babel-transpiler/test/helpers/projectLoader.ts
```diff
@@ -1,8 +1,9 @@
-import { File } from '@stryker-mutator/api/core';
 import * as fs from 'fs';
-import * as glob from 'glob';
 import * as path from 'path';
 
+import { File } from '@stryker-mutator/api/core';
+import * as glob from 'glob';
```
simondel: I don't really get this formatting.. why is glob here while path and fs are in a different batch?

packages/api/src/plugin/Plugins.ts
```diff
@@ -1,10 +1,12 @@
 import { InjectableClass, InjectableFunction, InjectionToken } from 'typed-inject';
+
 import { ConfigEditor } from '../../config';
 import { Mutator } from '../../mutant';
 import { Reporter } from '../../report';
 import { TestFramework } from '../../test_framework';
 import { TestRunner } from '../../test_runner';
 import { Transpiler } from '../../transpile';
+
 import { PluginContexts } from './Contexts';
```
simondel: Why are these paths in a different block than the relative paths above?

kmdrGroch:
The formatting is:
```
Built in modules

External modules

Index modules

Other internal modules```

packages/babel-transpiler/test/helpers/projectLoader.ts
kmdrGroch: Glob is external package while path and fs are "native node" packages

packages/api/src/plugin/Plugins.ts
kmdrGroch: The path is '../../' and './' so it's separated (I am not really able to configure autofix for each case, but I guess it isn't that big of a problem here :/ (since it still makes sense))

PR #1848, opened 2019-11-13
feat(.gitignore): add Stryker patterns to .gitignore file during initialization

### LINKED ISSUE #1683: As part of Init, add .styrker-tmp folders to gitignore

By default the stryker temp folders don't seems to be git ignored, this can be frustrating if one forgets to set it as git ignored. 

detect gitignore files in solution and append 

```bash
# stryker temp files
*.stryker-tmp
```
or use the console to Warn the User to gitignore the temp folder generated by Stryker as part of the `Styrker init` process.

1. manually change
2. remind user with console message

**Additional context**

simondel:
Fixes #1683


packages/core/src/initializer/GitignoreWriter.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const GITIGNORE_FILE = '.gitignore';
+
+export default class GitignoreWriter {
+  public static inject = tokens(initializerTokens.out);
+  constructor(private readonly out: typeof console.log) {}
+
+  public async addStrykerTempFolder() {
+    if (fsAsPromised.existsSync(GITIGNORE_FILE)) {
+      const strykerTempFolderSpecification = `${os.EOL}# stryker temp files${os.EOL}*.stryker-tmp${os.EOL}`;
```
nicojs: Please define ` '.stryker-tmp'` as a const to be reused (from `@stryker-mutator/api/config`), making it DRY

nicojs: Also: running stryker init multiple times will result in multiple rules being added to your .gitignore file. We can do a simple `if(!gitignoreContent.endswith(strykerTempFolderSpecification)` to prevent this. Not really necessary maybe, but nice none the less.

nicojs: Why `*.stryker-tmp` instead of `.stryker-tmp`?

packages/core/test/unit/initializer/StrykerInitializer.spec.ts
```diff
@@ -52,6 +53,7 @@ describe(StrykerInitializer.name, () => {
       .provideClass(initializerTokens.npmClient, NpmClient)
       .provideValue(initializerTokens.strykerPresets, presets)
       .provideClass(initializerTokens.configWriter, StrykerConfigWriter)
+      .provideClass(initializerTokens.gitignoreWriter, GitignoreWriter)
```
nicojs: I'm pretty sure this will result in the real one being included. Please use a mocked `GitingoreWriter` and test that it is called correctly.

PR #1853, opened 2019-11-13
feat: support json-file config

kmdrGroch:
featuring JSON file stryker config
It can be the first step to introduce schema for stryker (+ its nice feature I guess)


nicojs:
I think this is a great improvement. @simondel what do you think? It could also be the first step to allow *.json file.

PR #1859, opened 2019-11-15
maintenance(lint): adher to new linting rules

nicojs:


kmdrGroch:
@nicojs I guess prettier won't allow you to make it in one line since our limit is 150letters per row(?)

nicojs:
Yes, strange how the rules suddenly got stricter. I guess bugfixing in their part, and since we're not locking our versions, we get the latest and greatest each time.

Run this locally to reproduce:: `rm -rf node_modules packages/*/node_modules && npm i`

nicojs:
Whut???? No file changes? 😲

nicojs:
Abandon 🙈

kmdrGroch:
yea nico, there were some strange caches in package.json file. I had to reinstall whole stryker coz i had some problems with doubling references. *happens*

PR #1874, opened 2019-11-21
nit: remove trailing spaces from .github/workflows/*.yml

brodybits:


nicojs:
@brodybits did you use a linting tool for this? If so it might be worth adding it to our linting CI/CD phase so we can prevent this from happening again

PR #1880, opened 2019-11-22
Fix default value of `coverageAnalysis` option

Userbit:


kmdrGroch:
Actually, the default value is `off`. 
https://github.com/stryker-mutator/stryker/blob/master/packages/api/src/config/Config.ts#L30

simondel:
Yes, that's right. The default value is `off`.

I'll close this PR for now. I'm wrong about the default value, please make a new PR with a test that proves that `perTest` is actually the default.

kmdrGroch:
@simondel 
https://github.com/stryker-mutator/stryker/blob/master/packages/api/src/config/Config.ts#L30
^
It should be default :/

PR #1881, opened 2019-11-23
Remove the unneeded @types/execa devDependency

ankon:
This fixes a warning when running `npm install`:
~~~~
npm WARN deprecated @types/execa@2.0.0: This is a stub types definition. execa provides its own type definitions, so you do not need this installed.
~~~~


kmdrGroch:
nice catch!

nicojs:
Great, thanks!

PR #1887, opened 2019-11-24
Delete wrong text about that perTest is by default

Userbit:
In fact, `off` is the default value of `coverageAnalysis` option


simondel:
Thanks!

PR #1891, opened 2019-11-25
[RFC ...] additional value mutations

brodybits:
- add `null` value mutation for array, object, and string literal values for JavaScript & TypeScript
- add numeric value mutations: `0`, `1`, `-1`, `Infinity`, `-Infinity`, and `NaN` for JavaScript only (for now)
- update the mutated string value to: `**Stryker was here!**`
- include the mutated string value of `**Stryker was here!**` for all string literals, not just empty strings
- add incremented & decremented numeric value mutations
- additional refactoring & cleanup

Open question:

- [ ] I wonder if some of these additional mutations should be moved into new mutators and if they should go behind any new options.

General rationale:

- I do think these changes could help users avoid the trap of the "billion dollar mistake" of using null values
- I have seen cases, mostly related to numeric data values in SQL databases, where the proposed numeric value mutations could help identify some more desired test cases

The numeric value mutations have already helped me find some more tests needed in a private PGP utility that I am not yet ready to publish. Here are some results I found on a couple of public repos:

On https://github.com/jonschlinkert/remove-bom-buffer which seems to be used by Gulp:

- report on Stryker 2.4.0, without this proposal (jonschlinkert/remove-bom-buffer#5): https://chrisbrody.com/remove-bom-buffer-stryker-report/index.html
- report with this proposal: https://chrisbrody.com/remove-bom-buffer-stryker-avm-001/index.html

With this proposal, we can see some more surviving mutants related to a string slice function call.

On https://github.com/ericelliott/riteway (<https://github.com/ericelliott/riteway/pull/193>):

- without this proposal: https://chrisbrody.com/riteway-stryker-report-001/index.html
- with this proposal: https://chrisbrody.com/riteway-stryker-avm-wip001/index.html

Here are a couple more mutants that this proposal identifies on riteway:

```diff
index a68eb08..473d508 100644
--- a/source/riteway.js
+++ b/source/riteway.js
@@ -12,7 +12,7 @@ const withRiteway = TestFunction => test => {
     should = '',
     actual = undefined,
     expected = undefined
-  } = {}) => {
+  } = null) => {
     test.same(
       actual, expected,
       `Given ${given}: should ${should}`
@@ -51,7 +51,7 @@ const createStream = tape.createStream.bind(tape);
  * @param {object} [obj] The object whose keys you wish to count.
  * @returns {number}
  */
-const countKeys = (obj = {}) => Object.keys(obj).length;
+const countKeys = (obj = null) => Object.keys(obj).length;
 
 export default describe;
 export { describe, Try, createStream, countKeys };
```

I have verified that the mocha test continues to pass on JavaScript, TypeScript, and Vue mutators.

TODO items:

- [ ] When I pushed my work to GitHub, it seemed to report some e2e test failures. It looks like there is a Stryker run as well. I have not been able to run e2e test on my mac or a Ubuntu node on the cloud. Any pointers would be much appreciated.
- [ ] If this project is ready to accept the proposed numeric value mutations, I would be happy to try adding them for TypeScript as well.

### DIFF
--- a/packages/javascript-mutator/src/helpers/NodeGenerator.ts
+++ b/packages/javascript-mutator/src/helpers/NodeGenerator.ts
@@ -5,6 +5,10 @@ export class NodeGenerator {
     return NodeGenerator.createAnyLiteralValueNode(originalNode, 'BooleanLiteral', value) as types.BooleanLiteral;
   }
 
+  public static createStringLiteralNode(originalNode: types.Node, value: string): types.StringLiteral {
+    return NodeGenerator.createAnyLiteralValueNode(originalNode, 'StringLiteral', value) as types.StringLiteral;
+  }
+
   public static createIdentifierNode(originalNode: types.Node, name: string): types.Identifier {
     return NodeGenerator.createMutatedNode(originalNode, 'Identifier', { name }) as types.Identifier;
   }
--- a/packages/javascript-mutator/src/helpers/NodeGenerator.ts
+++ b/packages/javascript-mutator/src/helpers/NodeGenerator.ts
@@ -5,6 +5,10 @@ export class NodeGenerator {
     return NodeGenerator.createAnyLiteralValueNode(originalNode, 'BooleanLiteral', value) as types.BooleanLiteral;
   }
 
+  public static createIdentifierNode(originalNode: types.Node, name: string): types.Identifier {
+    return NodeGenerator.createMutatedNode(originalNode, 'Identifier', { name }) as types.Identifier;
+  }
+
   public static createAnyLiteralValueNode(originalNode: types.Node, type: string, value: any): types.Node {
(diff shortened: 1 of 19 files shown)


kmdrGroch:
I don't think it's a necessity of creating multiple values like this. It is quite repetetive...
Instead of it, you could think of how to fix some of halting problems

brodybits:
Agreed that there are probably more values than are really needed, but I would not consider these to be 100% equivalent mutants. As I already said in the description:

- I did discover some more surviving mutants in the remove-bom-buffer package
- and I did discover a couple of interesting mutants of replacing `{}` with null in the riteway package

I hope the Stryker project will at least consider a subset of the additional mutations that I proposed here. I hope I don't have to keep them all in my own fork somewhere.

brodybits:
Closing for now, due to apparent lack of interest. I guess I will just have to experiment in my own fork somewhere, for now.

PR #1898, opened 2019-11-27
JavaScript mutator improvements

brodybits:
* JavaScript mutator code improvements
  - use raw string mutations *internally* whenever possible
  - generate some of the mutations as AST nodes from the Babel types API
  - use some ternaries
  - remove some intermediate const declarations no longer needed
  - remove a couple of internal utility functions no longer needed
* add Arrow Function Mutator for JavaScript (already supported by the existing TypeScript mutator)
* remove `lodash.clonedeep` from dependencies in packages/javascript-mutator/package.json

These improvements should help reduce the amount of consistency between the JavaScript and TypeScript mutators that I raised in #1893.

/cc @kmdrGroch

Issue #1893, opened 2019-11-26
[rfc idea] unify JavaScript & TypeScript mutators


brodybits:
While working on the additional mutations I proposed in PR #1891, I started wondering if it might be practical to consider combine the JavaScript & TypeScript mutators together.

I can think of a couple of possible approaches:

- Prettier is able to have a unified "estree" format from Babel and from TypeScript using `@typescript-eslint/typescript-estree` package. Prettier is also able to use Babel to parse TypeScript, with some known limitations at this point, as proposed in prettier/prettier#6400.
- It would also be possible to use the TypeScript parser to parse both JavaScript and TypeScript, an idea discussed in prettier/prettier#6888.

Another thing that can probably help some parsing beginners like myself is: https://astexplorer.net/

P.S. I can think of some important points that should be summarized here in the description as well:

- My idea to use `@typescript-eslint/typescript-estree` is there to help bridge the differences in syntax tree output from Babel and `typescript` parsers, as I explained in multiple comments below.
- I am now thinking that the parser could be moved into its own pluggable stage
- The next challenge is how to deal with things like JSX, TSX, Vue with JavaScript, and Vue with TypeScript. (Prettier seems to already have a solution to deal with these formats.)

Please, if anything is not clear in this descriptions or in the comments below, ask.

brodybits:
From a quick look, I noticed that the arrow function mutator only exists on the TypeScript mutator, also seems to be missing here: <https://github.com/stryker-mutator/stryker-handbook/blob/master/mutator-types.md>

I would also very likely favor just using the TypeScript parser to parse both JavaScript and TypeScript. When working on PR #1891, the TypeScript mutators seemed to be much easier to write and update.

kmdrGroch:
It would be really nice feature I guess. People could use Babel + typescript and everything within one configuration <3 the problem is Vue xD it has its own mutator :x

brodybits:
> People could use Babel + typescript and everything within one configuration

Yes

https://iamturns.com/typescript-babel/

The one thing is that the Babel ESTree mutators for JavaScript do not seem to be as clean or easy to write as the TypeScript mutators. I would like to take a look at this at some point.

> the problem is Vue xD it has its own mutator

Right. I suspect we *should* be able to switch from using `vue-template-compiler` to putting Vue sources through the Babel pipeline.

kmdrGroch:
[...]

PR #1912, opened 2019-12-04
feat: exclude mutations (inProgress)

kmdrGroch:
fixes: #1472
closes: #1810

### DIFF
--- a/packages/api/src/core/MutatorDescriptor.ts
+++ b/packages/api/src/core/MutatorDescriptor.ts
@@ -1,6 +1,7 @@
 interface MutatorDescriptor {
   name: string;
   excludedMutations: string[];
+  excludedExpressions: string[];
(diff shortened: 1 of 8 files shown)


brodybits:
FYI I think there would be a conflict between the updates in this PR so far and some of the JavaScript mutator improvements in PR #1898. I wonder if we should consider breaking the functionality of JavaScriptMutator.mutate() a little more smartly in PR #1898, or revisit some time later.

kmdrGroch:
Well, we can always see later :/

For now, I think we should focus on this PR though since this feature is commonly required by people :/

What's left to do is:
- add an exclusion for a specific mutation in javascript mutator
- add these features to vue mutator

Things which should be done soon:
- improve reporters so these mutations will be called "excluded" or sth like this.

simondel:

Issue #1435, opened 2019-03-01
Clarity on projectType=react


coleturner:
### Summary

Hi! I'm getting started with this project and I wanted to share some confusion I had around `projectType=react`.

I think it's worth clarifying that `react` here is create-react-app and not general purpose React. During the installation I selected this thinking it came with specific Babel transforms or mutants. Then I ran into an error about missing `react-scripts/` which I was able to recognize. There's not much in the way of issues for this particular error however and that lead me to create this issue.

Would you consider expanding the terminology (particularly for installation) to clarify that it's create-react-app specifically?

simondel:
Thanks for opening this issue!

Do you mean the setting in the Jest runner as described in its README:
https://www.npmjs.com/package/@stryker-mutator/jest-runner#configuring-jest

Or do you mean the 'react' option in the preset configurations? (which also sets the jest-runner to 'react'):
![image](https://user-images.githubusercontent.com/4221997/53678415-13caf900-3cbf-11e9-95e1-e7cc9ddf9d0d.png)

coleturner:
Hey @simondel thanks for replying! That is the correct menu I selected though I think it might be worth keeping them both aligned. I know CRA is ubiquitous but I worry for those who have never used it and wouldn't recognize the error :)

simondel:
Well, the preset is actually for `create-react-app` projects so I think it's ok to rename that to `create-react-app`.

nicojs:
We could also simply improve the documentation in both places. Instead of another deprecation 😰

coleturner:
Good point @nicojs I was worried about the major semver change, I think even parenthesis with (create-react-app) would suffice. For the JS configuration files it's a little different but I think a JS comment would suffice too.

skyboyer:
it works for my non-CRA application but I had to put 
```
      projectType: 'custom',
```
and remove `config` entry from `stryker.conf.js` to make it run without errors

MrFix93:
I can confirm that the fix from @skyboyer works for non-'create-react-app' react projects. I think the suggestion from @simondel to rename the option to `create-react-app` in the cli, should be sufficient to inform people this only works for these kinds op react projects.

nicojs:
Ok, sounds good. We're welcoming PR's. Please implement it without breaking changes (log a deprecation warning accordingly).

kmdrGroch:
Is it still a valid issue? There were quite a lot of changes and also some deprecations and removals (especially in stryker 4)

[...]

Issue #1362, opened 2019-02-06
Support react-scripts-ts?


nicojs:
It seems that react-scripts-ts is no longer needed for typescript enabled react applications. More info here: https://vincenttunru.com/migrate-create-react-app-typescript-to-create-react-app/

Running jest 24 results in a failing test on windows: 

```
  Integration test for Strykers Jest runner
No tests found, exiting with code 1
Run with `--passWithNoTests` to exit with code 0
In C:\z\github\stryker-mutator\stryker\packages\stryker-jest-runner\testResources\reactTsProject
  9 files checked.
  testMatch: C:/z/github/stryker-mutator/stryker/packages/stryker-jest-runner/testResources/reactTsProject/src/**/__tests__/**/*.(j|t)s?(x),C:/z/github/stryker-mutator/stryker/packages/stryker-jest-runner/testResources/reactTsProject/src/**\?(*.)(spec|test).(j|t)s?(x) - 0 matches
  testPathIgnorePatterns: \\node_modules\\ - 9 matches
  testRegex:  - 9 matches
Pattern:  - 0 matches
```

I've created a small reproduction repo here: https://github.com/nicojs/jest-run-cli-windows

I have no clue why this happens, but this is the case since jest 24.

@mthmulders do you know what is the way to move forward here? Should we remove react-ts project type support? Or warn that it only works for 24? Is react-scripts-ts support being dropped in favor of react-scripts?

simondel:
Since nothing has happened on this for about a year and no one has upvoted it. I'm closing this issue.

PR #1978, opened 2020-01-21
feat(cli): change react to create-react-app

MrFix93:
As reported in issue  #1435, the react preset is only applicable to a react project set up by create-react-app. It causes some confusion whereas the error (unable to locate package react-scripts) isn't that useful either. Therefore, I propose we change the projectType to react-create-app as it resembles the applicability of this preset. 

For backwards compatibility, "react" and "react-ts" can still be used, however, it does log a deprecation warning. 

ps. First PR 🚀

### DIFF
--- a/packages/jest-runner/src/JestConfigEditor.ts
+++ b/packages/jest-runner/src/JestConfigEditor.ts
@@ -1,5 +1,7 @@
 import { Config, ConfigEditor } from '@stryker-mutator/api/config';
+import { Logger } from '@stryker-mutator/api/logging';
 import jest from 'jest';
+import { commonTokens, tokens } from '@stryker-mutator/api/plugin';
 
 import CustomJestConfigLoader from './configLoaders/CustomJestConfigLoader';
 import JestConfigLoader from './configLoaders/JestConfigLoader';
@@ -10,6 +12,10 @@ import JEST_OVERRIDE_OPTIONS from './jestOverrideOptions';
 const DEFAULT_PROJECT_NAME = 'custom';
 
 export default class JestConfigEditor implements ConfigEditor {
+  public static inject = tokens(commonTokens.logger);
+
+  constructor(private readonly log: Logger) {}
+
   public edit(strykerConfig: Config): void {
     // If there is no Jest property on the Stryker config create it
     strykerConfig.jest = strykerConfig.jest || {};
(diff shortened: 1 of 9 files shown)


packages/jest-runner/src/JestConfigEditor.ts
```diff
@@ -28,9 +34,19 @@ export default class JestConfigEditor implements ConfigEditor {
     switch (projectType.toLowerCase()) {
       case DEFAULT_PROJECT_NAME:
         return new CustomJestConfigLoader(process.cwd());
+      case 'create-react-app':
+        return new ReactScriptsJestConfigLoader(process.cwd());
+      case 'create-react-app-ts':
+        return new ReactScriptsTSJestConfigLoader(process.cwd());
       case 'react':
+        this.log.warn(
```
mthmulders: Is this a place where a fall-through would make sense? So the `react` case only does logging and then "falls through" to the `create-react-app` case? Same for `react-ts` and `create-react-app-ts` of course.

mthmulders:
Lookin' good!

Now that create-react-app also supports TypeScript, do the defaults of the `create-react-app` case still make sense? We could be seeing projects that use `create-react-app` (not `create-react-app-ts`) **and** use TypeScript at the same time.

packages/jest-runner/src/JestConfigEditor.ts
MrFix93: Have thought of that, however, I would argue that this is more readable as you don't have to think about which statements do fall-through..

mthmulders: Fair enough. Readability is very important!

nicojs:
@mthmulders good catch! 

See https://github.com/stryker-mutator/stryker/issues/1362 about CRA with TypeScript. This was still an open question from my end. 

@mthmulders or @MrFix93 do you know what the correct defaults are for different CRA setups? For example, our vue-cli has some more questions is asks before writing the config. We could do the same thing for CRA.

mthmulders:
> @mthmulders or @MrFix93 do you know what the correct defaults are for different CRA setups? For example, our vue-cli has some more questions is asks before writing the config. We could do the same thing for CRA.

I don't really know. I think the biggest difference is whether one uses TypeScript. If so, we should have Stryker use the TypeScript mutators instead of the JavaScript ones. Other than that none that I'm aware of.

MrFix93:
Good question @mthmulders. While I was implementing this, I questioned whether the react-ts is used at all. In addition, @nicojs, it looks like the only difference in both presets (jsx and tsx React) is that tsx uses `react-scripts-ts` instead of `react-scripts`. And like the source you mentioned in #1362 , it is all bundled in `react-scripts`. The difference is in Stryker, as it requires the typescript mutator instead of the javascript mutator.

So. We do need to know if the user wants tsx support, as it requires different dependencies (the typescript mutator). However, the create-react-app-ts preset in the jest-runner might be abandoned. Any views on this? 

In addition, it doesn't look like jest-runner supports Typescript and Javascript in a single project. Is that correct?

nicojs:
> However, the create-react-app-ts preset in the jest-runner might be abandoned. Any views on this?

That's an interesting idea. Let's do that in a future PR. It's smart to keep features in separate PR's.

> The difference is in Stryker, as it requires the typescript mutator instead of the javascript mutator.

In that case I would expect there to be 1 question after you choose for CRA. Namely: what source code language do you use. @MrFix93 do you mind adding that?

Thanks again for your work so far!

nicojs:
@MrFix93  If you want, we can also merge this PR and add the option in a separate PR. It's up to you

MrFix93:
> In that case I would expect there to be 1 question after you choose for CRA. Namely: what source code language do you use. @MrFix93 do you mind adding that?

There is already a question after the preset: 
https://github.com/stryker-mutator/stryker/blob/97eed0853195089eb8d70e4ded9562200df5207d/packages/core/src/initializer/presets/ReactPreset.ts#L41

That will suffice, right? 

If yes, let's merge 🚀

nicojs:
Ah yes, damn we we're smart in the past 😅

nicojs:
@MrFix93 congrats on your first PR 🎉

PR #1982, opened 2020-01-23
feat: Collecting anonymous user statistics

Jelle28:
Hi 🙋‍♂️
I've been working on an implementation to collect anonymous user statistics, which are being visualized giving insights on how Stryker is being used. 

When executing `stryker init` there's a new option at which a user can allow that Stryker will or will not collect and send the statistics. When this is allowed and Stryker is running, user statistics are being collected within the `Stryker` class and send to an Azure Function in a new class `Statistics`. Within Azure the statistics are processed and send to a database in MongoDB Atlas. I've been able to create these two dashboards (note that this is test data):

- https://charts.mongodb.com/charts-project-0-yurnw/public/dashboards/9fe96c7a-d427-4bd9-a677-fdc0356f8732
- https://charts.mongodb.com/charts-project-0-yurnw/public/dashboards/5f30c2da-172c-4991-bd2f-183e94725689

The code is not finished and needs some improvements, which includes (atleast) these two:
- Filtering the errormessage that's being send is a string that can occasionally include the user's folder tree at which the error occurs.
- The Statistics Class could be an Singleton object, which adds the option to add statistics any where within the process.

I hope you can use this further on.


kmdrGroch:
Hey! Thanks for this PR. it looks good at the first glance, but I think, "no" value should be a default one. we shouldn't enforce people to send data unless they want us to.

hugo-vrijswijk:
@Jelle28 Based on some discussion, we've decided the following:

- Your research report is the most valuable for us. It'll be used as the main document to decide on what statistics we want to collect and how to do it.
- The focus is still on opt-out collecting. We'll take another look at privacy for error messages and test frameworks, as both are quite dynamic.
- This PR will be closed for now. When we pick this up in the future we can use your code as a base to go on, but as is it's a little difficult to directly merge this (look at the merge conflicts already here, for example).

nicojs:
Thanks for the work you put in @Jelle28. We'll definitely work on this in 2020. Probably start smaller and build up in sever PR's

PR #1986, opened 2020-01-24
fix(Jest): show lacking Jest 25 support in peerDependencies

simondel:


simondel:
mvn this could cause builds to fail if people automatically get this version due to the unmet peer dependencies.

PR #2001, opened 2020-02-01
deps(log4js): lock log4js to version 6.1.0

simondel:


simondel:
Fixed with #2002 since this PR proved that the patch was the issue.

PR #2045, opened 2020-02-14
feat(command-runner): allow func as command param

### LINKED ISSUE #2033: Feature: pass a function to the command runner
I have an Angular + Jest setup. Running `ng test` normally completes all of my 692 tests in 1-1.5 minutes.
Stryker, using the the jest-runner plugin, takes over 17 hours to run a full mutation test, which I think is far too long.
Using the default command runner (which runs every test for each mutation) would take more than a week straight according to Stryker’s estimation.

After a lot of research and experimentation, I found the that Jest’s --test-path-pattern CLI param will pass through Angular CLI to Jest, and using that allows me to run just one test file at a time.

To set that param to the correct file each time, I would like to be able to pass a function to the command runner config instead of just a static string. That function should accept RunOptions as a parameter.

The following code takes the path of the mutated file and adds `.spec` to it to get the associated test file:  

```
 const command = 'npx ng test --runInBand --runTestsByPath';

function fileNameToSpecRegex(mutatedFileName) {
  return mutatedFileName.replace(/^.*\/sandbox\d+\//, '').replace(/\.ts$/, '.spec.ts').replace(/\./g, '\\.');
}

…
    commandRunner: {
      command: ({ mutatedFileName }) => {
        return mutatedFileName
          ? `${command} --passWithNoTests --test-path-pattern="${fileNameToSpecRegex(mutatedFileName)}"`
          : command;
      }
    }
…
```

This may leave out other tests related to the mutant, meaning some mutants could survive the mutation test when they would actually be killed in a full unit test run. But time saved makes it worth it. Using the code above with a modified CommandTestRunner.js, the mutation test ran in 4.5 hours.

Getting the coverage output from `ng test` might help, but I doubt it’d do any better than the Jest plugin’s use of --findRelatedTests.

**Additional context**
I already have a change working in CommandTestRunner.ts. If you’d like this feature, I’ll write an associated unit test for it and create a PR. If a similar solution is already available, please let me know because I've been wracking my brain for different ways to speed up this mutation test.

jpsilva:
Fixes #2033

### DIFF
--- a/packages/core/src/test-runner/CommandTestRunner.ts
+++ b/packages/core/src/test-runner/CommandTestRunner.ts
@@ -2,14 +2,16 @@ import { exec } from 'child_process';
 import * as os from 'os';
 
 import { StrykerOptions } from '@stryker-mutator/api/core';
-import { RunResult, RunStatus, TestRunner, TestStatus } from '@stryker-mutator/api/test_runner';
+import { RunOptions, RunResult, RunStatus, TestRunner, TestStatus } from '@stryker-mutator/api/test_runner';
 import { errorToString } from '@stryker-mutator/util';
 
 import { kill } from '../utils/objectUtils';
 import Timer from '../utils/Timer';
 
+type CommandFunction = (runOptions: RunOptions) => string;
+
 export interface CommandRunnerSettings {
-  command: string;
+  command: string | CommandFunction;
 }
 
 /**
@@ -45,11 +47,12 @@ export default class CommandTestRunner implements TestRunner {
     );
   }
 
-  public run(): Promise<RunResult> {
+  public run(runOptions: RunOptions): Promise<RunResult> {
     return new Promise((res, rej) => {
+      const command = typeof this.settings.command === 'string' ? this.settings.command : this.settings.command(runOptions);
       const timer = new Timer();
       const output: Array<string | Buffer> = [];
-      const childProcess = exec(this.settings.command, { cwd: this.workingDir });
+      const childProcess = exec(command, { cwd: this.workingDir });
       childProcess.on('error', error => {
         kill(childProcess.pid)
           .then(() => handleResolve(errorResult(error)))
--- a/packages/core/src/test-runner/CommandTestRunner.ts
+++ b/packages/core/src/test-runner/CommandTestRunner.ts
@@ -8,10 +8,11 @@ import { errorToString } from '@stryker-mutator/util';
 import { kill } from '../utils/objectUtils';
 import Timer from '../utils/Timer';
 
-type CommandFunction = (runOptions: RunOptions) => string;
+export type CommandProvider = (runOptions: RunOptions) => string;
 
 export interface CommandRunnerSettings {
-  command: string | CommandFunction;
+  command?: string;
+  commandProviderFile?: string;
 }
 
 /**
@@ -47,9 +48,15 @@ export default class CommandTestRunner implements TestRunner {
     );
   }
 
-  public run(runOptions: RunOptions): Promise<RunResult> {
+  public async run(runOptions: RunOptions): Promise<RunResult> {
+    let command = this.settings.command as string;
+
+    if (this.settings.commandProviderFile) {
+      const commandProvider: CommandProvider = await import(this.settings.commandProviderFile);
+      command = commandProvider(runOptions);
+    }
+
     return new Promise((res, rej) => {
-      const command = typeof this.settings.command === 'string' ? this.settings.command : this.settings.command(runOptions);
       const timer = new Timer();
       const output: Array<string | Buffer> = [];
       const childProcess = exec(command, { cwd: this.workingDir });
--- a/packages/core/test/integration/command-test-runner/CommandTestRunner.it.spec.ts
+++ b/packages/core/test/integration/command-test-runner/CommandTestRunner.it.spec.ts
@@ -13,7 +13,7 @@ describe(`${CommandTestRunner.name} integration`, () => {
 
   it('should report test as successful', async () => {
     const sut = createSut();
-    const result = await sut.run();
+    const result = await sut.run({ timeout: 100 });
     expect(RunStatus[result.status]).eq(RunStatus[RunStatus.Complete]);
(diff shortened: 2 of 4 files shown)


packages/core/src/test-runner/CommandTestRunner.ts
```diff
@@ -2,14 +2,16 @@ import { exec } from 'child_process';
...
 
 import { StrykerOptions } from '@stryker-mutator/api/core';
-import { RunResult, RunStatus, TestRunner, TestStatus } from '@stryker-mutator/api/test_runner';
+import { RunOptions, RunResult, RunStatus, TestRunner, TestStatus } from '@stryker-mutator/api/test_runner';
 import { errorToString } from '@stryker-mutator/util';
 
 import { kill } from '../utils/objectUtils';
 import Timer from '../utils/Timer';
 
+type CommandFunction = (runOptions: RunOptions) => string;
```
nicojs: Please rename to `CommandProvider`. I think that communicates clearer it's intent.

packages/core/test/integration/command-test-runner/CommandTestRunner.it.spec.ts
```diff
@@ -13,7 +13,7 @@ describe(`${CommandTestRunner.name} integration`, () => {
 
   it('should report test as successful', async () => {
     const sut = createSut();
-    const result = await sut.run();
+    const result = await sut.run({ timeout: 100 });
```
nicojs: Writing it like this makes it seem that `timeout` is somehow significant. I took me a couple of seconds before I figured out you only need it here to satisfy the TypeScript compiler.

Please define a `const UNUSED_RUN_OPTIONS: RunOptions =  { timeout: 100 }` and use that instead of `{ timeout: 100 }`

packages/core/test/unit/test-runner/CommandTestRunner.spec.ts
```diff
@@ -37,6 +37,14 @@ describe(CommandTestRunner.name, () => {
       expect(childProcess.exec).calledWith('some other command');
     });
 
+    it('should allow other commands using function-based configuration', async () => {
+      const command = sinon.fake.returns('some functional command');
```
nicojs: Out of curiosity, why create a sinon `fake` here? I've only worked with `stub()`. What is the difference exactly?

jpsilva: I don't know what the difference is. I'm not familiar with sinon, so I had I to dig into sinon's docs when I saw that's what you're using. I can change it to stub().

packages/core/test/integration/command-test-runner/CommandTestRunner.it.spec.ts
jpsilva: Yup, definitely cleaner.

packages/core/src/test-runner/CommandTestRunner.ts
jpsilva: Makes sense

jpsilva:
I'll wait on @simondel's feedback, but I agree with all your suggestions. Obviously JSON schema can't allow function types because JSON doesn't allow functions. Plus, while I don't think you allow the Stryker conf to be a JSON file right now, preserving the option to do so in the future makes sense.

nicojs:
You can already define your Stryker config in a stryker.conf.json file, this feature was recently added. We still need to update the docs I realize now 🙄

brodybits:
Sorry if this is off-topic and with the worst possible timing. I would actually prefer conf.js file over configuration JSON since I have generally had an easier time with conf.js. My impression and experience is that editing and adapting conf.js files for tools such as Babel and Stryker is more flexible and less rigid than dealing with a JSON file such as package.json. Just my (US) $0.02 worth.

I wonder if it could be ideal to consider something like YAML next to JSON, like ESLint and cosmicconfig do.

kmdrGroch:
@brodybits the only reason for JSON is that IDEs can provide IntelliSense there with $schema property. I don't know if it is possible in JS files :/

nicojs:
Indeed off topic, but here is my €0,02. 

> I would actually prefer conf.js file over configuration JSON since I have generally had an easier time with conf.js. 

We'll _always keep support for `stryker.conf.js` files_. It is very useful, for example to distinguish between build server vs local development. This cannot be done with regular json. Just two minor changes in the foreseeable future: 
1. We will deprecate the current way of working (karma style function export) in favor of just exporting the config object (webpack/eslint config style)
1. We will be validating the config with JSON schema validation, so the (core) config should be compatible with that (although plugins can still allow arbitrary config like regexp or functions, like we do now, with the disadvantage of only working in `.js` files and not `.json` files)

> IDEs can provide IntelliSense there with $schema property, so you don't have to look for options in docs. I don't know if it is possible in JS files :/

Sure it can. With the wonders of typescript.

```js
/**
* @type {import('@stryker-mutator/api/core').StrykerOptions}
*/
const options = {
  // intellisense here
}
module.exports = options
```

> I wonder if it could be ideal to consider something like YAML next to JSON, like ESLint and cosmicconfig do.

Since JSON is compatible with YAML, and code editors use JSON schemas to provide intellisense for it, we might allow it in the future if there is enough demand for it. I'd rather have support for JSON with comments (JSONC) in the short term. That would also be supported in most editors.

Please any further discussions in (new) issues 😅

jpsilva:
I didn't want to wait any longer. I made the requested changed. The tests pass and it's working for me locally. My implementation and unit testing feels a little clunky to me, so if you have any suggestions for making it more elegant, let me know.

kmdrGroch:
@nicojs thanks for showing that. We should also add this schema support :)
My idea is that we can choose in cli if we want JS or JSON file. Then we will create one of them with schema support and some default settings. I wouldnt hurt JS lovers nor JSON lovers. Then YAML support? Sure, just dump it in and we can choose out of three :)

nicojs:
@jpsilva are you still interested in this feature? I think the recent changes we made to the jest runner should improve performance by quite a bit. 

I'll close this for now. Please open a new PR when you feel like you still need this feature 👍

PR #2048, opened 2020-02-15
feat(karma-runner): disable client.clearContext 

### LINKED ISSUE #2049: Race condition in karma can result in a mutant being tested as a 'runtimeError'
**Summary**

There is a race condition in karma that can result in a mutant being reported as a runtime error. The race condition is explained in https://github.com/karma-runner/karma/issues/3424

I've found the issue in a gh workflow (on linux). I haven't been able to reproduce it on windows. As you can see in the stack trace: clearly the `onbeforeunload` is triggered from karma code (see https://github.com/karma-runner/karma/issues/3424 for more info)

```
ERROR: 'Window reload!!', 'Error
2020-02-15T23:56:13.4716280Z     at window.onbeforeunload (http://localhost:9877/base/test/AddSpec.js?cc8af791b57fdffcc1474a86f80f049403495bde:21:40)
2020-02-15T23:56:13.4716758Z     at navigateContextTo (http://localhost:9877/karma.js:135:18)
2020-02-15T23:56:13.4717186Z     at clearContext (http://localhost:9877/karma.js:161:5)
2020-02-15T23:56:13.4717823Z     at http://localhost:9877/karma.js:250:9'
```

**Stryker config**

This happens in our e2e test suite sometimes with this karma config:

```js
{
    mutate: ['src/*.js'],
    mutator: 'javascript',
    testFramework: 'mocha',
    testRunner: 'karma',
    reporters: ['clear-text', 'html', 'event-recorder'],
    karma: {
      config: {
        frameworks: ['mocha', 'chai'],
        files: ['src/*.js', 'test/*.js']
      }
    },
    logLevel: 'trace',
    timeoutMS: 60000,
    maxConcurrentTestRunners: 2,
    coverageAnalysis: 'perTest'
  }
```

**Stryker environment**

See our e2e tests. Both karma-mocha and karma-jasmine

**A possible solution**

We can implement the workaround proposed in https://github.com/karma-runner/karma/issues/3424 by forcing `client.clearContext` to `false`

nicojs:
Disable client.clearContext to avoid a race condition in Karma where a mutant is unfairly reported as "RuntimeError".

Fixes #2049


nicojs:
I just thought of a side effect of disabling `client.clearContext`. 

When I disabled `client.clearContext` the integration tests for karma were failing. It had to do with an error in our integration tests that occur after a timeout of 200ms (see my change in `packages/karma-runner/testResources/sampleProject/test-mocha/AddSpec.js`). I found it strange that this error only now arises, but then I realized that it has to do with `client.clearContext`. Clearing the clientContext (loading `about:blank` in the iframe as soon as tests are done executing) ensures that no other errors occur when the tests are actually already done. I think this is the primary reason for karma to clear the client context.  

This is a dilemma for us. We either disable `context.clearContext` so Stryker works for people that didn't forget to write there `await` statements and clean up after themselves, or we don't disable it and Stryker works somewhat for everyone.

I still think disabling clearContext is worth it, at least, until https://github.com/karma-runner/karma/issues/3424 is fixed.

We could also try to patch it more on our side. Right now, we ignore the `exitCode` that Karma provides us after a run. Instead, we report a `RuntimeError` whenever an error event occurred. We could use the `exitCode` that karma provides in our logic (0 = succeeded, 1 = failed tests OR error occurred).

https://github.com/stryker-mutator/stryker/blob/9d1fcc17e3e8125d8aa9174e3092d4f9913cc656/packages/karma-runner/src/KarmaTestRunner.ts#L146-L148

@simondel what do you think about this?

PR #2084, opened 2020-03-09
Javascript - Ignore configured expressions

itsmeraji:
Please check this PR. This is an attempt to configure the statements to be removed from getting mutated.

### DIFF
--- a/e2e/test/jest-react/stryker.conf.js
+++ b/e2e/test/jest-react/stryker.conf.js
@@ -1,7 +1,10 @@
 module.exports = function (config) {
   config.set({
     testRunner: "jest",
-    mutator: "javascript",
+    mutator: {
+      name: "javascript",
+      excludedExpressions: ["propTypes", "defaultProps"]
+    },
     transpilers: [],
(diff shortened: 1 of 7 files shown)


itsmeraji:
@nicojs / @kmdrGroch , we've been waiting for ignoring propTypes and such from mutation in Stryker. This PR has the relevant changes. Please check and let me know if this will hold good.

simondel:

itsmeraji:
@simondel, Thanks for the response. When can we expect that useful parser API ? Our team wants to implement mutation testing but we re waiting for this for a long time.

simondel:
We'll be working on this or the coming months. What's keeping you from using mutation testing without ignoring statements? It will result in a lower score, but that will simply be your baseline.

itsmeraji:
We have a lot of applications and the developers feel that their testing score is less because there are a lot of things that shouldn't be really tested (like propTypes, PropTypes.oneof, logger.debug, etc), but mutation is done for everything. So we wanted to ignore the unwanted mutations before widely practicing it.

PR #2087, opened 2020-03-10
build(dev-deps): Downgrade eslint-plugin-import to 2.20.0

simondel:
Fixes eslint import order errors on Windows systems


nicojs:
Fixed with #2091

PR #2093, opened 2020-03-11
feat(Initializer): Initialize config file as JSON by default

### LINKED ISSUE #2000: Stryker-CLI - create stryker.conf.json file with $schema
Since we have pushed schema file for Stryker config, w could use it. My idea is to create JSON file config (which we support too ;) ). And add $schema property so intellisence will give help popups:
```json
{
   "$schema": "url_to_schema_from_schemastore/stryker",
  The rest of the stryker options
}
```

simondel:
Fixes #2000 

During initialization, the user is now prompted to ask if they want their config to be in a JSON format instead of js. Both formats now have intellisense as well.


packages/core/src/initializer/StrykerInquirer.ts
```diff
@@ -90,4 +90,17 @@ export class StrykerInquirer {
...
     return options.filter(_ => _.name === answers.packageManager)[0];
   }
+
+  public async promptJsonConfigType(): Promise<boolean> {
+    const json = 'JSON';
+
+    const answers = await inquirer.prompt<{ configType: string }>({
+      choices: [json, 'JavaScript'],
+      default: json,
+      message: 'What kind of config do you want?',
```
nicojs:
```suggestion
      message: 'What file type do you want for your config file?',
```

PR #2123, opened 2020-03-27
Validate config using JSON schema

nicojs:
This merge request is far too big! (sorry 🙄)

A number of improvements:

* **Api**: Deprecate `Config` in favor of `StrykerOptions`             
    * Made `Config` an `interface` now, for the time being, to be removed in next major version
* **Api**: Export the `StrykerOptions` JSON schema from the API
* **Core**: Validate `StrykerOptions` using JSON schema
* **Util**: Refactor: replace `filterEmpty` in favor of `notEmpty`          
* **Core**: Feature: hide stacktrace on validation error                  
* **Util**: Refactor: move `noopLogger` to utils (used in 3 places now)
* **Util**: Remove old ES5 work around in StrykerError so `instanceof StrykerError` works.
* **Core**: Use AJV for fill in the default options.

Please don't squash merge (rebase and merge is fine), so the changelog will contain all these things.


packages/core/src/config/validationErrors.ts
```diff
@@ -0,0 +1,123 @@
...
+  const params: TypeParams = {
+    type: typeErrors.map(error => (error.params as TypeParams).type).join(',')
+  };
+  return {
+    ...typeErrors[0],
+    params
+  };
+}
+
+function describeError(error: ErrorObject) {
```
simondel, line 86: Could you specify a return type here and add comments describing what this method does?

nicojs: Thanks. Did that AND documented the code A LOT more. It should be clear now what we're doing and why, if we would ever want to refactor that part of the code (I'm not terribly proud of it 🙄)

nicojs:
Note: Mutation testing report can be found here:

https://dashboard.stryker-mutator.io/reports/github.com/stryker-mutator/stryker/test/feat/validate-config-using-json-schema

nicojs:
I just rebased on master, please rebase and merge instead of squash merge, if you're ready.

nicojs:
@simondel I've found this package and tested it out:

https://github.com/atlassian/better-ajv-errors

It actually works great. Although it won't filter the errors, so it would only be a replacement for `describeError`. It also comes with a small overhead: https://bundlephobia.com/result?p=better-ajv-errors@0.6.7. 

Some examples:

image
image
image

Also, it will not represent the actual code of your Stryker options are written in plain JS format rather than your JSON format.

nicojs:
I think I'm falling in love with better-ajv-errors 😍

nicojs:
I've opened an issue to move all the ajv validation logic to better-ajv-errors, let's see what they come up with:

https://github.com/atlassian/better-ajv-errors/issues/76

PR #2158, opened 2020-04-22
feat(validation): add validation on plugin options

nicojs:
A Stryker plugin is now able to contribute to the validation of StrykerOptions. Export a JSON schema from your plugin with the name `strykerValidationSchema` and it will be added to the validation.

Also some housekeeping: 🏡

* StrykerOptions now support additional options as `[k: string]: unknown` instead of `[k: string: any`. This added about 100 compiler errors that I had to fix.
* We apparently had some hidden features, like options for the clear text reporter and event recorder reporter. They are now documented in the core schema

Example:

```
$ npm run stryker

13:17:48 (24856) INFO ConfigReader Using stryker.conf.js
13:17:49 (24856) ERROR OptionsValidator Config option "mochaOptions['async-only']" has the wrong type. It should be a boolean, but was a string.
13:17:49 (24856) ERROR OptionsValidator Config option "tsconfig" has the wrong type. It should be a object, but was a string.
13:17:49 (24856) ERROR StrykerCli Please correct these configuration errors and try again.
```

TODO

- [x] Add unit tests and integration tests
- [x] Add an e2e test for some wrong config
- [x] Remove additional logging from Ajv


e2e/helpers.ts
```diff
@@ -19,6 +19,10 @@ type WritableMetricsResult = {
   -readonly [K in keyof MetricsResult]: MetricsResult[K];
 };
 
+export function logFileContent(fileName = path.resolve('stryker.log')): Promise<string> {
```
simondel, line 22: This does not log anything :P It reads the file content

simondel: If this is meant to be read as `logfileContent` I would suggest `readLogFife`

packages/core/src/di/PluginLoader.ts
```diff
@@ -27,6 +32,10 @@ export class PluginLoader implements PluginResolver {
     });
   }
 
+  public resolveValidationSchemaContributions(): object[] {
```
simondel, line 35: Why don't you use a getter here?

packages/core/src/config/buildSchemaWithPluginContributions.ts
```diff
@@ -0,0 +1,27 @@
+import { commonTokens, PluginResolver, tokens } from '@stryker-mutator/api/plugin';
+import { Logger } from '@stryker-mutator/api/logging';
+
+import { coreTokens } from '../di';
+
+function mergedSchema(mainSchema: any, additionalSchemas: any[]): object {
```
simondel, line 6: aren't these `object`s? (a bit more strict than any)

nicojs: Yeah, but I didn't want to bother with trying to explain TypeScript that it has `properties` and `definitions`. I know they have them and it is a private API for the time being. Didn't want to waste time on writing code to validate that the schemas are valid json schemas.

packages/core/src/di/PluginLoader.ts
nicojs: Yeah, I guess we could. Wanted to keep it in line with the other method. Also want to keep it open for us to do a dynamic search in the future.

e2e/helpers.ts
nicojs: Yeah, i'll improve the name

PR #2164, opened 2020-04-24
feat(validation): warn about unknown stryker config options

### LINKED ISSUE #2103: Improve validation on Stryker options.
Right now there is no standardized and complete way (for both plugins as well as the core) to do the validation. 

For example, this config file will result in no errors at the moment:

```js
{
  "concurrency": 2,
  "reporter": ["clear-text"]
  "thresholds": {
    "error": 80
   }
}
```

**Solution**

I want to allow use [JSON Schema](https://json-schema.org/) to validate. We can use [AJV](https://github.com/epoberezkin/ajv) as validator. As for human-readable error messages we can use [AJV errors](https://github.com/epoberezkin/ajv-errors). As a bonus, we could let AJV [assign defaults](https://github.com/epoberezkin/ajv#assigning-defaults).

Plugins could contribute parts of the schema dynamically by adding a `schema/stryker-plugin.json` (name pending) file in their respective plugins.

Verifying that no additional properties are present can be done by running the schema validation with `additionalProperties: false`. They should be logged as warning. We should allow JSON comments by not warning about settings with the postfix `"comment"`. We could allow a `--noWarnOnAdditionalConfiguration` (name pending) to not warn about additional settings (otherwise people might get frustrated).

nicojs:
Add a warning when options where detected that did not contribute to the option validation.

Example: 

```
16:40:37 (23836) WARN markUnknownOptions Unknown stryker config option "concurrency".
16:40:37 (23836) WARN markUnknownOptions
   Possible causes:
   * Is it a typo on your end?
   * Did you only write this property as a comment? If so, please postfix it with "_comment".
   * You might be missing a plugin that is supposed to use it. Stryker loaded plugins from: ["@stryker-mutator/*"]
   * The plugin that is using it did not contribute explicit validation. 
   (disable "warnings.unknownOptions" to ignore this warning)`
```

You can still disable this by setting `warnings.unknownOptions` to false.

Fixes #2103


nicojs:
Maybe be should call the warning `"unknownOptions"` instead of `"excessOptions"`. What do you think @simondel ?

nicojs:
Maybe be should call the warning `"unknownOptions"` instead of `"excessOptions"`. What do you think @simondel?

nicojs:
I've renamed `excess` to `unknown`. In general: don't use fancy words you barely understand 😉

packages/core/test/unit/config/OptionsValidator.spec.ts
```diff
@@ -221,3 +222,62 @@ describe(OptionsValidator.name, () => {
...
+  it('should validate the options using given optionsValidator', () => {
+    const options = { foo: 'bar' };
+    const output = validateOptions(options, (optionsValidatorMock as unknown) as OptionsValidator);
+    expect(options).eq(output);
+    expect(optionsValidatorMock.validate).calledWith(options);
+  });
+});
+
+describe(markUnknownOptions.name, () => {
+  it('should not warn when there are no unknown properties', () => {
```
nicojs, line 242: test

PR #2172, opened 2020-05-04
feat: add regexp mutator

brodybits:
In parallel to the Array literal mutator and string literal mutator, I would like to propose a RegExp mutator that mutates regex literals and RegExp constructor function calls to help ensure that use of expressions is properly tested.

In case of a non-empty regexp such as `/my-pattern/`, `/my-pattern/g`, `new RegExp('my-pattern')`, or `RegExp(argument)`, mutate as: `new RegExp('')`

In case of an empty regexp such as `new RegExp()` or `RegExp('')`, mutate as: `/Hello from Stryker/`

I hope I do not offend anyone by raising this proposal as a PR. My idea is that the implementation should hopefully explain exactly what I think we should do.

I would like to leave it open for discussion whether or not there is anything we should do with regex flags.

P.S. TODO item is to:

- [ ] document this in the Stryker Handbook, if it will be accepted.

P.S. 2: I would strongly favor getting this proposal merged in a single squash commit, if accepted.

### DIFF
--- a/packages/javascript-mutator/src/mutators/RegExpMutator.ts
+++ b/packages/javascript-mutator/src/mutators/RegExpMutator.ts
@@ -1,0 +1,25 @@
+import * as types from '@babel/types';
+
+import { NodeMutator } from './NodeMutator';
+
+/**
+ * Represents a mutator which can mutate RegExp literal or constructor function call
+ */
+export default class RegExpMutator implements NodeMutator {
+  public name = 'RegExp';
+
+  public mutate(node: types.Node): Array<[types.Node, types.Node | { raw: string }]> {
+    if (types.isRegExpLiteral(node)) {
+      return [[node, { raw: "new RegExp('')" }]];
+    } else if ((types.isCallExpression(node) || types.isNewExpression(node)) && types.isIdentifier(node.callee) && node.callee.name === 'RegExp') {
+      const args = node.arguments;
+      return [
+        args && args.length && !(types.isStringLiteral(args[0]) && args[0].value.length == 0)
+          ? [node, { raw: "new RegExp('')" }]
+          : [node, { raw: '/Hello from Stryker/' }],
+      ];
+    } else {
+      return [];
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


nicojs:
Great idea to add a regex mutator. We're actually working on a regex mutator in the background. @richardwerkman is supervising a graduate student that investigated this and has an implementation for a regex mutator in C#. It does some more interesting stuff (no offense). 

@brodybits would you be willing to help land the feature in Stryker? @richardwerkman is there an online resource that could help brody?

If you don't have the time, that's fine as well. We would be willing to accept this PR in the meantime.

brodybits:
Thanks for the response. Yeah nice to see someone else with the same idea, especially with the same idea "with more interesting stuff". Unfortunately I have very limited time right now, may have some more time in 1-2 months.

I would actually favor holding off on this PR until we can see how we should align the functionality between some of the various languages.

richardwerkman:
@brodybits Our idea is to mutate the logic inside of the regex. We do this by parsing the AST from the regex and mutating the AST just like we do with regular code. We are looking into mutations like:

`[abc]` -> `[ab], [ac], [bc]`
`a{5,10}` -> `a{4,10}, a{6,10}, a{5,11}, a{5,9}`

And many more! You can check the current progress at (wip):

https://github.com/PimRheiter/stryker-net/blob/master/src/Stryker.RegexMutators/Stryker.RegexMutators/RegexMutantOrchestrator.cs

nicojs:
> may have some more time in 1-2 months.

If that's the case, I would suggest waiting a bit. As you know, we're starting to implement mutation switching soon. Once we have that, there will be a single mutator (works for both TS and JS), so it should be easier to implement changes to the mutator.

simondel:
Thanks for opening this PR! As Nico mentioned, we'll be working on mutation switching and we suggest adding this mutator after that work is done. For now I'll close this PR and we hope to see you soon <3

brodybits:
I do completely understand the motivation to hold off on new mutators until the mutation switching is finished. My one concern is losing track of these, in case the author (myself) cannot come back to deal with these mutators for some reason. I think it would be ideal to track these ideas in issues which I would be happy to raise, if they would be welcome for tracking.

nicojs:
If you want to keep track, add an issue for each mutator you want to add/change. That's generally a good idea anyway, as that is where others will look as well.

PR #2175, opened 2020-05-04
feat: add object shorthand property mutator

brodybits:
As I said before, I have been using Stryker pretty heavily and happily in my create-react-native-module utility. I have discovered a few cases over the past year where there were object properties that were simply not needed, sometimes because of functionality removed and sometimes for other reasons. Yes I am the one to be held responsible but think it should be possible for Stryker to help find cases like this.

In short, this is a proposal to try removing shorthand object properties, for example:

- `{ bar, baz: "quz" }` --> `{ baz: "quz" }`

Here are a couple of cleanup fixes that could have been discovered by using this proposal:

- <https://github.com/brodybits/create-react-native-module/pull/276>
- <https://github.com/brodybits/create-react-native-module/pull/328>

I made this proposal on the JavaScript mutator only, would be happy to take a look into the TypeScript version in case there is any interest. I do think it would be ideal to consider using @typescript/eslint-parser which should let us work with the TypeScript parser without having to maintain multiple implementations of the ES mutators.

TODO items __updated__:

- [x] ensure the CI build is green (yes I did check lint & test in my workarea)
- [ ] consider looking for a more succinct name
- [ ] consider if and how this could potentially overlap with other mutations and what we can do
- [ ] add implementation for TypeScript mutator
- [ ] update Stryker Handbook

P.S. I would highly recommend a squash commit merge in case this proposal is accepted.

### DIFF
--- a/packages/javascript-mutator/src/mutators/ObjectShorthandPropertyMutator.ts
+++ b/packages/javascript-mutator/src/mutators/ObjectShorthandPropertyMutator.ts
@@ -1,0 +1,23 @@
+import * as types from '@babel/types';
+
+import { NodeGenerator } from '../helpers/NodeGenerator';
+
+import { NodeMutator } from './NodeMutator';
+
+/**
+ * Represents a mutator which can mutate an object shorthand property.
+ */
+export default class ObjectShorthandPropertyMutator implements NodeMutator {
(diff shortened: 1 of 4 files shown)


simondel:
See #2172

PR #2176, opened 2020-05-04
feat: add arrow function property mutator

brodybits:
Similar to my proposal in #2175 - mutate arrow function properties, for example:

- `({ bar }) => baz` --> `({}) => baz`
- `({ bar, baz = "qux" }) => baz` -->
  - `({ baz = "qux" }) => baz`
  - `({ bar }) => baz`

This proposal could have automatically discovered the following opportunities for cleanup fixes:

- https://github.com/brodybits/create-react-native-module/pull/329
- https://github.com/brodybits/create-react-native-module/pull/330

I have only made this on the JavaScript mutator, would be happy to look into the TypeScript version if there is interest in this proposal.

Some TODO items in case of interest in this proposal:

- [ ] ensure the CI build is green
- [ ] consider looking for a more succinct name
 - [ ] consider if and how this could potentially overlap with other mutations and what we can do
 - [ ] add implementation for TypeScript mutator
 - [ ] update Stryker Handbook

I do think this proposal should be merged in a squash commit, if accepted.

### DIFF
--- a/packages/javascript-mutator/src/mutators/ArrowFunctionPropertyMutator.ts
+++ b/packages/javascript-mutator/src/mutators/ArrowFunctionPropertyMutator.ts
@@ -1,0 +1,25 @@
+import * as types from '@babel/types';
+
+import { NodeGenerator } from '../helpers/NodeGenerator';
+
+import { NodeMutator } from './NodeMutator';
+
+/**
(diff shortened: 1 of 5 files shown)


simondel:
See #2172

PR #2208, opened 2020-05-14
fix(options): resolve false positives in unknown options warning

### LINKED ISSUE #2209: False positive for "unknown options" warning
**Summary**

While testing out the latest new release (3.2.1), I've noticed we get the unknown properties warning on e2e tests.

```
08:35:07 (928) WARN markUnknownOptions Unknown stryker config option "packageManager".
08:35:07 (928) WARN markUnknownOptions Unknown stryker config option "spec".
08:35:07 (928) WARN markUnknownOptions Unknown stryker config option "parent".
```

Only `spec` is a good warning here. Rest are false positives:

* `packageManager` was added by Stryker, shouldn't give a warning.
* `parent` is a side effect from the way we use [commander](npmjs.com/package/commander). 

**Stryker config**

```json
{
  "$schema": "https://raw.githubusercontent.com/stryker-mutator/stryker/master/packages/api/schema/stryker-core.json",
  "mutator": "vue",
  "mutate": [
    "{src,lib}/**/*.+(js|vue)",
    "!{src,lib}/**/*.spec.js"
  ],
  "tempDirName": "stryker-tmp",
  "packageManager": "npm",
  "testRunner": "jest",
  "maxConcurrentTestRunners": 2,
  "coverageAnalysis": "off",
  "reporters": [
    "event-recorder",
    "progress",
    "clear-text",
    "html"
  ],
  "spec": [
    "dist/js/chunk-vendors.js",
    "dist/js/tests.js"
  ]
}

```

nicojs:
* Don't give a warning for "packageManager" config option
* Use `commander.opts` to retrieve options to make sure we don't get unknown properties.

Fixes #2209


simondel:
Do we even have to write it to the config file?

nicojs:
Nope, we dont. But we did it in the past, and that will lead to a lot of warnings. I can see some use for it in the future. Let me write some more tests for this one

nicojs:
Done, added tests for all CLI options

PR #2211, opened 2020-05-15
chore: Add publish command

simondel:


simondel:
We can tweak this in our long-lived feature branch to have a different effect (e.g.: publish prerelease versions with a specific name)

Issue #1980, opened 2020-01-22
Specify lines to mutate


vitaly-rudenko:
**Is your feature request related to a problem? Please describe.**
I would like to be able to specify certain lines to mutate. This is necessary to automatically test changes in pull requests.

**Describe the solution you'd like**
`-m some/file.js:1-5;8-12,some/other/file.js:21,30-40` would mutate only `1-5`, `8-12` lines of `some/file.js`, and `21`, `30-40` lines of `some/other/file.js`

simondel:
Isn’t what you’re looking for more in the lines of incremental analysis? #322

Vitaly-Rudenko:
I'm not sure. I need this to only perform mutation testing for specific lines that have been changed in a specific Pull Request. So I have list of files with corresponding lists of lines that have been changed. I want to provide that into the Stryker so it can check only specified lines of code.

Vitaly-Rudenko:
"Incremental analysis" implies that the feature I describe exists.

Djaler:
> I'm not sure. I need this to only perform mutation testing for specific lines that have been changed in a specific Pull Request. So I have list of files with corresponding lists of lines that have been changed. I want to provide that into the Stryker so it can check only specified lines of code.

But what if your PR changes test?

Vitaly-Rudenko:
I need to run mutation test using all PR's specs, but only mutate selected lines of specified files.
Is this what you're asking?

Garethp:
I'd also really like this feature. My use-case is *similar* to incremental analysis, but more straightforward. I'm building a IntelliJ plugin for Stryker, and I'm looking to implement a `Re-run failed mutants` feature, similar to how you can rerun only failed Jest/Mocha tests. Knowing what mutants failed in the last test run, it'd be easy to just pass in the info needed to identify which mutants I want to run again. 

Though the problem sounds the same as incremental analysis (being able to skip mutants you know are dead), I think the end user use-case is different (being able to only check changed code vs. being able to target specific mutants for specific reasons)

Garethp:
One example for an end-user where specifying lines to mutate would be better than incremental analysis would be a use-case where you've got your CI build telling you that lines 23 and 55 have surviving mutants, so you checkout master and want to run just those two lines through the mutator. No need to rebuild the incremental analysis when changing branches, just specify those two lines and run

Vitaly-Rudenko:
Yes, that's exactly what I meant. This feature will be very useful for CI and scripts.

[...]

Issue #1514, opened 2019-04-19
Implement mutation switching 👽🔀 


nicojs:
## Intro

_[Mutation switching](https://stryker-mutator.io/blog/2018-10-6/mutation-switching)_ is a technique that can be used to improve the performance of mutation testing. It is used in Stryker.NET as well as Stryker4s. We believe this has a lot of benefits for JavaScript/TypeScript as well. 

In short, with mutation switching all mutants will be placed into the source code at once. Only one mutant can be active at the same time. Here is a small example:

```js
/* 1*/ // original
/* 2*/ function add(a, b) {
/* 3*/   return a + b;
/* 4*/ }
/* 5*/ 
/* 6*/ // Might be instrumented to:
/* 7*/ function add(a, b) {
/* 8*/   if (global.activeMutant === 0) {
/* 9*/      // block mutation, empty function
/*10*/   } else {
/*11*/     return global.activeMutant === 1? a - b: a+b;
/*12*/   }
/*13*/ }
```

## Potential performance improvements

Although JavaScript is not a compiled language, there is still huge performance improvements to be had.

* **Less I/O**. This can be a big performance overhead, since not only the mutated file has to be cleared from [node's require cache](https://nodejs.org/api/modules.html#modules_require_cache), but the entire object graph that is dependent on that file, which in practice means clearing the entire cache (100's of files for running Stryker on Stryker). For karma it means a reload between test runs with sometimes MB's of files to reload.
* **One-time transpiling**. A lot of projects depend on some kind of transpiler/bundler, also to run the tests. Think of TypeScript, Webpack, Vue CLI, Angular CLI, Jest-ts, ts-node, etc. They will all have to transpile only once, instead of once per mutant.
* **Hot reload**. Test runners won't have to clear their test suite between tests (there is a caveat here regarding static code, see hurdles). 
[...]

PR #2240, opened 2020-06-01
feat(checker): add checker api

nicojs:
Adds `Checker` plugin type, to be used in #1514 (and can later be used to filter mutants in #1980).

The `Checker` interface looks like this:

```ts
interface Checker {
  init(): Promise<void>;
  check(mutant: Mutant): Promise<CheckResult>;
}
interface CheckResult {
  reason?: string;
  status: CheckStatus;
}
enum CheckStatus {
   Passed = 'passed',
   CompileError = 'compileError',
  // in the future, can include "Ignore" for example.
}
```

## Init

The idea is that, before mutation testing (or even before files are instrumented), we initialize the checker. It can be used as a sanity check. The `TypescriptChecker` can use this to load all files and compile once (with `--no-emit`) to make sure errors don't occur under normal circumstances. 

## Check

Each mutant is checked, one by one. A checker can decide to assign any of the `MutantStatus`'s we've currently reserved for the `report` api. If we agree we should move stuff used by both the `checker` api and `report` api to `core`.

@simondel @Vitaly-Rudenko do you agree with this api?


nicojs:
EDIT: I've included the `MutantStatus` in the changed files. There you can see I've added 2 new states: `Init` and `Ignored`. I've also chosen to implement this with a string enum since that makes it easier to read the states.

We can choose to remove `Ignored` for now, and add it later so we don't make the mutation switching epic too big.

Garethp:
So far looks good. Definitely looks like an interface that I could use to implement a checker for #1980, but I am curious about precedence of Status. If I've got two checkers (`TypeChecker`, `SpecificLinesChecker`) for example, would there be two results on the same mutant? Or would any "non-error" (Ignored, Killed, Timedout etc.) response from the first checker simply stop it from being checked by the second checker? In that case, do the checkers run sequentially (despite returning promises)?

Also, when would a checker return a `Killed` status? My understanding was that checkers were there to try and filter out tests before they reached the test runner stage. What kind of pre-check would you run that would kill a mutant that wouldn't fall into the other status'?

nicojs:
> I am curious about the precedence of Status. If I've got two checkers (`TypeChecker`, `SpecificLinesChecker`) for example, would there be two results on the same mutant?

Great question. The idea I now have is that they would be chained. So the first status != `Init` wins. So in your use case the checkers should be specified as `"checkers": ["SpecificLinesChecker", "TypeChecker"]`.

> Also, when would a checker return a `Killed` status?

A great observation. The only valid results I can think of right now are 
* `Init`: proceed to next checker
* `Ignored`: filter this mutant, don't run it.
* `CompileError`: mutant results in a type error and is therefore invalid

Should we limit the mutant state in the check result to one of these 3 states? We can always add more later if people need it.

Garethp:
Yeah, I think limiting it to those three are a good idea, it makes it a lot more clear when implementing as an outsider to only have those options

Garethp:
Oh, I noticed that `initialize` doesn't accept any arguments. How would the plugin accept external input? Does it need some definitions for dependency injecting settings/arguments?

nicojs:
Yes, good observation once more :). It will be able to let a bunch of stuff be injected. Like all plugins.

See https://www.github.com/stryker-mutator/stryker-handbook/tree/master/stryker%2Fapi%2Fplugins.md

nicojs:
I've renamed the `MutantStatus` enum to `CheckStatus`. It can now only have 3 results:

```ts
enum CheckStatus {
  Ok = 'ok',
  Ignore = 'ignore',
  CompileError = 'compileError',
}
```

We'll probably remove `ignored` before merging this, in order to keep the scope of mutation switching limited. We'll add the `ignored` state in Stryker soon after mutation switching is implemented.

simondel:
Perhaps `Passed`

nicojs:
Yes. That might be better. One passes a check.

Garethp:
`Passed` sounds good. This interface looks good. I can implement a `SpecificLinesFilter` checker once this API is hooked in so that you can get feedback on that specific use-case before release

nicojs:
> This interface looks good. I can implement a `SpecificLinesFilter` checker once this API is hooked in so that you can get feedback on that specific use-case before release

Sure you can! We can also add the specific line number and column number location to the mutant. Right now it only contains `range`, which you can use to calculated it to lines/columns, but that is error prone.

kmdrGroch:
The reason is `?string`? I don't like this version that much :/ I think there should always be reason

nicojs:
I've renamed `Ok` to `Passed`. I've also removed `Ignored` for now. I've also moved `Mutant` from `api/mutant` to `api/core` (since it is now used in 2 api packages). 

Ready for review @simondel

packages/api/src/check/CheckStatus.ts
```diff
@@ -0,0 +1,4 @@
+export enum CheckStatus {
+  Passed = 'passed',
+  CompileError = 'compileError',
```
kmdrGroch, line 3: It doesn't throw an error on `,` ending object?

nicojs: You mean it should give a lint error? Guess not for enums.

kmdrGroch: IMO it should. I don't know why it doesn't :/ Have you changed lint settings? (or maybe during update something was broken ._.

kmdrGroch: Should I make PR for it? I have pretty up to date config with mostly the same options like here :)

nicojs: Sure, a serape PR is fine.

packages/api/src/check/Checker.ts
```diff
@@ -0,0 +1,9 @@
+import { Mutant } from '../../core';
+
+import { CheckResult } from './CheckResult';
+
+export interface Checker {
+  initialize(): Promise<void>;
```
nicojs, line 6: Maybe rename this to `init()`, to be in sync with other apis (i.e. `TestRunner`)
```suggestion
  init(): Promise<void>;
```

Vitaly-Rudenko:
I haven't investigated the stryker's source code yet, but this feature looks very good. Filtering mutants would solve my problem for sure.
Implementing a custom checker will be easy with such API.

Thank you for your hard work!

packages/api/src/check/CheckStatus.ts
Vitaly-Rudenko: I think trailing commas are quite convenient.

PR #2243, opened 2020-06-07
lint: update dependencies and files accordingly

kmdrGroch:
- updates linting dependencies to a current version
- fix issues with linting (duplication of configurations, see: https://github.com/eslint/eslint/issues/13385)
- fix object ending `,`


kmdrGroch:
Small question, should I enable `'@typescript-eslint/ban-types': 'off'` to `error` as it was before? Tho I'll have to change some of the crucial code, since newer ban-types doesn't allow `object` type, and requires more specific Record type (which, I think is actually better than pure object)

kmdrGroch:
@nicojs?

.eslintrc.js
```diff
@@ -23,13 +32,10 @@ module.exports = {
       }
     ],
     '@typescript-eslint/await-thenable': 'off',
-    '@typescript-eslint/ban-ts-ignore': 'error',
```
nicojs, line 26: Why disable this rule? I think `@ts-ignore` should remain banned in favor of [`@ts-expect-error`](https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#ts-expect-error-comments)

.eslintrc.js
```diff
@@ -12,9 +12,18 @@ module.exports = {
   rules: {
     'import/newline-after-import': 1,
     'import/order': ["error", { "newlines-between": "always", "groups": ["builtin", ["external", "internal"], "parent", "sibling", "index"] }],
-    'prettier/prettier': ['error'],
-    'sort-imports': 'off', // No auto-fix!
+    'prettier/prettier': [
```
nicojs, line 15: Can we move this to `.prettierrc`?

.eslintrc.js
```diff
@@ -39,11 +45,10 @@ module.exports = {
     '@typescript-eslint/generic-type-naming': 'off',
     indent: 'off',
     '@typescript-eslint/indent': 'off',
-    '@typescript-eslint/interface-name-prefix': ['error', 'never'],
```
nicojs, line 42: Why remove this rule? I like this one. It's not needed to prefix interfaces with `I` in typescript.

.eslintrc.js
```diff
@@ -23,13 +32,10 @@ module.exports = {
       }
     ],
     '@typescript-eslint/await-thenable': 'off',
-    '@typescript-eslint/ban-ts-ignore': 'error',
-    '@typescript-eslint/ban-types': 'error',
+    '@typescript-eslint/ban-types': 'off',
```
nicojs, line 35: Wy turning this rule off? Seems like a sane list https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md#default-options

package.json
```diff
@@ -45,9 +45,9 @@
   "scripts": {
     "all": "npm run clean && npm run build && npm run lint && npm run test",
     "postinstall": "lerna bootstrap --no-ci && link-parent-bin",
-    "lint": "eslint . --ext .ts,.tsx",
-    "lint:log": "eslint . --ext .ts,.tsx -f compact -o lint.log",
-    "lint:fix": "eslint . --ext .ts,.tsx --fix",
+    "lint": "eslint . --resolve-plugins-relative-to . --ext .ts,.tsx",
```
nicojs, line 48: Why is `--resolve-plugins-relative-to` needed. What does it do?

.eslintrc.js
```diff
@@ -12,9 +12,18 @@ module.exports = {
...
     'import/order': ["error", { "newlines-between": "always", "groups": ["builtin", ["external", "internal"], "parent", "sibling", "index"] }],
-    'prettier/prettier': ['error'],
-    'sort-imports': 'off', // No auto-fix!
+    'prettier/prettier': [
+      'error',
+      {
+        'printWidth': 150,
+        'singleQuote': true,
+        'endOfLine': 'auto',
+        'trailingComma': 'none'
```
nicojs, line 21: To be honest, I kinda like trailing commas. It makes it easier to reformat the code (change the order, add a property, etc). I don't mind the readability.

I think that is also why prettier decided to change the default value to always enforce trailing commas: https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-trailingcomma-to-es5-6963httpsgithubcomprettierprettierpull6963-by-fiskerhttpsgithubcomfisker

@simondel what is your feeling about this?

.eslintrc.js
kmdrGroch: it is removed in newest version

.eslintrc.js
kmdrGroch: `
Small question, should I enable '@typescript-eslint/ban-types': 'off' to error as it was before? Tho I'll have to change some of the crucial code, since newer ban-types doesn't allow object type, and requires more specific Record type (which, I think is actually better than pure object)
`
I was asking bout it coz it was changed a bit and it doesn't like `object`

package.json
kmdrGroch: `fix issues with linting (duplication of configurations, see: eslint/eslint#13385)`
{ extends: "../../tslint.config.json" (or whatever) } doesnt work in newer versions of eslint (7.2.0, tho for now we should stay with 6.8.0 because import plugin doesnt work with newest)

.eslintrc.js
kmdrGroch: I think it was removed in newest versions (gonna check)

.eslintrc.js
kmdrGroch: yea, all information about prettier config can be in eslintrc :) (1 config file less. Everything is in one place <3)

.eslintrc.js
kmdrGroch: yea, it is removed in newest version

kmdrGroch:
cool, so now we are waiting for simons answer :) Meanwhile I'll try fixing ban-types i think. Unless you want to discuss them more @nicojs ?

.eslintrc.js
kmdrGroch: They look pretty ugly, but i mean, its your choice :D

.eslintrc.js
nicojs: Yes, but some people might like to use the prettier plugin in their respective dev env. It's fine for the root of our repo to consist of only config files I think.. that's the way of jsdev nowadays

kmdrGroch: ok, sure

kmdrGroch:
@nicojs finally eslint 7.2.0 <3 I will revert that change with trailing `,` and try fixing this ban-types problem now :)

simondel:
I prefer having trailing commas :)

kmdrGroch:
ok. So I have reverted changes, updated deps and fixed issues with ban-types
anything more @simondel / @nicojs?

.eslintrc.js
nicojs: Ah, I was wondering why I see the `Record` type now in the code. Let me see...

kmdrGroch:
So @nicojs are you okay with these changes? Or should I fix something?

simondel:
I've changed the target branch to our _epic_ branch

packages/core/src/child-proxy/ChildProcessProxyWorker.ts
```diff
@@ -76,7 +76,7 @@ export default class ChildProcessProxyWorker {
     }
   }
 
-  private doCall(message: CallMessage): {} | PromiseLike<{}> | undefined {
+  private doCall(message: CallMessage): Record<string, any> | PromiseLike<Record<string, any>> | undefined {
```
nicojs, line 79: I think al lot of the `Record<string, any>` things could actually be `Record<string, unknown>`. That way we don't accidentally use it when we should narrowed the type with a type guard.

kmdrGroch:
@nicojs finally green with all changes Simon and You have asked for :)

nicojs:
You agree with them as well, right? 😨

PR #2267, opened 2020-06-18
 feat(karma-runner): implement v2 test runner api

nicojs:
* Implement new v2 test runner API (see #2249 )
* Make sure to not report coverage when it isn't needed
* Add a framework adapter file that adds mutant coverage to the karma result, to be picked up by the stryker reporter
* Add test filtering capability when running with jasmine or mocha
* Add integration tests on instrumented code.
* Some more housekeeping.


packages/karma-runner/src/karma-plugins/StrykerReporter.ts
```diff
@@ -98,3 +111,7 @@ export default class StrykerReporter extends EventEmitter implements karma.Repor
     }
   }
 }
+
+// export function isStrykerCoverageReport(event: unknown): event is { event: 'strykerCoverageReport', mutantCoverage: MutantCoverage } {
```
simondel, line 115: Do you still need this?

packages/karma-runner/src/karma-plugins/TestHooksMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+  public currentTestHooks = '';
+
+  public static get instance(): TestHooksMiddleware {
+    if (!this._instance) {
+      this._instance = new TestHooksMiddleware();
+    }
+    return this._instance;
+  }
+
+  public configureTestFramework(frameworks?: string[]) {
```
simondel, line 36: Why is this an array of frameworks and why is it optional?

packages/karma-runner/src/karma-plugins/TestHooksMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+        this.currentTestHooks = `window.${SHOULD_REPORT_COVERAGE_FLAG} = true`;
+        break;
+      case 'off':
+        this.currentTestHooks = `window.${SHOULD_REPORT_COVERAGE_FLAG} = false`;
+        break;
+    }
+  }
+
+  public configureActiveMutant({ activeMutant, testFilter }: MutantRunOptions) {
+    this.currentTestHooks = `window.${SHOULD_REPORT_COVERAGE_FLAG} = false;window.__activeMutant__=${activeMutant.id};`;
```
simondel, line 55: Wouldn't it be nicer to use the coverage analysis configuration for this?
```ts
this.configureCoverageAnalysis('off');
this.currentTestHooks += `;window.__activeMutant__=${activeMutant.id};`;
```

packages/karma-runner/src/karma-plugins/TestHooksMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (testFilter) {
+      switch (this.testFramework) {
```
simondel, line 57: Why not make this an exhaustive switch with a `never` type?

packages/karma-runner/test/helpers/setupTests.ts
```diff
@@ -0,0 +1,24 @@
+import 'source-map-support/register';
```
simondel, line 1: Is this file still used? I don't think the helper files are called anymore from the package.json

packages/karma-runner/src/KarmaTestRunner.ts
```diff
@@ -32,27 +40,32 @@ export default class KarmaTestRunner implements TestRunner {
...
+  }
+
+  public async mutantRun(options: MutantRunOptions): Promise<MutantRunResult> {
+    TestHooksMiddleware.instance.configureActiveMutant(options);
+    const dryRunResult = await this.run();
+    return toMutantRunResult(dryRunResult);
+  }
+
+  private async run(): Promise<DryRunResult> {
+    // this.testHooksMiddleware.currentTestHooks = testHooks || '';
```
simondel, line 67: This was the other instance of commented code that I just mentioned

packages/karma-runner/src/karma-plugins/TestHooksMiddleware.ts
nicojs: We're essentially passing the `frameworks` from karma options. It is in theory optional since you could choose to roll-your-own test framework. 

I wanted to pass them all, since I want all test framework specific stuff to be in the test hooks middleware.

packages/karma-runner/src/karma-plugins/StrykerReporter.ts
nicojs: haha, guess not

nicojs: Removed

packages/karma-runner/src/karma-plugins/TestHooksMiddleware.ts
nicojs: If a test framework isn't selected, I don't want it to break. Instead, just run karma for all tests.

packages/karma-runner/test/helpers/setupTests.ts
nicojs: Helper files configured using the `.mocharc.jsonc` file. I'm migrating mocha as I go along to update modules.

packages/karma-runner/src/KarmaTestRunner.ts
nicojs: Thanks 😅

packages/karma-runner/src/karma-plugins/TestHooksMiddleware.ts
nicojs: Great find! 💖

PR #2268, opened 2020-06-18
Small fixes for Readme

Tummerhore:


simondel:
Thanks! I've changed the npm query to look for `stryker-plugin` instead of `@stryker-plugin`

PR #2269, opened 2020-06-23
 feat(core): simple mutation switching process 

nicojs:
Implement a simple mutation testing process using mutation switching. It represents a new way of running mutation testing, where we place all mutants in the code at the same time. This is only the first step and lacks support for more advanced use cases.

This is a non-exhaustive list of changes:
* Add `Location` to the `Mutant` API, for ease of reporting
* Implement `TestRunner2` factory
* Revamped the way the main Stryker class works. All of the internals are moved to separate `Executor` classes. An `Executor` is responsible for _managing_ a part of the process. I think this helps enormously to understand the general process of mutation testing.
  * `PrepareExecutor`: Responsible to read configuration and input files from disk.
  * `MutantInstrumenterExecutor`: Responsible for instrumenting the source code and generating mutants
  * `DryRunExecutor`: Responsible for performing the initial test run.
  * `MutationTestExecutor`: Responsible for running actual mutation testing.
* Revamped the `Sandbox` class. It is now only responsible for filling a sandbox with files. It no longer manages mutants inside the sandbox or the test runner instance. 
* Revamped the `SandboxPool` -> `TestRunnerPool`, which now manages only test runner instances.
   * An improvement I've managed to sneak in, is that the test runner process from the dry run is now reused in the mutation test run. Not a huge deal, but a nice bonus.
* Removed a bunch of dead code related to the `TestFramework` and `Mutator` plugins, see breaking changes
* Re-enabled 3 simple e2e tests 🎉 

BREAKING CHANGES:

Most of these changes are related to internals of Stryker plugins.

* Transpilers plugins are no longer supported. Transpiler-like functionality will be possible with a build command and checker API. More on those in future PR's
* Custom mutators are no longer supported, Stryker now brings its own mutator in the form of `@stryker-mutator/instrumenter`. It uses babel and an HTML parser under the hood so it has support for all the usual suspects.
* Test Framework plugins are no longer supported. Test runners are supposed to bring their own support for test frameworks they might want to support. This has no consequences for the list of officially supported test runners.
* Test runner v1 API is no longer supported. Please move to the new Test Runner API.


nicojs:
Todo:

- [ ] Unit tests for `PrepareExecutor`
- [ ] Unit tests for `MutantInstrumenterExecutor`
- [ ] Unit tests for `DryRunExecutor`
- [ ] Unit tests for `MutationTestExecutor`
- [ ] Implement `Reporter.onAllMutantsMatchedWithTests`
- [ ] Implement command test runner.
- [ ] Make sure html report works (mutants are not visible a.t.m.)
- [ ] Make sure the progress reporter is working again (`onAllMutantsMatchedWithTests` is needed for that).
- [ ] Remove dead code.

packages/core/src/SandboxPool.ts
```diff
@@ -10,7 +10,7 @@ import { filter, flatMap, map, merge, tap, zip } from 'rxjs/operators';
 import { Disposable } from 'typed-inject';
 
 import { coreTokens } from './di';
-import LoggingClientContext from './logging/LoggingClientContext';
+import { LoggingClientContext } from './logging/LoggingClientContext';
```
kmdrGroch, line 13: Couldn't it be imported from logging index?

nicojs: `SandboxPool` is one of the "dead code" classes I'll remove before marking this pull request as ready for review. Its replacement is the `TestRunnerPool`.

Thanks for asking though.

nicojs:
So yeah the PR turned out to be pretty massive. I've revamped the entire core of Stryker. I also had to break the reporter api in some places, we might need to break it more later. I'll update the PR's description to reflect the current state.

packages/core/src/mutants/MutantTestMatcher.ts
```diff
@@ -1,218 +1,218 @@
...
-    } else {
-      return null;
-    }
-  }
-
-  private isCoveragePerTestResult(_coverage: CoverageCollection | CoveragePerTestResult | undefined): _coverage is CoveragePerTestResult {
-    return this.options.coverageAnalysis === 'perTest';
-  }
-}
+// import { Mutant, StrykerOptions } from '@stryker-mutator/api/core';
```
kmdrGroch, line 1: commented file?

nicojs: Yeah, one of the ones that need to go. Probably best if I remove it indeed. I do want to keep some of it for now. For example, the new `Checker` code can be loosely based on the `Transpiler` code.

kmdrGroch: Although I think it is better to remove it, since we could forget about it later, ok. We just need to keep it in mind :)

nicojs: No, you're right. Classes that are already refactored/replaced should be removed.

packages/core/src/SandboxPool.ts
nicojs: This is now removed

packages/core/src/mutants/MutantTestMatcher.ts
nicojs: Done, all dead code is gone now.

Issue #2278, opened 2020-06-29
@stryker-mutator/core contains vulnerability


peternedap:
**Is your maintenance request related to a problem? Please describe.**
`@stryker-mutator-core` doesn't pass `yarn audit` due to a high risk vulnerability (Command Injection).

**Describe the solution you'd like**
An update to `tree-kill` would fix this.

**Describe alternatives you've considered**
Could this be temporarily fixed with an override (via peer dependency?)

**Additional context**
```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high          │ Command Injection                                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ tree-kill                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=1.2.2                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ @stryker-mutator/core                                        │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ @stryker-mutator/core > tree-kill                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://www.npmjs.com/advisories/1432                        │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

kmdrGroch:
Thanks for opening this issue!
I hope I could fix it with #2279 :) You might want to take a look (I don't have yarn so I don't know if it still marks it as a vulnerability)
Cheers!

peternedap:
Your fix seems to resolve the vulnerability :)

kmdrGroch:
this should be fixed as was merged in #2295

kmdrGroch:
@peternedap stryker 4 has it fixed. Could it be closed? :)

PR #2279, opened 2020-06-29
fix: update tree-kill to the latest version (#2278)

kmdrGroch:
fixes: #2278


nicojs:
... Yeah. We want to merge this to master I think (instead of the mutation switch epic), because maintenance releases are still from master.

kmdrGroch:
I also saw we again introduced ~ and ^ in versions... I think we should change them again to *one version* and leave it like that, since sometimes as could be seen in this example, package is not updated / installed properly... We would have better control over the versions (also it's time to update inside packages not only main package.json)

nicojs:
I think dependabot is also taking care of updates within packages, no? But it won't make a merge request if the new release is within semver match of the current version. 

I think we should keep `~`, it should make sure we keep getting patch releases, right?

For kill-tree, could you make a new PR to master? This one has changes from the epic branch 🤷‍♂️

kmdrGroch:
I think it doesn't check internal package.json files... At least I am not sure if it does. Doing PR now :)

PR #2300, opened 2020-07-07
refactor(mocha-runner): introduce MochaAdapter

nicojs:
Introduce the mocha adapter. It contains all mocha polyfills.


packages/mocha-runner/test/integration/MochaFramework.it.spec.ts
```diff
@@ -1,95 +1,95 @@
-import * as path from 'path';
+// import * as path from 'path';
```
kmdrGroch, line 1: Another file with comments?

PR #2313, opened 2020-07-12
fix(ArrowFunction mutator): don't mutate () => undefined

### LINKED ISSUE #2306: ArrowFunction mutation question
**Question**

I don't get this mutation. Why did this fail?

```

[2020-07-09T21:03:12.538Z] 3. [Survived] ArrowFunction

[2020-07-09T21:03:12.538Z] /src/accordion/accordion-base.tsx:77:13

[2020-07-09T21:03:12.538Z] -       onClick: () => undefined,

[2020-07-09T21:03:12.538Z] +       onClick: () => undefined,
```

simondel:
Fixes #2306

### DIFF
--- a/packages/instrumenter/src/mutators/arrow-function-mutator.ts
+++ b/packages/instrumenter/src/mutators/arrow-function-mutator.ts
@@ -8,9 +8,13 @@ import { NodeMutator } from '.';
 export class ArrowFunctionMutator implements NodeMutator {
   public name = 'ArrowFunction';
 
+  private readonly undefinedIdentifierName = 'undefined';
+
   public mutate(path: NodePath): NodeMutation[] {
-    return path.isArrowFunctionExpression() && !types.isBlockStatement(path.node.body)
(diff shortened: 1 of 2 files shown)


kmdrGroch:
It is so big if statement I would rather use normal "if/else" instead of ternary. It really reduces readability

kmdrGroch:
And could you also check if "() => void" is mutated?

brodybits:
> [...] ternary. It really reduces readability

I think there are 2 major causes:

- using extra-large `---print-width` setting in Prettier
- indenting issue that I just reported in <https://github.com/prettier/prettier/issues/8741>

This is really unfortunate, since I find ternaries to be much more too the point in general.

rodoabad:
Yahoo! Thanks, @nicojs !

nicojs:
Sure thing @rodoabad ! Just remember that this will be part of Stryker 4.0. I'm planning to release a new version of the beta later this week and we'll be working to a release this month or next month.

PR #2326, opened 2020-07-20
Clean up code - provide more accurate types, fix small issues

kmdrGroch:
- Removed many `any` types, while providing better, more accurate type.
- Used concurrency instead of deprecated maxConcurrentTestRunners
- bump up core package version dependency

### DIFF
--- a/packages/core/src/child-proxy/ChildProcessProxyWorker.ts
+++ b/packages/core/src/child-proxy/ChildProcessProxyWorker.ts
@@ -52,8 +52,8 @@ export default class ChildProcessProxyWorker {
     this.log = getLogger(ChildProcessProxyWorker.name);
     this.handlePromiseRejections();
     let injector = buildChildProcessInjector(message.options);
-    const locals = message.additionalInjectableValues as any;
-    for (const token of Object.keys(locals)) {
+    const locals = message.additionalInjectableValues as Record<string, unknown>;
+    for (const token in locals) {
       injector = injector.provideValue(token, locals[token]);
     }
     const RealSubjectClass = require(message.requirePath)[message.requireName];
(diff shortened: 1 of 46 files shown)


kmdrGroch:
@nicojs could you take a look at this small feature? :)

packages/core/src/child-proxy/ChildProcessProxyWorker.ts
```diff
@@ -52,8 +52,8 @@ export default class ChildProcessProxyWorker {
     this.log = getLogger(ChildProcessProxyWorker.name);
     this.handlePromiseRejections();
     let injector = buildChildProcessInjector(message.options);
-    const locals = message.additionalInjectableValues as any;
-    for (const token of Object.keys(locals)) {
+    const locals = message.additionalInjectableValues as Record<string, unknown>;
+    for (const token in locals) {
```
nicojs, line 56: `for` ... `in` will also iterate over properties on prototype objects. But in this case, it doesn't matter, because `additionalInjectableValues`'s prototype is `Object` (and does not have any properties to iterate over).

https://stackoverflow.com/questions/43446241/for-in-or-for-of-object-keys

packages/core/src/test-runner/TestRunnerPool.ts
```diff
@@ -58,8 +58,8 @@ export class TestRunnerPool implements Disposable {
       numConcurrentRunners--;
```
nicojs, line 58: Wow, we still have a `TestRunnerPool`? This class is replaced by the generic `Pool<T>` 😅. Will remove in another PR.

packages/karma-runner/package.json
```diff
@@ -44,7 +44,7 @@
     "karma-mocha": "^2.0.0"
   },
   "peerDependencies": {
-    "@stryker-mutator/core": "^3.0.0"
+    "@stryker-mutator/core": "4.0.0-beta.1"
```
nicojs, line 47: This feels strange. Versioning should have been done lerna.

nicojs:
Thanks!

packages/core/src/child-proxy/ChildProcessProxyWorker.ts
kmdrGroch: Yea, I checked it first of course to see if it breaks :)

PR #2374, opened 2020-08-12
Cleaning up

kmdrGroch:
- potential bug fixed in execa
- rename some variables for better context
- remove unnecessary type checks, refactoring them
- fix some type definitions
- Timer and mutators refactor and better typing
- remove unnecessary file


packages/core/src/utils/StringBuilder.ts
```diff
@@ -23,7 +23,7 @@ export default class StringBuilder {
   public static concat(...builders: StringBuilder[]): string {
     return builders
       .map((b) => b.toString())
-      .filter(Boolean)
```
nicojs, line 26: I actually like the `filter(Boolean)`. It's quite common in my day-to-day job. :)

packages/core/src/utils/Timer.ts
```diff
@@ -39,21 +40,21 @@ export default class Timer {
...
-      return `${elapsedMinutes} minute `;
-    } else {
+    if (elapsedMinutes === 0) {
       return '';
+    } else {
+      return this.formatTime('minute', elapsedMinutes);
     }
   }
+
+  private static formatTime(word: 'minute' | 'second', elapsed: number) {
```
nicojs, line 55: Great addition!

packages/core/src/utils/fileUtils.ts
```diff
@@ -14,9 +14,9 @@ export function glob(expression: string): Promise<string[]> {
 }
 
 export function deleteDir(dirToDelete: string): Promise<void> {
-  return new Promise<void>((resolve, reject) => {
+  return new Promise<void>((_, reject) => {
```
nicojs, line 17: So this promise is never resolved?

packages/core/test/integration/child-proxy/Echo.ts
```diff
@@ -26,9 +26,6 @@ export class Echo {
 
   public exit(code: number) {
     process.exit(code);
-    return new Promise(() => {
```
nicojs, line 29: This is a test case, it is actually supposed to never resolve.

packages/core/src/utils/StringBuilder.ts
kmdrGroch: Ok :)

packages/core/src/utils/fileUtils.ts
kmdrGroch: I think it should not resolve like this at least. From what i could see Error callback always return Error so check error ? reject() : resolve()  will always return reject()

kmdrGroch:
That should satisfy you @nicojs :)

packages/core/test/integration/child-proxy/Echo.ts
kmdrGroch: It is not even running tho :/ process.exit works like return so that promise line is not even executed. I mean as long as its fine with you its ok (tslint or sth just kept removing it :x)

nicojs: Ah, I see. You are correct in removing it then.

packages/core/src/utils/fileUtils.ts
```diff
@@ -16,8 +16,9 @@ export function glob(expression: string): Promise<string[]> {
 export function deleteDir(dirToDelete: string): Promise<void> {
   return new Promise<void>((resolve, reject) => {
```
nicojs, line 17: refactor tip:

```ts
import { promisify } from 'util';

export const deleteDir = promisify(rimraf);
```

PR #2380, opened 2020-08-13
fix(exit prematurely): exit when no tests were executed

nicojs:
BREAKING CHANGE: Stryker will now exit with exit code 1 when no tests were executed in the initial test run.


e2e/test/exit-prematurely-no-tests-executed/src/add.js
```diff
@@ -0,0 +1,6 @@
+module.exports.add = function(num1, num2) {
+  return num1 + num2;
+};
+
```
kmdrGroch, line 4: Are all these new lines necessary tho? :c

Issue #2253, opened 2020-06-10
Append a plugin with command line (without removing configured plugins)


Garethp:
**Is your feature request related to a problem? Please describe.**
I want to be able to add a plugin through command line (A new reporter), without removing existing plugins. When writing a tool that tried to add functionality to Stryker, I don't know what plugins are being included with the configuration file, but if I try to add it through `--plugins`, then it removes any plugins already specified (Such as `typescript`)

**Describe the solution you'd like**
I'd like a new `--plugin` argument that *appends* to the array of plugins, rather than replacing it

**Describe alternatives you've considered**
I've considered parsing the config file, but that seems rather onerous since my tool isn't built in JS. I've also considered just adding `@stryker-mutator/*`, but the user might be using other plugins (Like I intend to with Specific Lines filter). 

A workaround that I'll likely use without this feature is to ask users to have this plugin specified in their config file, but it's not something that my tool can verify that they're using (and therefore would cause an error if I try to specify that this reporter should be used), so it would better if my tool can explicitly say that this plugin should be in use

nicojs:
Cool! I think I now understand what you mean.

I see 2 ways of implementing this. 

1. We change the way the `--plugins` works and let it just append plugins (breaking change). We could still allow users to still delete all plugins with `--no-plugins` for example.
2. We add a new command-line argument `--append-plugins` which prevents `plugins` to be cleared (minor change). 

I'm actually leaning towards option 2. 

What do you think @Garethp?

cc @simondel

Garethp:
Option 2 sounds like the clearest way to interact with it

nicojs:
Merged with #2385

PR #2385, opened 2020-08-15
feature(core): Adding `appendPlugins` commandline option

Garethp:
A PR for issue #2253

### DIFF
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -186,23 +186,6 @@
           ],
           "default": null
         },
-        "appendPlugins": {
-          "anyOf": [
-            {
-              "type": "array",
-              "items": {
-                "type": [
-                  "string",
-                  "array"
-                ]
-              }
-            },
-            {
-              "type": "null"
-            }
-          ],
-          "default": null
-        },
         "excludedMutations": {
           "type": "array",
           "items": {
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -186,6 +186,23 @@
           ],
           "default": null
         },
+        "appendPlugins": {
+          "anyOf": [
+            {
+              "type": "array",
+              "items": {
+                "type": [
+                  "string",
+                  "array"
+                ]
+              }
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null
+        },
         "excludedMutations": {
           "type": "array",
           "items": {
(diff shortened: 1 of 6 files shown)


packages/core/test/unit/di/buildMainInjector.spec.ts
```diff
@@ -98,6 +98,7 @@ describe(buildMainInjector.name, () => {
     const expected: MutatorDescriptor = {
       name: 'javascript',
       plugins: null,
+      appendPlugins: null,
```
nicojs, line 101: Plugins of the mutator descriptor is a list of _babel plugins for parsing source files_, it doesn't have anything to do with `appendPlugins` 😉

packages/core/src/StrykerCli.ts
```diff
@@ -79,6 +79,11 @@ export default class StrykerCli {
       .option('--transpilers <listOfTranspilers>', 'A comma separated list of transpilers to use.', list)
       .option('--reporters <name>', 'A comma separated list of the names of the reporter(s) you want to use', list)
       .option('--plugins <listOfPlugins>', 'A list of plugins you want stryker to load (`require`).', list)
+      .option(
+        '--appendPlugins <listOfPlugions>',
+        'A list of plugins you want stryker to load (`require`) without overwriting the plugins in the config file',
```
nicojs, line 84: > "A list of additional plugins you want Stryker to load (`require`) without overwriting the (default) `plugins`."

packages/api/schema/stryker-core.json
```diff
@@ -220,6 +220,23 @@
           ],
           "default": null
         },
+        "appendPlugins": {
+          "anyOf": [
```
nicojs, line 224: Please force this to be an array. Default would be an empty array.

packages/core/src/StrykerCli.ts
```diff
@@ -79,6 +79,11 @@ export default class StrykerCli {
       .option('--testRunner <name>', 'The name of the test runner you want to use')
       .option('--reporters <name>', 'A comma separated list of the names of the reporter(s) you want to use', list)
       .option('--plugins <listOfPlugins>', 'A list of plugins you want stryker to load (`require`).', list)
+      .option(
+        '--appendPlugins <listOfPlugions>',
+        'A list of additional plugins you want Stryker to load (`require`) without overwriting the (default) `plugins`.',
```
nicojs, line 84: Why is this description so different from the one on `stryker-core.json`? I usually try to keep both explanations and the `readme` in sync.

nicojs (approved):
I would like to use the `appendPlugins` option in the e2e tests, that way we know for sure it works as expected. You can choose an existing e2e test for this, maybe one that uses 2 plugins?

I think this can land as is. Please let me know if you want to add that

nicojs:
Thanks for the recent changes. Will merge when the pipeline is green,

Issue #2383, opened 2020-08-13
Remove/deprecate old stryker options


nicojs:
We should remove or deprecate the stryker options that v4 no longer uses. A (maybe non-extensive) list:

* `transpilers`
* `mutator.name`
* Defining `mutator` as string
* `testFramework`

We should remove these, or log a warning when people use them. Maybe we can give a comprehensive error message pointing people to the blog article (that needs to be written).

kmdrGroch:
@nicojs I think as long as we update schema and people would actually use it, they will know that they cant use these options anymore (thats for "log a warning when people use them")

nicojs:
Yeah, true. But that might be a unfriendly message. I was thinking to write a custom message with a link to a future blog article or migration guide thing

PR #2395, opened 2020-08-18
[WIP] Remove/deprecate old stryker options

### LINKED ISSUE #2382: Update `stryker init` with new template for Stryker v4
With Stryker v4 a number of things have changed that also need to be reflected in `stryker init`:

* No more test frameworks (simply remove that question).
* No more transpilers -> Replace with question about which Checker to enable.
* Add question about a build command to run.

Also some configuration options have been changed/added. These should be renamed in the template for the options. Also update `initStrykerConfig` in each package.json

kmdrGroch:
fixes: #2383
fixes #2382


kmdrGroch:
@nicojs I have so far successfully removed mutator and transpilers (I mean, it builds), tho I still can't make some tests pass (dont look at e2e, I haven't even touched them so far). If you have any idea what about these failing tests I would appreciate some guide

kmdrGroch:
@nicojs there is only one small issue in E2E now, which if possible I would like to ask you to fix. Otherwise everything should be alright (only README changes left at this stage IMO)

Issue #2362, opened 2020-08-07
Stryker v4 Instrumentor breaks enzyme snapshots


Garethp:
**Summary**
When using Stryker v4, my snapshots (using enzyme) break because where in the main run it recognizes the component name, in the mutant run it just calls it `Component`.

Example: 

```tsx
const ErrorMessage = ({ message, id }: Props) => (
  <React.Fragment>
    {message.length > 0 && (
      <span id={id} className="error-message">
        {message}
        <span className="hidden">Error: {message}</span>
      </span>
    )}
  </React.Fragment>
);

export default ErrorMessage;
```

```tsx
interface IFormLabelProps {
  label: string;
  forId: string;
  errorMessage?: string;
}

const FormLabel: FunctionComponent<IFormLabelProps> = ({
  label,
  forId,
  errorMessage
}) => {
  return (
    <>
      <label className="label" htmlFor={forId}>
        {label}
      </label>
      {!!errorMessage?.length && (
        <ErrorMessage id={`${forId}-error`} message={errorMessage} />
      )}
    </>
  );
};
```

```tsx
it("should render with error message", () => {
  const formLabel = shallow(
    <FormLabel
      label="test label"
      forId="testId"
      errorMessage="test error message"
    />
  );
  expect(formLabel).toMatchSnapshot();
});
```

In the main run it will return the snapshot 

```js
exports[`should render with error message 1`] = `
<Fragment>
  <label
    className="label"
    htmlFor="testId"
  >
    test label
  </label>
  <ErrorMessage
    id="testId-error"
    message="test error message"
  />
</Fragment>
`;
```

While in the mutant it'll render

```js
exports[`should render with error message 1`] = `
<Fragment>
  <label
    className="label"
    htmlFor="testId"
  >
    test label
  </label>
  <Component
    id="testId-error"
    message="test error message"
  />
</Fragment>
`;
```

The difference being that `ErrorMessage` has been replaced with `Component`. This is because the instrumentor replaced the main definition of `ErrorMessage` with a ternary, and it appears that enzyme (or react, not sure) is looking for the variable name of the original variable that the component is assigned to. 

If you replaced ErrorMessage with

```tsx
const ErrorMessage = true ? ({ message, id }: Props) => (
  <React.Fragment>
    {message.length > 0 && (
      <span id={id} className="error-message">
        {message}
        <span className="hidden">Error: {message}</span>
      </span>
    )}
  </React.Fragment>
) : () => undefined;
```

[...]

PR #2404, opened 2020-08-20
bug(instrumentor): Adding instrumentor bug e2e's

Garethp:
* Variables that are defined as `export const  a = () => <div></div>` should *not* have a node name #2362
 * For some reason `https://` as part of a string is instrumented into an unterminated string

### DIFF
--- a/e2e/test/jest-react-enyme-ts/src/OtherChild.spec.tsx
+++ b/e2e/test/jest-react-enyme-ts/src/OtherChild.spec.tsx
@@ -1,0 +1,9 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { OtherChild } from './OtherChild';
+
+describe('OtherChild Component', () => {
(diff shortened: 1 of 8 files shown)


nicojs:
@Garethp are you sure arrow functions that are exported should not have a `name`? It _does_ have a name in es-modules in nodejs. 

Example:

```js
// a.mjs
export const a = () => {}

// b.mjs
import { a } from './a.mjs';

console.log(a.name);
```

Now `node b.mjs` prints: `a` 🤷‍♂️

Garethp:
I've made a PR with an e2e test that should show the use case that I'm talking about. I don't fully understand it myself to be honest

Garethp:
I'm not sure. I know that in this specific instance the snapshot shows up without a name. I'll dig into it to see what's going on

nicojs:
Closed as discussed in Stryker community meeting.

PR #2405, opened 2020-08-20
improve default sandbox option strings to ignore tests & test directories

brodybits:
context: sandbox workarounds needed to work with React - part of issue #2403

Issue #2403, opened 2020-08-20
Multiple config workarounds needed for dry run to work for React [Stryker 4.0.0-beta]


brodybits:
**Question**

As I said in #1514, I needed some config workarounds for the dry run to work with <https://github.com/facebook/react>. I got the dry run working, now on 4.0.0-beta.3, with a bit trial-and-error:

```js
/**
 * @type {import('@stryker-mutator/api/core').StrykerOptions}
 */
module.exports = {
  mutator: 'javascript',
  packageManager: 'yarn',
  reporters: ['html', 'clear-text', 'progress'],
  testRunner: 'command',
  transpilers: [],
  coverageAnalysis: 'off',
  // workarounds starting here:
  buildCommand: 'yarn',
  symlinkNodeModules: false,
  sandbox: {
    fileHeaders: {},
    stripComments: false
  },
};
```

I don't know if any of the config items may be redundant. This is not a blocker but seems to be a bit counterintuitive.

**Stryker environment**

```
$ npm ls | grep stryker
├─┬ @stryker-mutator/core@4.0.0-beta.3
│ ├─┬ @stryker-mutator/api@4.0.0-beta.3
│ ├─┬ @stryker-mutator/instrumenter@4.0.0-beta.3
│ │ ├── @stryker-mutator/api@4.0.0-beta.3 deduped
│ │ ├── @stryker-mutator/util@4.0.0-beta.3 deduped
│ ├─┬ @stryker-mutator/util@4.0.0-beta.3
```

Test runner is command.

As a side point, I suspect it should be possible to get the Jest runner working with React with the right kind of utility script rework.

**Additional context**

I am generally testing with `stryker.conf.js` which seems to be a bit more flexible than working with JSON.

I did have to comment a shebang out of `scripts/jest/jest.js` in React to get past issue #2398.

I think this is less important than other issues #2400, #2399, #2398.

nicojs:
Thanks again for the help!

I _think_ the problems is mostly with the `sandbox` settings.

```
sandbox: {
    fileHeaders: {},
    stripComments: false
  },
```

(btw, great you found those!)

[...]

PR #2406, opened 2020-08-20
cleanup & simplify default sandbox update options

brodybits:
__updated:__

## Proposed changes

- simplify & cleanup the extension matching pattern
- only update TypeScript sources
- skip update of `**/*.spec.ts*` & `**/*.test.ts*`
- remove the `eslint-disable ` line
- update test cases

## Motivation

- Coming from issue #2403, I started thinking that the sandbox updates are only needed to work around the issues between the mutator and TypeScript.
- I think it should be possible to skip test files, which should be straightforward for projects that use `*.spec.{js|ts}*` or `*.test.{js|ts}*`.
- I am also assuming that a smart Stryker configuration would *not* run a lint step during mutation testing.
- I tried something more elaborate in PR #2405, but it does not look practical without some more updates.

## Additional notes

This proposal does not skip test files in a `test`, `tests`, or `__tests__` directory if they do not use the standard `*.spec.{js|ts}*` or `*.test.{js|ts}*` naming conventions, in <https://github.com/facebook/react> for example. But this should not even be an issue with React, which is in JavaScript, if we would only update the TypeScript sources.

To skip the test files without these standard naming conventions would involve a bit more work. I am not so sure this would be worthwhile, maybe better to have TypeScript users switch to these standard naming conventions if needed for the test files.

---

I am proposing several improvements together, and would be happy to back some of the changes if needed.

### DIFF
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -176,7 +176,7 @@
             "type": "string"
           },
           "default": {
-            "**/*+(.js|.ts|.cjs|.mjs)?(x)": "/* eslint-disable */\n// @ts-nocheck\n"
+            "**/!(*.spec,*.test).{ts,tsx}": "// @ts-nocheck\n"
           }
         },
         "stripComments": {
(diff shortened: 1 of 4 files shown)


brodybits:
Closing in favor of #2412

Issue #2407, opened 2020-08-21
Only update sources to be mutated in the sandbox


brodybits:
**Is your feature request related to a problem? Please describe.**

One part of issue #2403 is that when we add file headers and strip comments, it leads to issues between the React library (<https://github.com/facebook/react>) and Jest. This can be solved by excluding the JavaScript sources, like I proposed in #2406, but it got me thinking about a more general solution.

If we would only update sources to be mutated in the sandbox, and keep the other sources as-is, it may avoid this kind of an issue with some other projects in the future.

**Describe the solution you'd like**

rework the sandbox setup as follows:

- get all source files, like Stryker already does
- only match files to mutate with the `sandbox` patterns
- if a file both is in the files-to-mutate list and matches `sandbox.fileHeaders` pattern, then add the header
- if a file both is in the files-to-mutate list and matches `sandbox.stripComments` pattern, then strip the comments

**Describe alternatives you've considered**

__updated:__

- my proposal in PR #2412 should solve the same problem with smaller projects
- exclude `**/*.spec.*` and `**/*.test.*`, like I proposed in PR #2406
- exclude sources in `**/__tests__/**`, `**/tests/**`, `**/test/**`, `**/spec/**`, like I tried to do in PR #2405
- configure `sandbox` to not update any sources, like I did for React as described in issue #2403

**Additional context**

I tried working on this idea myself. I think this should be pretty straightforward with Injector from `typed-inject` but do not yet really understand how to work with the `typed-inject` API.

---

P.S. After some testing, I think this would not solve the problem with the React library. So this idea may not be worth considering for the near future.

nicojs:
Hi @brodybits once again. I overlooked this issue, sorry for the late(r) reply.

> If we would only update sources to be mutated in the sandbox, and keep the other sources as-is, it may avoid this kind of an issue with some other projects in the future.

Unfortunately, this won't work for the typescript use case, as stated in #2415 

For example:

```ts
// concat.ts
function concat(a: string, b: string) {
   return a + b; // return type of this function will change because of mutation switching
}
// __tests__/concat.spec.ts
it('should concat', () => {
  const actual: string = concat('a', 'b');
  expect(actual).eq('ab');
});
```

In this case, `__tests__/concat.spec.ts` will give a compile error. You would notice it if you're using `buildCommand: "tsc -b"` or you use `jest-ts` for example, the test runner that supports type checking.

[...]

PR #2412, opened 2020-08-21
improve & focus sandbox preprocessing options

brodybits:
- add `sandbox.ignorePatterns` functionality to sandbox preprocessing, ignores test files & `__tests__` directory by default
- rename sandbox.fileHeaders -> sandbox.addHeaders, for consistency (all verbs)
- rename sandbox.stripComments -> sandbox.removeComments (bikeshedding)
- simplify default sandbox file patterns
- limit default sandbox preprocessing to ts & tsx files
- remove eslint-disable line from default sandbox.addHeaders - I think that smart Stryker configuration would not run eslint during mutation testing

This proposal was triggered by my investigation of __#2403__. I tried a couple other ideas but they are probably not worth considering.

It is very opinionated to the point of bikeshedding, but I broke the changes into individual commits. We should be able to revert or remove any changes not wanted in this proposal.

We may want to rename a few modules if we rename the options like I proposed, which should be pretty trivial.

### DIFF
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -188,7 +188,7 @@
             "type": "string"
           },
           "default": {
-            "**/*.{ts,tsx}": "/* eslint-disable */\n// @ts-nocheck\n"
+            "**/*.{ts,tsx}": "// @ts-nocheck\n"
           }
         },
         "removeComments": {
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -168,6 +168,18 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "ignorePatterns": {
+          "description": "Ignore these patterns when preprocessing the sandbox files",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [
+            "**/__tests__/**/*+(.js|.ts|.cjs|.mjs)?(x)",
+            "**/?(*.)+(spec|test)+(.js|.ts|.cjs|.mjs)?(x)",
+            "**/*+(Spec|Test)+(.js|.ts|.cjs|.mjs)?(x)"
+          ]
+        },
         "fileHeaders": {
           "type": "object",
           "title": "SandboxFileHeaders",
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -180,7 +180,7 @@
             "**/*+(Spec|Test)+(.js|.ts|.cjs|.mjs)?(x)"
           ]
         },
-        "fileHeaders": {
+        "addHeaders": {
           "type": "object",
           "title": "SandboxFileHeaders",
           "description": "Configure additional headers to be added to files inside your sandbox. These headers will be added after Stryker has instrumented your code with mutants, but before a test runner or build command is executed. This is used to ignore typescript compile errors and eslint warnings that might have been added in the process of instrumenting your code with mutants. The default setting should work for most use cases.",
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -191,7 +191,7 @@
             "**/*+(.js|.ts|.cjs|.mjs)?(x)": "/* eslint-disable */\n// @ts-nocheck\n"
           }
         },
-        "stripComments": {
+        "removeComments": {
           "description": "Configure files to be stripped of comments (either single line with `//` or multi line with `/**/`. These comments will be stripped after Stryker has instrumented your code with mutants, but before a test runner or build command is executed. This is used to remove any lingering `// @ts-check` or `// @ts-expect-error` comments that interfere with typescript compilation. The default setting allows comments to be stripped from all JavaScript and friend files in your sandbox, you can specify a different glob expression or set it to `false` to completely disable this behavior.",
           "anyOf": [
             {
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -175,9 +175,9 @@
             "type": "string"
           },
           "default": [
-            "**/__tests__/**/*+(.js|.ts|.cjs|.mjs)?(x)",
-            "**/?(*.)+(spec|test)+(.js|.ts|.cjs|.mjs)?(x)",
-            "**/*+(Spec|Test)+(.js|.ts|.cjs|.mjs)?(x)"
+            "**/__tests__/**/*.{js,ts,cjs,mjs}?(x)",
+            "**/*.{spec,test}.{js,ts,cjs,mjs}?(x)",
+            "**/*{Spec,Test}.{js,ts,cjs,mjs}?(x)"
           ]
         },
         "addHeaders": {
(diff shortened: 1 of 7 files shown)


packages/api/schema/stryker-core.json
```diff
@@ -168,18 +168,30 @@
...
           "additionalProperties": {
             "type": "string"
           },
           "default": {
-            "**/*+(.js|.ts|.cjs|.mjs)?(x)": "/* eslint-disable */\n// @ts-nocheck\n"
+            "**/*.{ts,tsx}": "// @ts-nocheck\n"
           }
         },
-        "stripComments": {
+        "removeComments": {
```
nicojs, line 194: So `remove` is better than `strip`? I guess I've used the naming of the `strip-comments` dependency. Remove sounds better indeed.

packages/api/schema/stryker-core.json
```diff
@@ -168,18 +168,30 @@
...
+        "addHeaders": {
           "type": "object",
           "title": "SandboxFileHeaders",
           "description": "Configure additional headers to be added to files inside your sandbox. These headers will be added after Stryker has instrumented your code with mutants, but before a test runner or build command is executed. This is used to ignore typescript compile errors and eslint warnings that might have been added in the process of instrumenting your code with mutants. The default setting should work for most use cases.",
           "additionalProperties": {
             "type": "string"
           },
           "default": {
-            "**/*+(.js|.ts|.cjs|.mjs)?(x)": "/* eslint-disable */\n// @ts-nocheck\n"
+            "**/*.{ts,tsx}": "// @ts-nocheck\n"
```
nicojs, line 191: Actually, `@ts-nocheck` is also needed for `js` files if typescript is compiling with the [`--checkJS` option](https://www.typescriptlang.org/tsconfig#checkJs)

packages/api/schema/stryker-core.json
```diff
@@ -168,18 +168,30 @@
...
+        },
+        "addHeaders": {
           "type": "object",
           "title": "SandboxFileHeaders",
           "description": "Configure additional headers to be added to files inside your sandbox. These headers will be added after Stryker has instrumented your code with mutants, but before a test runner or build command is executed. This is used to ignore typescript compile errors and eslint warnings that might have been added in the process of instrumenting your code with mutants. The default setting should work for most use cases.",
           "additionalProperties": {
             "type": "string"
           },
           "default": {
-            "**/*+(.js|.ts|.cjs|.mjs)?(x)": "/* eslint-disable */\n// @ts-nocheck\n"
```
nicojs, line 179: I see you've remvoed the `eslint-disable` check. You think build commands won't fail on eslint checks?

packages/core/README.md
```diff
@@ -294,12 +295,21 @@ The `clear-text` reporter supports three additional config options:
 
 The `dashboard` reporter sends a report to https://dashboard.stryker-mutator.io, enabling you to add a mutation score badge to your readme, as well as hosting your html report on the dashboard. It uses the [dashboard.*](#dashboard) configuration options. See [the Stryker handbook](https://github.com/stryker-mutator/stryker-handbook/blob/master/dashboard.md) for more info.
 
-<a name="sandbox.fileHeaders"></a>
-### `sandbox.fileHeaders` [`object`]
+<a name="sandbox.ignorePatterns"></a>
+### `sandbox.ignorePatterns` [`string[]`]
```
nicojs, line 299: So this is like a "master switch"? Wouldn't that be confusing? What does this mean:

```js
{
  sandbox: {
    ignorePatterns: "**/*.js",
    removeComments: "**/*.js"
  }
```

packages/core/README.md
```diff
@@ -294,12 +295,21 @@ The `clear-text` reporter supports three additional config options:
...
 
-Default: `{ "**/*+(.js|.ts|.cjs|.mjs)?(x)": "/* eslint-disable */\n// @ts-nocheck\n" }`
+Default: `['**/__tests__/**/*.{ts,tsx}', '**/*.{spec,test}.{ts,tsx}', '**/*{Spec,Test}.{ts,tsx}']`
```
nicojs, line 301: Ignoring the test files will not help in scenario's where typescript is used.

For example:

```js
// concat.ts
function concat(a: string, b: string) {
   return a + b; // return type of this function will change because of mutation switching
}

// __tests__/concat.spec.ts
it('should concat', () => {
  const actual: string = concat('a', 'b');
  expect(actual).eq('ab');
});
```

In this case, `__tests__/concat.spec.ts` will give a compile error. You would notice it if your using `buildCommand: "tsc -b"` or you use `jest-ts` for example.

packages/core/README.md
```diff
@@ -310,7 +320,7 @@ The key here is a [glob expression](https://globster.xyz/), where the value poin
 <a name="sandbox.stripComments"></a>
 ### `sandbox.stripComments` [`false` | `string`]
```
nicojs, line 321: Note: should be removed still to `removeComments`

packages/core/src/sandbox/strip-comments-preprocessor.ts
```diff
@@ -1,7 +1,7 @@
 import path = require('path');
```
nicojs, line 1: This file should also be renamed to `remove-comments-preprocessor`

packages/core/src/sandbox/strip-comments-preprocessor.ts
```diff
@@ -17,13 +17,13 @@ export class StripCommentsPreprocessor implements FilePreprocessor {
   constructor(private readonly options: StrykerOptions) {}
```
nicojs, line 17: This class should also be renamed to `RemoveCommentsPreprocessor`.

nicojs:
> I think React may be a special case

So is Angular or Vue. They all need special configuration to make work. 

I was thinking, instead of improving the defaults for these use cases (always a trade-off, improving one worsens another)  we could also choose to add "preset" functionality.

For example, `--preset react` would set the sandbox config needed to run "react". That would lessen the need for more custom overrides.

However, we might still need a way to leave the `@jest-environment` in the code, while adding `@ts-nocheck` and removing all other `@ts-xxx` directives. (for the `jest-ts` use case, the jest transpiler that _does type checking_).

packages/core/README.md
brodybits: > So this is like a "master switch"?

yes

My idea was that it should be possible to ignore all files in `__tests__` for example, even though some files in `__tests__` would match a pattern such as `**/*.{js,ts}` or `**/!(*.spec,*.test,*Spec).{js,ts}`.

But yeah, this idea would probably be too confusing. Back to the drawing board.

brodybits:
By "React may be a special case" I had meant "React library (<https://github.com/facebook/react>) may be a special case". I updated my comment to make this more clear.

That said, the idea of adding and using the "preset" functionality sounds nice and may prove quite helpful down the road. I imagine it would be nice for a user to start with a "preset" and customize a limited number of items, like I see in the Renovate bot. Maybe this is what you meant from the beginning.

Unfortunately I am a bit lost in terms of the Jest functionality here. But I was just thinking it may (or may not) be useful if we would only do the sandbox preprocessing updates on files to be mutated and not test files. That's what I wanted to do with the ignorePatterns option, since I could not see a way to have the existing fileHeaders & stripComments ignore all files in `__tests__` directory.

But as I said in the comment above, I think you are right that this would be too confusing. I did raise another idea in #2407 to only update sources to be mutated in the sandbox, but think it would not solve with problem with the React library.

As I said before, my proposal to rename "stripComments" to "removeComments" may be a form of bikeshedding. Some could find "strip" to be a little offensive, that is why I proposed renaming it. But not a really big deal on my part.

I would be happy to move the changes to rename the 2 sandbox options into a separate PR, and rename the source modules accordingly in the separate PR. Maybe sometime later this week.

nicojs:
Closed in favor of #2446

PR #2416, opened 2020-08-24
feature(jest-runner): Pass the active mutant id through the env vars rather than config

Garethp:
This represents a rather massive difference in speed on our project. I can get propper numbers tomorrow, but from a quick look it's somewhere between 2x-4x at least.

It looks like the integration tests for jest-runner are failing. I'm not sure if it's because I missed something, or the integration tests missed something. It looks like the integration mutants aren't picking up the mutant id from the environment variables


packages/jest-runner/testResources/exampleProject/src/Add.js
```diff
@@ -1,30 +1,21 @@
 // @ts-nocheck
-var __global_69fa48 = function (g) {
-  g.__mutantCoverage__ = g.__mutantCoverage__ || {
-    static: {},
-    perTest: {}
-  };
-
+var __global_69fa48 = (function(g){
+  if (g.__activeMutant__ === undefined && g.process && g.process.env && g.process.env.__STRYKER_ACTIVE_MUTANT__) {
+    g.__activeMutant__ = Number(g.process.env.__STRYKER_ACTIVE_MUTANT__);
```
nicojs, line 4: This was the issue. As you can see, the `process.env` variable name was out-dated here (`__ACTIVE_MUTANT__` -> `__STRYKER_ACTIVE_MUTANT__`). 

The code here was instrumented with an older version of the instrumenter.

packages/jest-runner/test/integration/JestTestRunner.it.spec.ts
```diff
@@ -126,6 +125,26 @@ describe(`${JestTestRunner.name} integration test`, () => {
 
       assertions.expectSurvived(runResult);
     });
+
+    it('should be able to let a mutant survive after killing mutant 1', async () => {
```
nicojs, line 129: I've added this test to make sure 2 subsequent tests also succeed.

nicojs:
@Garethp I've just released `v4.0.0-beta.4`. I'm very curious if this yields a performance improvement.

If you want to help further with the improvement of your use case (Jest/react), feel free to help us by finding a project we can use for our performance testing efforts. See also #2434

Garethp:
@nicojs Beta 4 does bring some really great performance improvements, thanks! The memory leak that I've noticed is still dragging down performance, but if I patch that out by restarting the process every 20 runs the performance improvement is easily 6x or more than compared to Beta 3. 

I'll look into bringing a project to you for use of real world testing

Garethp:
Well, I spent all day, but I found out Jest just appears to be leaking for no reason at all. I've logged a [bug report](https://github.com/facebook/jest/issues/10467)

nicojs:
That's very interesting! I will do some checks as well and report my findings there. 

I've got a feeling that there should be a way to prevent memory leaks in Jest. Maybe some sort of `cleanup` we can call between test runs? Will have to investigate.

Issue #2435, opened 2020-08-28
Optionally keep the .stryker-tmp dir


nicojs:
**Is your feature request related to a problem? Please describe.**
While fixing #2315, we've introduced a side effect. Stryker will now always delete the `.stryker-tmp` dir after a run, also when the run is not successful. This can make it hard to debug test runs.

**Describe the solution you'd like**

Either bring back the old functionality or make it configurable with `cleanTempDir: true` or something.

Issue #2315, opened 2020-07-12
Stryker 4.0 beta doesn't close when an error occurs


nicojs:
**Summary**

Stryker 4.0 beta doesn't close when an error occurs. Instead, it keeps hanging with the last line `ERROR StrykerCli There were failed tests in the initial test run.`. This happens because the injectors are not disposed of correctly. 

We should either implement https://github.com/nicojs/typed-inject/issues/24 and use `rootInjector.dispose()`, or implement this:

```ts
    let mutantInstrumenterInjector: Injector<MutantInstrumenterContext> | undefined;
    let dryRunExecutorInjector: Injector<DryRunContext> | undefined;
    let mutationRunExecutorInjector: Injector<MutationTestContext> | undefined;
    try {
      // 1. Prepare. Load Stryker configuration, load the input files and starts the logging server
      const prepareExecutor = this.injector.provideValue(coreTokens.cliOptions, this.cliOptions).injectClass(PrepareExecutor);
      mutantInstrumenterInjector = await prepareExecutor.execute();

      // 2. Mutate and instrument the files and write to the sandbox.
      const mutantInstrumenter = mutantInstrumenterInjector.injectClass(MutantInstrumenterExecutor);
      dryRunExecutorInjector = await mutantInstrumenter.execute();

      // 3. Perform a 'dry run' (initial test run). Runs the tests without active mutants and collects coverage.
      const dryRunExecutor = dryRunExecutorInjector.injectClass(DryRunExecutor);
      mutationRunExecutorInjector = await dryRunExecutor.execute();

      // 4. Actual mutation testing. Will check every mutant and if valid run it in an available test runner.
      const mutationRunExecutor = mutationRunExecutorInjector.injectClass(MutationTestExecutor);
      const mutantResults = await mutationRunExecutor.execute();
      return mutantResults;
    } finally {
      if (mutationRunExecutorInjector) {
        await mutationRunExecutorInjector.dispose();
      } else if (dryRunExecutorInjector) {
        await dryRunExecutorInjector.dispose();
      } else if (mutantInstrumenterInjector) {
        await mutantInstrumenterInjector.dispose();
      }
    }
```

However, this wouldn't be enough, because the `DryRunExecutor` currently creates its own test runner:

```ts
    const testRunnerInjector = this.injector
      // ...
      .provideFactory(coreTokens.testRunnerPool, createTestRunnerPool);
[...]

PR #2444, opened 2020-08-29
feat: cleanTempDir

kmdrGroch:
fixes: #2435


kmdrGroch:
@nicojs something like this? 🤔

packages/core/src/utils/TemporaryDirectory.ts
```diff
@@ -12,10 +12,12 @@ import { deleteDir } from './fileUtils';
 export class TemporaryDirectory implements Disposable {
   private readonly temporaryDirectory: string;
   private isInitialized = false;
+  private readonly removeDuringDisposal: boolean;
 
   public static readonly inject = tokens(commonTokens.logger, commonTokens.options);
   constructor(private readonly log: Logger, options: StrykerOptions) {
     this.temporaryDirectory = path.resolve(options.tempDirName);
+    this.removeDuringDisposal = options.cleanTempDir ?? true;
```
nicojs, line 20: This `??` isn't needed right? Since it has a default value.

nicojs (changes requested):
Seems good. One remark.

I've been thinking. Probably the best user experience is to leave the temp dir when an error occurred by default (our old behavior). For example, when `cleanTempDir` is true, only delete it when mutation testing was successful. 

To do this we would need to communicate the fact that an error occurred from `Stryker`'s `catch`. 

@kmdrGroch do you agree? If you need help, I could also implement this with a commit here.

kmdrGroch:
Updated:
- now the functionality should be the same as it was before #2315 with a possibility to overwrite :)

nicojs:
I think the old behavior is still not there,. The `dispose` method of the `TemporaryDirectory` is called no matter what (error or no error), see https://github.com/stryker-mutator/stryker/blob/6885e16fad7699ba93e6ebbbf0755c7d98c50c5a/packages/core/src/Stryker.ts#L59.

kmdrGroch:
right! I am a dummy :D

nicojs:
>  If you want removing all disposes, i think what should be done is adding "enableDisposal" option or sth like this 🤔

No, we need to `dispose`, otherwise Stryker hangs forever 🙄. `dispose()` makes sure all child processes are closed and nodejs is allowed to exit.

What I meant was something like this:

Inside Stryker.ts

```ts
// 1. Prepare. Load Stryker configuration, load the input files and starts the logging server
      const prepareExecutor = loggerProvider.provideValue(coreTokens.cliOptions, this.cliOptions).injectClass(PrepareExecutor);
      const mutantInstrumenterInjector = await prepareExecutor.execute();

      try {
        // 2. Mutate and instrument the files and write to the sandbox.
        const mutantInstrumenter = mutantInstrumenterInjector.injectClass(MutantInstrumenterExecutor);
        const dryRunExecutorInjector = await mutantInstrumenter.execute();

        // 3. Perform a 'dry run' (initial test run). Runs the tests without active mutants and collects coverage.
        const dryRunExecutor = dryRunExecutorInjector.injectClass(DryRunExecutor);
        const mutationRunExecutorInjector = await dryRunExecutor.execute();

        // 4. Actual mutation testing. Will check every mutant and if valid run it in an available test runner.
        const mutationRunExecutor = mutationRunExecutorInjector.injectClass(MutationTestExecutor);
        const mutantResults = await mutationRunExecutor.execute();

        return mutantResults;
      } catch (error) {
       // Mark current run as failed, so tempoary directory can choose if it needs to clean or not.
        mutantInstrumenterInjector.resolve(coreTokens.temporaryDirectory).markRunErrored();
        throw error; // rethrow
      }
```

nicojs:
Is this clear @kmdrGroch ? I don't see another way, other than adding a parameter to the `dispose` and alter `typed-inject` to accomondate this. Not sure yet if I would want to add that.

kmdrGroch:
I don't quite get it 😕

nicojs:
I had some time, so implemented it.

What I did:

* Made sure to only clean the temp dir when Stryker succesfully executed (it's quite simple, please have a look 👍)
    ```ts
    catch (error) {
      const log = loggerProvider.resolve(commonTokens.getLogger)(Stryker.name);
      log.debug('Not removing the temp dir because an error occurred');
      mutantInstrumenterInjector.resolve(coreTokens.temporaryDirectory).removeDuringDisposal = false;
      throw error;
    }      
    ```
* Made `true` the default value (so after a succesful run, the `.stryker-tmp` dir is gone.
* I've added `--cleanTempDir` as a cli option
* I've documented it in the readme.md

nicojs (approved):

Issue #2330, opened 2020-07-21
Get control over dependencies


kmdrGroch:
Right now, the only support for dependencies we get is Dependabot. Which unfortunatelly has its limitations. So far, I have noticed it doesn't check packages deeply, so while we try maintaining packages in root directory, other (like 10) packages have VERY outdated dependencies (just check some logs while running E2E [we have over 3000 minor issues { says NPM }] throughout the application).
Things that should be done (the faster the better [I know we can do it anytime, but this anytime was several years ago now ;) ]):
1. Configure dependabot so it checks package.json inside packages, e2e etc.
2. FInally update types! Guys we are using 5.0.5 for sinon (THERE IS 9x!)
3. Update the rest of packages *if possible
4. Change ^ to ~ for all packages [or even ""] (unless it has to be "" (linting) or "^" (node, peerDeps)) {or use lockfiles?} - it will require more merging, yes, but on the other hand we will have better control over packages and smaller chance of "breaking change" during minor release (like NYC once had and probably several other packages) and better support over other issues (look #2278)
5. Decide if we want Node@CurrentX or Node@MinimumSupportedX, since its quite inconsistent - 14x v 10x

nicojs:
Thanks for opening this issue! 

> Configure dependabot so it checks package.json inside packages, e2e, etc.

I don't think this is as simple as that. Inside `e2e/test`, there are projects that need their own maintenance lifecycle. For example, angular projects can't simply update dependencies to `rxjs` or `typescript`, they need to wait for the next major version. So what we need to do is to schedule a recurring issue for that every 6 months. 

However, I think the `e2e` project itself should be able to version automagically with dependabot. Most projects use the dependencies of the `e2e` project itself to run, I think only 4 projects don't, so that would already provide a big improvement.

> FInally update types! Guys we are using 5.0.5 for sinon (THERE IS 9x!)

Types should be updated with dependabot like everything else. `@types/sinon` is an exception here. It's mostly https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36019 that is holding is back. I'm willing to give it another shot after our v4 release. I've created #2337 for this purpose.

> Change ^ to ~ for all packages

I agree we should use `~` for everything _except_ our own dependencies (`@stryker-mutator/api`, `@stryker-mutator/util`, etc). Can we add a linting rule for this? 

> or use lockfiles?

[...]

PR #2445, opened 2020-08-30
change dependencies version tags

kmdrGroch:
fixes: #2330


kmdrGroch:
partially fixes 2330 (there are still some deps with couldnt be updated... like @types/sinon), but updating will be in other PR. This one is just to make most of them `~`

nicojs:
Awesome, thanks!!

 > btw, it seems lerna doesnt change versions in peerDependencies

What do you mean? peerDependencies other than `@stryker-mutator/***` dependencies?

kmdrGroch:
no no, only @styker-mutator peerDependencies are not changed (At least i think so)

Issue #2163, opened 2020-04-24
Allow to mutate files in place


nicojs:
**Is your feature request related to a problem? Please describe.**

There is a class of projects that do not play nice with Stryker's way of mutating files inside the sandbox (a subdirectory inside `.stryker-tmp`). We can spend a lot of time debugging the issues, going through the test runner issue tracker to create an issue, creating PR's/making noise to please fix our use case. Or we can allow files to be mutated in place.

List of issues that might be helped by such an option:
#2147
#1564
#717 
#1847
#1957
#2185
#2274 

**Describe the solution you'd like**

I would like to allow Stryker to mutate files "in-place". Meaning to _not create a sandbox_, but instead mutating the _actual input files_. I want mutation switching (#1514) to be implemented first, that way we can keep using parallel processes (since all mutants are in the code, we don't need to write every mutant to disk one by one).

We should be careful to warn users of the consequences. Some thoughts:

* Enable this new functionality with a new cli flag `--inPlace`, or using a config file `"inPlace": true`, default is `false`.
* When enabled, make a copy of the source code in a temp dir. Preferably in the os's temp dir instead of the current working directory.
* After instrumentation, copy the instrumented files to the current working directory, instead of the sandbox file.
* Set back the file's original content after Stryker is done.
* Use a [node exit hook](https://nodejs.org/api/process.html#process_event_exit) to really make sure we allow us to set original files back.

**Describe alternatives you've considered**

What we've been doing until now is saying to everyone to fix the issue for us.

nicojs:
If you're looking at this issue after the fact: it is released in Stryker 4.4, see https://stryker-mutator.io/docs/stryker/configuration/#inplace-boolean

PR #2446, opened 2020-08-30
feat(typescript): Disable type checking

### LINKED ISSUE #2438: Discussion: TypeScript compile errors with mutation switching
See #1514 .

In a mutation switching world, we will create (typescript/flow) type errors. This is by design; it would be almost impossible to prevent it (we would need a team and community at least as large as https://github.com/Microsoft/TypeScript).

I don't think [flow](https://flow.org/) will be an issue because flow code is transpiled without type checking (using a babel plugin). However, TypeScript is a different beast all together..

There are a couple of approaches I've tried:

## Attempt 1: Transpile only

Running the typescript transpiler with configuration similar to [ts-node's --transpile-only](https://github.com/TypeStrong/ts-node#options-via-tsconfigjson). 

However, after a little investigation, I learned that `transpileOnly` only works with [`--isolatedModules`](https://www.typescriptlang.org/tsconfig#isolatedModules). This means that only a subset of typescript will be supported (although a best practice IMHO). We can't rely on this for all our users.

The same goes for skipping type checking by using [babel](https://devblogs.microsoft.com/typescript/typescript-and-babel-7/); it won't support all use cases.

## Attempt 2: Skip Typechecking

With a little fudging in the [TypeScript compiler api](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API) you can get a long way. It would allow us to basically transpile all typescript projects while skipping type checks. See https://github.com/microsoft/TypeScript/issues/29651#issuecomment-616113744 for an example.

However, this approach falls short when you consider more typescript use cases. Calling the TypeScript compiler directly is just _one_ of the ways typescript compilation might take place. A non-extensive list of ways to compile typescript:
* ts-node
* angular-cli
* vue-cli
* react-scripts
* karma-typescript
* webpack with ts-loader
* webpack with awseome-ts-loader
* jest-ts
* rollup with typescript plugin
* Probably 100s more of tools/plugins 

## Solution: // @ts-nocheck

The solution I ended up with is to use the [`// @ts-nocheck`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#-ts-nocheck-in-typescript-files) directive on top of each file (js or ts, since js files can also result in type errors). It works from TypeScript 3.7 onward and will work with all use cases (as long as you're using a fairly recent TypeScript version).

_Note that errors can also be produced by test files, even though they're not mutated. Mutating production code can change the return type, for example `function foo() { return 'bar'; }` is mutated to `function foo() { activeMutant === 1? {} : return 'bar'; }`, which changes it's return type from `string` to `string | undefined`, which probably results in a type error somewhere in your test files_

### Problem 1: Still type checking...

However, this presented me with another issue. Even with `@ts-nocheck`, a file can _still provide type errors_, this is because other directives can result in errors, for example [@ts-expect-error](https://github.com/microsoft/TypeScript/issues/39427), but `@ts-check` has the same result. 

That's why I decided to remove all comments from JS and TS files. We're using the [strip-comments](https://www.npmjs.com/package/strip-comments) package for that, which uses regular expressions to identify the comments. 

You can configure this with 2 settings:

```js
  "sandbox": {
    "stripComments": "**/*+(.js|.ts|.cjs|.mjs)?(x)",
    "fileHeaders": {
      "**/*+(.js|.ts|.cjs|.mjs)?(x)": "/* eslint-disable */\n// @ts-nocheck\n"
    }
  }
```

(see readme for specifics)

### Problem 2: significant comments

I _assumed_ stripping comments wouldn't cause a problem, because comments should be comments. However, this is JavaScript we're talking about. This is a list of comments that are significant and we should actually keep if possible:

- [@jest-environment](https://jestjs.io/docs/en/configuration.html#testenvironment-string)
- [@flow](https://flow.org/en/docs/usage/#toc-prepare-your-code-for-flow)
- ... probably more?

Personally, I feel like removing all comments from all files is a shotgun approach. Simply removing only `@ts-xxx` comments would be better, but how would we do that? I don't want to be parsing all input files (seriously, it is already a pain to parse only the files to be mutated) and `strip-comments` doesn't support comment filtering.

### Problem 3: `strip-comments` has known issues

We've [recently found an issue with the `strip-comments` package where it leaves a file in an invalid state](https://github.com/stryker-mutator/stryker/issues/2402#issuecomment-682498851). If this would happen to someone than the only thing he/she can do is to disable the removal of comments altogether (with `"sandbox.stripComments": false`) or only for that file specifically.

### Problem 4: incompatible scenario with jest.

Even though users can customize to fit their use case with `stripComments` and `fileHeaders`, we still don't support all scenarios. For example, running `@stryker-mutator/jest-runner` together with the `jest-ts` jest plugin might result in a situation where you don't want to strip comments (because of `@jest-environment`), but need to strip comments because you're using `@ts-expect-error` somewhere in your tests. 

Any input would be very helpful 😇

@hugo-vrijswijk @brodybits @Lakitna

nicojs:
Disable type checks by inserting `// @ts-nocheck` and removing all other `// @ts-xxx` directives. 

It can be configured with:

```js
{
  "disableTypeChecks": "{test,src,lib}/**/*.{js,ts,jsx,tsx,html,vue}" // configure a pattern or `false` here
  "warnings": {
    "preprocessorErrors": true // enable/disable error logging here (default: true)
  }
}
```

You can disable it completely by setting `disableTypeChecks` to `false`. The default is `"{test,src,lib}/**/*.{js,ts,jsx,tsx,html,vue}"`.

The disabling of type checks is done in the `DisableTypeChecksPreprocessor` class. It calls the function in the `instrumenter` package to do the actual disabling. The reason for this is that files need to be parsed to distinguish between code and comments (`@ts-xxx` directives are only removed when they appear inside comments). It also works with HTML and vue files, so it also supports use cases where type checking is done inside vue files (with the vue-cli). 

Whenever parsing a file results in an error, a warning is logged and mutation testing will still proceed. The warning can be disabled by setting `warnings.preprocessorErrors` to `false`.

This setting replaces both `sandbox.fileHeaders` and `sandbox.stripComments`, since those resulted in a lot of problems (see #2438 ).

Fixes #2438


packages/api/schema/stryker-core.json
```diff
@@ -191,7 +190,21 @@
...
+            {
+              "enum": [
+                false
+              ]
+            },
+            {
+              "type": "string"
+            }
+          ],
+          "default": "**/*+(.js|.ts)?(x)"
```
brodybits, line 207: Simplify - ???

```suggestion
          "default": "**/*.{js,jsx,ts,tsx}?(x)"
```

packages/instrumenter/src/disable-type-checking.ts
```diff
@@ -0,0 +1,47 @@
...
+import { createParser } from './parsers';
+import { AstFormat } from './syntax';
+
+export async function disableTypeChecking(file: File, options: InstrumenterOptions) {
+  const parse = createParser(options);
+  const ast = await parse(file.textContent, file.name);
+  switch (ast.format) {
+    case AstFormat.JS:
+    case AstFormat.TS:
+      return new File(file.name, `// @ts-nocheck\n${removeTSDirectives(file.textContent, ast.root.comments)}`);
```
brodybits, line 15: nit: I find that a bit hard to read and am not so fond of the hardcoding here. How about moving the whole template literal part into a utility function?

nicojs: Done that 👍

packages/api/schema/stryker-core.json
nicojs: I've changed it to `**/*.{js,ts,jsx,tsx,html,vue}`

nicojs:
@brodybits Everything is now ready for review, please have a look 👍

packages/api/schema/stryker-core.json
```diff
@@ -168,19 +168,8 @@
...
-          "additionalProperties": {
-            "type": "string"
-          },
-          "default": {
-            "**/*+(.js|.ts|.cjs|.mjs)?(x)": "/* eslint-disable */\n// @ts-nocheck\n"
-          }
-        },
-        "stripComments": {
-          "description": "Configure files to be stripped of comments (either single line with `//` or multi line with `/**/`. These comments will be stripped after Stryker has instrumented your code with mutants, but before a test runner or build command is executed. This is used to remove any lingering `// @ts-check` or `// @ts-expect-error` comments that interfere with typescript compilation. The default setting allows comments to be stripped from all JavaScript and friend files in your sandbox, you can specify a different glob expression or set it to `false` to completely disable this behavior.",
+        "disableTypeChecking": {
```
nicojs, line 171: Does it still make sense to keep the `sandbox` options? Or should we place this on the root level of `StrykerOptions`?

We might want to add more preprocessor options in the future.

 @brodybits do you have an opinion?

nicojs: I'm leaning towards moving it to the toplevel 🤔

brodybits: I have very mixed feelings. I think keeping the option as part of `sandbox` would improve the abstraction. OTOH, it may add a little extra complexity.

brodybits:
@nicojs I did take a quick look earlier today, may need some more time to take a more careful look. Thanks for your work so far on this!

packages/api/schema/stryker-core.json
nicojs: The thing is, "sandbox" is a term pinned by Stryker. It doesn't really mean anything for the average user. I've discussed introducing a `sandbox.mode` for the #2163 , but maybe that should just be called `--in-place` (boolean). 

Not adding the `sandbox` term in the configuration might make it easier to understand.

packages/api/schema/stryker-core.json
```diff
@@ -191,7 +180,7 @@
               "type": "string"
             }
           ],
-          "default": "**/*+(.js|.ts|.cjs|.mjs)?(x)"
+          "default": "**/*.{js,ts,jsx,tsx,html,vue}"
```
nicojs, line 183: I have second doubts about this default. It might be too broad. For example, when running Stryker on express you get this warning:

```
18:36:16 (20065) WARN DisableTypeCheckingPreprocessor Unable to disable type checking for file "/home/nicojs/stryker/perf/test/express/examples/ejs/views/footer.html". Shouldn't type checking be disabled for this file? Consider configuring a more restrictive "sandbox.disableTypeChecking" settings (or turn it completely off with `false`) ParseError: Parse error in /home/nicojs/stryker/perf/test/express/examples/ejs/views/footer.html (1:0) Unexpected closing tag "body". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags
    at ngHtmlParser (/home/nicojs/stryker/perf/test/express/node_modules/@stryker-mutator/instrumenter/dist/src/parsers/html-parser.js:51:15)
    at async Object.parse (/home/nicojs/stryker/perf/test/express/node_modules/@stryker-mutator/instrumenter/dist/src/parsers/html-parser.js:33:18)
    at async DisableTypeCheckingPreprocessor.disableTypeChecking [as impl] (/home/nicojs/stryker/perf/test/express/node_modules/@stryker-mutator/instrumenter/dist/src/disable-type-checking.js:16:17)
    at async /home/nicojs/stryker/perf/test/express/node_modules/@stryker-mutator/core/src/sandbox/disable-type-checking-preprocessor.js:30:32
    at async Promise.all (index 36)
    at async DisableTypeCheckingPreprocessor.preprocess (/home/nicojs/stryker/perf/test/express/node_modules/@stryker-mutator/core/src/sandbox/disable-type-checking-preprocessor.js:27:30)
    at async MultiPreprocessor.preprocess (/home/nicojs/stryker/perf/test/express/node_modules/@stryker-mutator/core/src/sandbox/multi-preprocessor.js:14:25)
    at async MutantInstrumenterExecutor.execute (/home/nicojs/stryker/perf/test/express/node_modules/@stryker-mutator/core/src/process/2-MutantInstrumenterExecutor.js:24:23)
    at async Stryker.runMutationTest (/home/nicojs/stryker/perf/test/express/node_modules/@stryker-mutator/core/src/Stryker.js:32:44)
18:36:16 (20065) WARN DisableTypeCheckingPreprocessor (disable "warnings.preprocessorErrors" to ignore this warning
```

Maybe a better default would be `{test,src,lib}/**/*.{js,ts,jsx,tsx,html,vue}`?

nicojs: I've implemented this.

packages/api/schema/stryker-core.json
nicojs: I've removed it. Think it is the best option.

nicojs:
@brodybits I'll be merging this as soon as the tests pass, since it is blocking some progress on other stuff. Feel free to still finish you're review and comment (or again a PR), we can always improve later 🙏

brodybits:
I would probably rename the option (and ideally the modules) to disableTypeCheck or disableTypeChecks.

My apologies for the delays.

brodybits:
Thanks! Will you also rename the modules?

nicojs:
Excellent point! Much better. Thanks a lot! And just in time for this PR 👍 

> Thanks! Will you also rename the modules?

Yeah, forgot about that at first, did that right after. We're doing squash merges, so it will look nice and tidy afterward ;)

> My apologies for the delays.

No worries whatsoever. I'm not entitled to your time, I know that there are more important things than this in the world 🤗

Thank you for the review! I'm glad you find naming things just as important as we do 👍

PR #2458, opened 2020-09-04
fix(reporters): correctly report avg tests/mutants

nicojs:
Correctly report the average tests/mutant metric in the clear-text reporter


nicojs:
I thought this one was already working since all unit tests passed. Turned out that the integration with the reporters didn't work correctly. I've implemented a `reporters-e2e` e2e test which specifically tests the clear text reporter.

PR #2467, opened 2020-09-09
Fix/cache clean mem leak

### LINKED ISSUE #2461: Memory leaks in both the mocha runner and the jasmine runner
**Summary**

Both the `@stryker-mutator/jasmine-runner` and the `@stryker-mutator/mocha-runner` use a simple clear of [node's require cache](https://nodejs.org/dist/latest-v14.x/docs/api/modules.html#modules_require_cache) between runs. However, we forgot to include a clear of the [children](https://nodejs.org/dist/latest-v14.x/docs/api/modules.html#modules_module_children) property of the module that was responsible for first loading the test files.

**Example**

This is a piece of example code I used to debug the memory leak:

```js
const log4js = require('log4js');
const { createMochaTestRunner } = require('@stryker-mutator/mocha-runner');
const { createInjector } = require('typed-inject');

const logger = log4js.getLogger();
const mochaRunner = createInjector().provideValue('logger', logger).provideValue('options', {
  mochaOptions: {
    spec: ['test/helpers/**/*.js', 'test/unit/**/*.js']
  }
}).injectFunction(createMochaTestRunner);

async function main() {
  await mochaRunner.init();
  await infinyRun(0);
}

async function infinyRun(n) {
  console.time(++n);
  await mochaRunner.dryRun({ coverageAnalysis: 'off', timeout: 30000 });
  console.timeEnd(n);
  await infinyRun(n);
}

main().catch(err => {
  console.error(err);
  process.exitCode = 1;
});
```

You can run the sample with `--inspect` to be able to debug it in chrome and take memory snapshots. After running with `--inspect`, you can navigate to `chrome://inspect/` and select your js JVM. See resources like https://nodesource.com/blog/memory-leaks-demystified for more info.

I think the jasmine runner has a similar problem.

**NOTE**: The act of adding the chrome profiler to the process is cause for a memory leak, so if you fixed all other leaks, you're left with this:

image

This means that the only thing holding on to a lot for files is the `SourcePositionTableWithFrameCache`. Try to run without `--inspect`, memory issues should be gone.

**Test runner config**

**Stryker environment**

```
+-- @stryker-mutator/core@4.0.0-beta.4
+-- @stryker-mutator/mocha-runner@4.0.0-beta.4
```

Thanks to @hugo-vrijswijk to help debug this.

nicojs:
Fix a memory leak that keeps required files in-memory between test runs. Properly clear the require cache by also clearing [the mocha's/jasmine's module children](https://nodejs.org/api/modules.html#modules_module_children). This makes sure the required files by mocha can be garbage collected.

Fixes #2461

### DIFF
--- a/packages/jasmine-runner/.mocharc.jsonc
+++ b/packages/jasmine-runner/.mocharc.jsonc
@@ -3,5 +3,6 @@
     "test/unit/**/*.js",
     "test/integration/**/*.js"
   ],
-  "file": "test/helpers/setupTests.js"
+  "file": "test/helpers/setupTests.js",
+  "timeout": 10000
 }
--- a/packages/jasmine-runner/src/JasmineTestRunner.ts
+++ b/packages/jasmine-runner/src/JasmineTestRunner.ts
@@ -26,7 +26,7 @@ export function createJasmineTestRunner(injector: Injector<PluginContext>) {
   return injector.provideClass(pluginTokens.directoryRequireCache, DirectoryRequireCache).injectClass(JasmineTestRunner);
 }
 
-export default class JasmineTestRunner implements TestRunner2 {
+export class JasmineTestRunner implements TestRunner2 {
   private readonly jasmineConfigFile: string | undefined;
   private readonly Date: typeof Date = Date; // take Date prototype now we still can (user might choose to mock it away)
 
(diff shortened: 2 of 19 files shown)


kmdrGroch:
Love this! I think ill check number of objects made before and after this change :)

nicojs:
I've also added 2 integration tests (1 for mocha and 1 for jasmine), where I load large objects in files and then run the test runner 50 times. It ran out-of-memory in the past, now passing with flying colors :).

PR #2496, opened 2020-09-25
feat(Initial test run): start timing the test run when logging that the run is starting

### LINKED ISSUE #2477: 4.0.0.Beta does not correctly report the initial test run time
The beta currently does not report the initial test run correctly. For example:

```
10:26:23 (3511) INFO DryRunExecutor Initial test run succeeded. Ran 59 tests in 30 seconds (net 21857 ms, overhead 34 ms).
```

The time here (30 seconds), is the total time until the initial test run completed. It would be better if we start the timer just before we start the test run, for example, when we `Starting initial test run. This may take a while.`

_Note: this is purely cosmetics. The timeout calculation is using the exact initial test run time_

simondel:
Fixes #2477

### DIFF
--- a/packages/core/src/process/3-DryRunExecutor.ts
+++ b/packages/core/src/process/3-DryRunExecutor.ts
@@ -58,6 +58,11 @@ interface Timing {
    * So the time it took to start the test runner and to report the result.
    */
   overhead: number;
+
+  /**
+   * The total time spent (net + overhead) in a human readable format
+   */
+  humanReadableTimeElapsed: string;
 }
 
 function isFailedTest(testResult: TestResult): testResult is FailedTestResult {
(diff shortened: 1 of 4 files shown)


nicojs (changes requested):
Great!!!

Add a test? 🤷‍♂️

nicojs (approved):

simondel:
> 
> 
> Great!!!
> 
> Add a test? 🤷‍♂️

@nicojs How would you go about testing this?

nicojs (changes requested):
We should not use the global timer here. You're working on a fix if I'm not mistaken.

simondel:
> 
> 
> We should not use the global timer here. You're working on a fix if I'm not mistaken.

Yes, we still use the global timer, but we get the time based on the marker that was placed at the start of the test run. This marker was already used to report the `net` and `overhead` ms time.

nicojs (approved):
I've added a unit test for `Timer`

Issue #2429, opened 2020-08-27
Deprecate support of module.export = function(config) {}


kmdrGroch:
With our current support of JS and JSON files I think we don't need support for `module.exports = function(config) {}` in config files.
It only takes more space, and gives no benefits IMO. Just decreases maintainability and adds more, unnecessary functionalities.
It would allow us to refactor ConfigReader, which right now is a little bit messy...

nicojs:
I agree with this, but we are losing functionality here. Implementing it as a function allows users to use the defaults to determine the values.

For example: 

```js
module.exports = function(config) {
  config.fileLogLevel = config.logLevel; // this would no longer be 
}
```

I'm not sure if anyone uses this functionality.

I would like to deprecate it for 1 major version with a deprecation message like:

`DEPRECATED: exporting a function from your stryker config file is deprecated and will be removed in the future. Please export an object or put your configuration in "stryker.conf.json" file. If you think this is a mistake, please report it here <LINK TO ISSUE>`

kmdrGroch:
So am I allowed to take this issue and deprecate it's usage? We could remove it it 4.1.0 or 4.0 release

nicojs:
No, we're planning to release 4.0 as soon as mutation switching is implemented (hopefully within a month or so (knock on wood 🪑🔨)). I would remove it in v5 or even later. We're not sure yet, so don't mention the specific version in the deprecation message, should be fine.

Lakitna:
Is it currently possible to make this an async function?

```js
module.exports = async function(config) {
  // ...
}
```

I see some (niche) use cases for this. If this is currently possible I'd consider removing it feature regression.

kmdrGroch:
I think it is not possible, although you would have to try... in you search for `config.set` in source files you will see how it is made :)

Lakitna:
It took some time, but the function is not awaited: https://github.com/stryker-mutator/stryker/blob/epic/mutation-switching/packages/core/src/config/ConfigReader.ts#L34

So no, the function can not be async.

nicojs:
Yeah, currently not possible. 

We could add support in the future by exporting a promise, or using top level await (supported in native esmodules from nodejs 14).

Lakitna:
This issue is about removing support for having a function, so I suggest not adding support 😄

nicojs:
hahaha yeah, I meant in a separate issue and PR. I should communicate more clearly, thank you! 😁

kmdrGroch:
[...]

PR #2499, opened 2020-09-25
fix(config): deprecate function based config

kmdrGroch:
Export a function from you're config file is deprecated. Please export an object, or use a json file.

Fixes: #2429

### DIFF
--- a/packages/core/src/config/ConfigReader.ts
+++ b/packages/core/src/config/ConfigReader.ts
@@ -30,6 +30,7 @@ export default class ConfigReader {
     const configModule = this.loadConfigModule();
     let options: StrykerOptions;
     if (typeof configModule === 'function') {
+      this.log.warn('Usage of `module.export = function(config) {}` is deprecated. Please use `module.export = {} or JSON file respectively');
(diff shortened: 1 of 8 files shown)


packages/core/src/config/ConfigReader.ts
```diff
@@ -30,6 +30,7 @@ export default class ConfigReader {
     const configModule = this.loadConfigModule();
     let options: StrykerOptions;
     if (typeof configModule === 'function') {
+      this.log.warn('Usage of `module.export = function(config) {}` is deprecated. Please use `module.export = {} or JSON file respectively');
```
nicojs, line 33:
```suggestion
      this.log.warn('Usage of `module.export = function(config) {}` is deprecated. Please use `module.export = {}` or a "stryker.conf.json" file. For more details, see https://stryker-mutator.io/blog/2020-03-11/stryker-version-3#new-config-format');
```

kmdrGroch:
done :)

nicojs:
Thanks! I've also updated the bug report issue template. Will merge when green

PR #2506, opened 2020-09-27
linting: no-else-return

kmdrGroch:
Problem:
- remove `if / else` syntax for returns and throws in favour of simple ifs

Motivation:
using `if else`, especially nested ones, increases the amount of indentation, very often reducing readability. (mostly because you don't know which if and else match). Using ifs in new lines should make it more readable, reduce horizontal space.

The best example of motivation is probably in: packages/util/src/errors.ts file


kmdrGroch:
Also removed file which was nowhere used ;)

Issue #2423, opened 2020-08-26
Support `mutator.excludedMutations`


nicojs:
**Is your feature request related to a problem? Please describe.**
Stryker 3 supports users configuring their `excludedMutations`. We also want to support those in Stryker 4.

**Describe the solution you'd like**
Still collect the mutants, but mark them as `ignored` (and not place them in the code). Report them as `ignored` as well.

**Describe alternatives you've considered**
Not supporting `excludedMutations`, but I want to reach feature parity as much as possible.

Lakitna:
Duplicate of #2396

Just to be clear, this does not include #1472 right?

nicojs:
haha :) thanks! We're getting there.

PR #2513, opened 2020-09-30
feat(instrumenter): add excludedMutations support

### LINKED ISSUE #2396: Reimplement `excludedMutations` with mutation switching
The current `mutator.excludedMutations` is not yet implemented for mutation switching (v4 beta). Let's bring that back

nicojs:
Add support to configure `mutator.excludedMutations`. This was already supported in Stryker@<3, but not yet in Stryker4 beta.

Instead of not generating the mutants, it generates them without placing them in the code. They get reported as `ignored` in your mutation testing report.

image

Also, add error message, ignore reason, and "killed by" test name in the "show more" modal dialog.

image
 
Fixes #2396


nicojs:
Fixes #2423

packages/core/src/process/4-MutationTestExecutor.ts
```diff
@@ -76,6 +77,14 @@ export class MutationTestExecutor {
     };
   }
 
+  private executeIgnore(input$: Observable<MutantTestCoverage>) {
+    const [notIgnoredMutant$, ignoredMutant$] = partition(input$.pipe(shareReplay()), ({ mutant }) => mutant.ignoreReason === undefined);
```
simondel, line 81: Are the `notIgnoredMutant$` and ignoredMutant$` observables arrays? If so, could you make their name plural to show this?

packages/core/src/process/4-MutationTestExecutor.ts
```diff
@@ -76,6 +77,14 @@ export class MutationTestExecutor {
...
+    return {
+      ignoredResult$: ignoredMutant$.pipe(map(({ mutant }) => this.mutationTestReportHelper.reportMutantIgnored(mutant))),
```
simondel, line 83: Could you format this to multiple lines? Right now, it's not immediatly visible that the `notIgnoredMutant$` is outside the pipe operator.

For example:
```ts
    return {
      ignoredResult$: ignoredMutant$.pipe(
        map(({ mutant }) => this.mutationTestReportHelper.reportMutantIgnored(mutant))
      ),
      notIgnoredMutant$,
    };
```

simondel (approved):

packages/core/src/process/4-MutationTestExecutor.ts
kmdrGroch: I do agree with this ^ looks much cleaner

packages/core/src/process/4-MutationTestExecutor.ts
nicojs: No, it is an observable of mutants. It is kind of a convention to name them plural with a `$` instead of `s`

packages/core/src/process/4-MutationTestExecutor.ts
nicojs: Thanks, will do that before merging. Will do that for all observables in this file

PR #2525, opened 2020-10-03
Change serialization

kmdrGroch:
I was looking at the interfaces and I think we could use JSON.stringify instead of serialize. E2E tests pass. Unit won't for now since we are testing `File`. Although I think it is possible to solve it too. Since the only place it can occur is "CallMessage" in args (its any type), so we could just map through args, find if there is File instance, serialize it, and JSON.stringify entire object.
Deserialization would be opposite, JSON.parse and then map serialize on args

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -39,7 +39,6 @@
   },
   "dependencies": {
     "mutation-testing-report-schema": "~1.4.0",
-    "surrial": "~2.0.2",
     "tslib": "~2.0.0"
   },
   "devDependencies": {
--- a/packages/api/src/core/File.ts
+++ b/packages/api/src/core/File.ts
@@ -1,9 +1,7 @@
-import { Surrializable, surrial } from 'surrial';
-
 /**
  * Represents a file within Stryker. Could be a strictly in-memory file.
  */
-export default class File implements Surrializable {
+export default class File {
   private _textContent: string | undefined;
   private readonly _content: Buffer;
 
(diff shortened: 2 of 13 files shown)


kmdrGroch:
Edit: Looking further has shown that we actually are not using any non serializable type in processes, so JSON.stringify / parse are enough to handle it. Based on this, I removed unit tests which were testing functionalities/cases which were not used in our code base.

nicojs:
Thanks for taking a look at our serialization.

Using `JSON.stringify` works in most situations, but there are situations where it doesn't. For example, if a user configures a custom karma configuration inside a stryker.conf.js file with some more complex stuff.

```js
module.exports = {
    karma: {
        config: {
            proxyReq: function(proxyReq, req, res, options) {
              proxyReq.setHeader('Referer', 'https://www.example.com/');
            },
            client: {
              mocha: {
                grep: /fooComponent.*/
              }
           }
        }
    }
}
```

See: https://karma-runner.github.io/5.2/config/configuration-file.html

Same goes for other test runners.

This is sort of a hidden feature, so we might be able to get rid of this feature (obviously our e2e tests don't have anything like this). 

We could also opt for using [`serialize-javascript`](https://github.com/yahoo/serialize-javascript), which is wildely used. We're using it in mocha as well, to serialize state in `--parallel` mode.

kmdrGroch:
Uhh.... could we try at least making this as an option? Like making option:
```js
{
  'files': ...,
  'testRunner': ...,
  'advanced': {
     'serialization': true/false
  }
}
```
and just making: `const serialization = config.advanced.serialization ? surrial : JSON.stringify`,
Something like this

nicojs:
>  Honestly I see no difference between serialize-javascript and surrial. Their basic behaviour is the same, which adds up to the same issues :/

Yes. It's more to improve maintenance than improve performance 🤷‍♂️

PR #2526, opened 2020-10-04
fix(presets): update `init` templates for 4.0 release

### LINKED ISSUE #2521: Make the `stryker init` template projects compatible with Stryker 4

nicojs:
Update the VueJS, react and angular presets for the 4.0 release.

Closes #2521

### DIFF
--- a/packages/core/src/initializer/StrykerConfigWriter.ts
+++ b/packages/core/src/initializer/StrykerConfigWriter.ts
@@ -1,6 +1,6 @@
 import { existsSync, promises as fs } from 'fs';
 
-import { StrykerOptions } from '@stryker-mutator/api/core';
+import { PartialStrykerOptions, StrykerOptions } from '@stryker-mutator/api/core';
 import { Logger } from '@stryker-mutator/api/logging';
 import { commonTokens, tokens } from '@stryker-mutator/api/plugin';
 import { childProcessAsPromised } from '@stryker-mutator/util';
@@ -60,14 +60,14 @@ export default class StrykerConfigWriter {
    */
   public async writePreset(presetConfig: PresetConfiguration, exportAsJson: boolean) {
     const config = {
-      comment: `This config was generated using a preset. Please see the handbook for more information: ${presetConfig.handbookUrl}`,
+      _comment: `This config was generated using a preset. Please see the handbook for more information: ${presetConfig.handbookUrl}`,
       ...presetConfig.config,
     };
 
     return this.writeStrykerConfig(config, exportAsJson);
   }
 
-  private writeStrykerConfig(config: Partial<StrykerOptions>, exportAsJson: boolean) {
+  private writeStrykerConfig(config: PartialStrykerOptions, exportAsJson: boolean) {
     if (exportAsJson) {
       return this.writeJsonConfig(config);
     } else {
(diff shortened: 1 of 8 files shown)


nicojs:
I've updated the presets in this PR. There is however an issue with the vuejs preset.

The VueJS preset is quite old. It actually is not compatible with the vue cli. If you want to get your vue cli project to work with Stryker, your best chance at the moment is to look at our e2e tests (so either `vue-cli-javascript-jest` or `vue-cli-typescript-mocha` depending on which test runner you are using). I think we might want to add better vue support in the future, but especially with mocha, that isn't easy. 

But, what to do with the vue preset in the meantime? 🤷‍♂️

kmdrGroch:
neat! So much code went out

nicojs:
>  E2E tests are working with VUE, and not much needs to be done.

I disagree. Quite a lot needs to be done to make vue work with mocha actually.

1. Create a custom "webpack.test.config.js" with this content:
    ```js
    const conf = require('@vue/cli-service/webpack.config.js');
    conf.entry = {
      tests: ['./test.js']
    }
    module.exports = conf;
    ```
1. Create a custom "test.js" with the following content:
    ```js
    require('./test/setup.js'); // your setup file here.

    // Then we find all the tests.
    const context = require.context('./src', true, /\.spec\.ts$/);
    // And load the modules.
    context.keys().map(context);
    ```
1. Install "webpack-cli" (vue-cli projects don't have this installed out-of-the-box)
1. Create a "stryker.conf.json" with the following content:
    ```json
    {
     "buildCommand": "webpack --config webpack.test.config.js",
     "mochaOptions": {
        "require": [
          "@vue/cli-plugin-unit-mocha/setup.js"
        ],
        "spec": ["dist/js/chunk-vendors.js", "dist/js/tests.js"]
      },
    }
    ```
1. Run stryker with `NODE_ENV=test VUE_CLI_BABEL_TARGET_NODE=true stryker run` (needs to be reworked)

Even with this, it might not work as expected, because there are still differences between Stryker's way of running the tests and vue-cli way of running it.

Basically, vue-cli it is a big mess IMHO. Vue cli uses `mocha-pack` with additional stuff. `mocha-pack` uses `webpack` and `mocha` with additional stuff. No fun for Stryker.

At this point, I feel like we have 3 options.

1. Drop support for Vue CLI during `stryker init` altogether. Just point people to our e2e test setup in docs. 
1. Or do a PR on the vue cli itself to improve the mess they made.

kmdrGroch:
Oww... i havent checked Mocha yet. Only Jest since we have jest tests for vue :/

simondel:
@nicojs Does the situation improve now Vue v3 is live?

nicojs:
Nope they're still using the mochapack spaghetti

kmdrGroch:
> > I think we should just make another E2E
> 
> What's wrong with the existing one?

"Quite a lot needs to be done to make vue work with mocha actually."
Maybe we both misunderstand each other. :x What i am saying is that we should make it easy to go from `E2E mocha-vue` to `user app` (small changes in configs and it works)

nicojs:
I've updated the vuejs guide: https://github.com/stryker-mutator/stryker-handbook/blob/master/stryker/guides/vuejs.md#vuejs and I've updated the vue templates that are generated when a user chooses mocha. 

I've managed to remove step 2: "Create a custom "test.js" with the following content: ", but I'm still not happy with the result. At least it _does work_ with a standard vue-cli project a.t.m. 

IMO we can release v4 with this... I guess.

nicojs:
> What i am saying is that we should make it easy to go from E2E mocha-vue to user app (small changes in configs and it works)

I tried my best, but I cannot bend steel with my bare hands 😑

Issue #2417, opened 2020-08-25
Performance benchmark: Express


Lakitna:
As discussed here https://github.com/stryker-mutator/stryker/issues/1514#issuecomment-678246993

To get a better feel for the performance impact of changes in Stryker we should add some more benchmarks. 

For this benchmark, we'll use https://github.com/expressjs/express.

This is a simple one for a straightforward node case:

- Common JS with no instrumentation
- Well tested (1147 tests)
- Decent size (1800 lines of source code, 4064 lines total)
- Mocha test framework

### Stryker@3.3.1

| % score | # killed | # timeout | # survived | # no cov | # error | duration 
|---------|----------|-----------|------------|----------|---------|-----|
|   98.21 |     1432 |       492 |         35 |        0 |     155 | 00:08:15

### Stryker@4.Beta.3

| % score | # killed | # timeout | # survived | # no cov | # error | duration 
|---------|----------|-----------|------------|----------|---------|-----|
|  89.01  |   1541 |     201 |        207 |       8 |    68 | 00:04:17

With `"maxConcurrentTestRunners": 8`:

| % score | # killed | # timeout | # survived | # no cov | # error | duration 
|---------|----------|-----------|------------|----------|---------|-----|
|   88.50 |    1623 |      109 |         217 |        8 |     68 | 00:05:55

Lakitna:
I've tried to get a basic idea for how long this benchmark would take (on my pc anyway) and I actually ran into a very obscure bug.

It's so obscure that I'm not even sure if it should be fixed. I've detailed it in #2418

nicojs:
Thanks a lot for wanting to add the benchmark! And the issue is also a juicy one we should fix. 

Beers for everyone! 🍻

Lakitna:
> Beers for everyone! 🍻

But not yet in file paths! 🍻

nicojs:
The issue should be fixed @Lakitna so feel free to proceed 😉

Do you want me to release a new beta version?

Lakitna:
Awesome!

I'll make it work without a new release. Though it is a shame that we can't make this benchmark work for Stryker@3 as a comparison.

nicojs:
Hmm good point. Maybe we can remove the tests that use the fixture?

nicojs:
Or we can backport this fix, your choice.

Lakitna:
> Maybe we can remove the tests that use the fixture?

[...]

PR #2536, opened 2020-10-07
Add cpu profiling

Lakitna:
As discussed in https://github.com/stryker-mutator/stryker/issues/2417#issuecomment-704842467

### DIFF
--- a/packages/core/src/child-proxy/ChildProcessProxy.ts
+++ b/packages/core/src/child-proxy/ChildProcessProxy.ts
@@ -50,11 +50,11 @@ export default class ChildProcessProxy<T> implements Disposable {
     workingDirectory: string
   ) {
     const execArgv = [];
-    // if (process.argv.includes('--inspect')) {
-    execArgv.push('--cpu-prof');
-    execArgv.push(`--cpu-prof-name=child-process-${n}.cpuprofile`);
-    n++;
-    // }
+    if (process.env.STRYKER_PROFILE) {
+      execArgv.push('--cpu-prof');
+      execArgv.push(`--cpu-prof-name=child-process-${n}.cpuprofile`);
+      n++;
+    }
 
     this.worker = fork(require.resolve('./ChildProcessProxyWorker'), [autoStart], { silent: true, execArgv: execArgv });
     this.initTask = new Task();
--- a/packages/core/src/child-proxy/ChildProcessProxy.ts
+++ b/packages/core/src/child-proxy/ChildProcessProxy.ts
@@ -27,6 +27,7 @@ const BROKEN_PIPE_ERROR_CODE = 'EPIPE';
 const IPC_CHANNEL_CLOSED_ERROR_CODE = 'ERR_IPC_CHANNEL_CLOSED';
 const TIMEOUT_FOR_DISPOSE = 2000;
 
+let n = 0;
 export default class ChildProcessProxy<T> implements Disposable {
   public readonly proxy: Promisified<T>;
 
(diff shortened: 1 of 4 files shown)


packages/core/src/child-proxy/ChildProcessProxy.ts
```diff
@@ -27,6 +27,7 @@ const BROKEN_PIPE_ERROR_CODE = 'EPIPE';
 const IPC_CHANNEL_CLOSED_ERROR_CODE = 'ERR_IPC_CHANNEL_CLOSED';
 const TIMEOUT_FOR_DISPOSE = 2000;
 
+let n = 0;
```
nicojs, line 30: What is this doing here? `n` is never updated.

nicojs (changes requested):
Thanks. Could you give me a small introduction about what `--cpu-prof` does? 😅

packages/core/src/child-proxy/ChildProcessProxy.ts
Lakitna: `n` is updated on 56 and used on 55. Basically, I use it to enumerate the child process output files to check how many of them are missing. Another naming scheme is fine with me.

packages/core/src/child-proxy/ChildProcessProxy.ts
```diff
@@ -48,7 +49,14 @@ export default class ChildProcessProxy<T> implements Disposable {
     additionalInjectableValues: unknown,
     workingDirectory: string
   ) {
-    this.worker = fork(require.resolve('./ChildProcessProxyWorker'), [autoStart], { silent: true, execArgv: [] });
+    const execArgv = [];
+    if (process.env.STRYKER_PROFILE) {
```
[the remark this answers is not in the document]
Lakitna: Yes, we can, I initially had it set up with a flag. The downside here is that we will have to add the flag to the `StrykerCli.ts`, so the option would be exposed to the user.

Lakitna:
In simple terms, `--cpu-prof` calls the V8 profiler where `--prof` uses a Node-specific profiler. The upside of using the V8 one is tooling support.

Running a script with `--cpu-prof` will output a `.cpuprofile` file. This file can be read by Chromes profiler (chrome://inspect -> Open dedicated DevTools for Node -> Profile tab -> load).

image

In the profiler tooling, you can see at what point of the scripts life cycle it was doing what. For example, in the image above you can see that the first part of the script is all `require`s, and then it starts running mutants. In fact, a simple ctrl+f tells me this particular child worker only executed tests for 3 mutants.

packages/core/src/child-proxy/ChildProcessProxy.ts
nicojs: Should we not reuse this `n`? For example, with `--concurrency 4`, we might want to use `1`, `2`, `3`, and `4`, but never `5`. Whenever a child process crashes, it can reuse the number.

Lakitna: No, that'll result in overwritten `.cpuprofile` files.

`n` here is not a good solution by any means, but it is a human-readable approach to unique file names. Doing it sequenced like this allows a human to see how many processes crashed, as crashed processes have no `.cpuprofile` file.

For example:

image

We can see that 1, 3, 6-10, 12-14, and 17 are missing here. These processes must have crashed for some reason.

This insight is the only reason why we need `n`.

packages/core/src/child-proxy/ChildProcessProxy.ts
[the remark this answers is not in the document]
Lakitna: See commit, I've changed things over to the `--cpu-prof` flag on `stryker run`.

nicojs:
I was thinking, maybe it might be a better idea to add `--workerNodeArgs` support. It would be a really low level and simple way to support both `--cpu-prof`, `--prof` and any other arguments, like `--inspect-brk`. 

It would be a simple (dumb) sollution, so users should combine this with `--concurrency`.

It would allow for low maintenance and maximum flexibility.

@Lakitna , what do you think?

nicojs:
This use case will be supported with #2609

Issue #2338, opened 2020-07-28
Unify the casing in all file names


nicojs:
**Is your maintenance request related to a problem? Please describe.**
We've been back and forth with the casing in our file names. We started out with `PascalCase` for all files. Later, we diverged for files that contain only a function, or for files that contain multiple things, to `cammelCase`.

In recent times, we see more and more projects to only use [kebab-case](https://en.wikipedia.org/wiki/Letter_case#Special_case_styles) for files. I've started doing that in the new `instrumenter` package.

**Describe the solution you'd like**
Align with all js projects and use only kebab case.

**Describe alternatives you've considered**
\-

**Additional context**
I don't know if there is a way to automate this process. However, it has the potential to be a huge merge conflict thing. It is a one time pain and I want to do this **AFTER** our 4.0 release.

kmdrGroch:
I think it can be checked in linter... I could setup it later :)

PR #2537, opened 2020-10-07
maintenance(eslint): kebab-case file names 

kmdrGroch:
fixes: #2338
Recommended: clone stryker again - especially on windows, some files might be broken after fetching (or use `git -rm --cached -r ./packages`, `git add .`)


kmdrGroch:
Finally @nicojs :)
Small breaking changes:
- entry point in `core` is `stryker` now not `Stryker`. 
- entry point for cli is `stryker-cli` not `StrykerCli`
- entry points (like core.ts, plugin.ts), might have a little bit different name (like going from Plugin to plugin)

nicojs:
>  Small breaking changes:

I don't think changing private entry points is "breaking". We've never "officially" told people to import `import StrykerCli from '@stryker-mutator/core/src/StrykerCli`. Same goes for Stryker. We might want to add `StrykerCli` and `Stryker` in an `index.ts` and refer that in the `main` of the package.json. After that it is public and we should never change that.

nicojs (approved):
I think this is fine. How did you change the file names? By hand or with a

kmdrGroch:
By hand :) linter only told which files, and then everything by hand. Imports errors too :)
So there is a small chance I misspelled something but we will find it sooner than later

nicojs:
YOLO

Issue #2188, opened 2020-05-12
HTML Report in one single file 


HonzaMac:
**Is your feature request related to a problem? Please describe.**

We are using Jenkins for builds and we are attaching mutation html report as artefact.

![image](https://user-images.githubusercontent.com/3055648/81700718-58c4f080-9469-11ea-8853-cb6917669bce.png)

Problem is, when I click on report I see only that custom elements are not supported. 
Root cause are security policies for downloading/linking files from html file for JS files.

<img width="1866" alt="Screenshot 2020-05-12 at 15 59 58" src="https://user-images.githubusercontent.com/3055648/81701066-ca9d3a00-9469-11ea-9936-d2b7243259a5.png">

User receive errors from loading JS files.

**Describe the solution you'd like**

It would be best, if all 3 required files would be inlined in one html report. 
We would not need any alowance and sending over this report would be far easier then using compressed files.

nicojs:
Thanks @HonzaMac for opening this feature request.

I've tried to merge it all into one file once before. However this poses a problem if your JavaScript source code has a string `'</script>'`, as it will close the html script tag. I didn't want to spent a ton of time investigating it. 

Do you know a way around the issue? Otherwise I guess we could enable it with an optional config setting.

HonzaMac:
@nicojs I think we should go with escaping of that values. 
Do you have in some branch that "merge" into files? I can follow up and fix escaping.

HonzaMac:
Thank you!

nicojs:
Sure thing! We will release a new version later this week.

PR #2540, opened 2020-10-07
feat: single file HTML report

kmdrGroch:
Output the `index.html` file without the need for external resources. This will make using it in CI/CD pipelines.

With this, you can simply open the `index.html` file in your browser, without first needing to extract all files to a directory. This will also make Stryker compatible with the [azure-devops-mutationreport-publisher](https://github.com/stryker-mutator/azure-devops-mutationreport-publisher).

If you were relying on the `reports/mutation/html/bind-mutation-test-report.js` file containing the report JSON, please migrate to the new JSON reporter: https://stryker-mutator.io/docs/stryker/configuration#reporters-string

fixes: #2188


kmdrGroch:
adds single file report alongside existing one. I guess it should be tested(?), but still for now it's only proof of concept

packages/core/src/reporters/html/single-file-template/single-file-template.ts
```diff
@@ -0,0 +1,25 @@
...
+  </script>
+</head>
+<body>
+  <img class="stryker-image" alt="Stryker" src="stryker-80x80.png" style="position: fixed; right: 0; top: 0; z-index: 10">
+  <mutation-test-report-app titlePostfix="Stryker">
+    Your browser doesn't support <a href="https://caniuse.com/#search=custom%20elements">custom elements</a>.
+    Please use a latest version of an evergreen browser (Firefox, Chrome, Safari, Opera, etc).
+  </mutation-test-report-app>
+  <script>
+    document.querySelector('mutation-test-report-app').report = ${JSON.stringify(report)};
```
nicojs, line 21: This actually won't work if you have a `file` in the result which contains a `</script>` because HTML parsing takes precedence over JS parsing. 

I'm pretty sure Stryker.NET solved this by simply splitting the string there and concatenating it. So something like

```js
JSON.stringify(report).replace(/<\/script>/g, '</sc" + "ript>');
```

This is maybe not the best way, and might not work in all circumstances, so I'm very open to suggestions here.

nicojs (changes requested):
Thanks for picking this up!

packages/core/src/reporters/html/single-file-template/single-file-template.ts
kmdrGroch: we can go with this simple solution for now. I mean, users will have two options: Bulletproof multifile and basic one file

nicojs: I found the workaround:

https://github.com/stryker-mutator/stryker-net/blob/e847896228aad54809ad0db0c188ef0a5d7d7d9f/src/Stryker.Core/Stryker.Core/Reporters/JsonReporter/JsonReport.cs#L67

Probably do it the exact same way 🤷‍♂️

nicojs:
I'm actually thinking that we should add this with a configuration option. For big projects, it's kind of a waste to have 2 reports for each run, since file size can rise significantly. 

@simondel what is your take on this? 

1. Make it configurable?
1. If so. what will the default be? What will the configuration option name be? `htmlReporter.singleFile`?
1. If not: force both? Or only 1?

kmdrGroch:
In any case, unfortunatelly I have very little time during next two weeks or so. Mostly university stuff + personal problems, so if you could carry on this issue from this point :)

HonzaMac:
I would prefer to have only one index.html with all-in-one.

nicojs:
@kmdrGroch I'm finishing this one. Want to get it in the 4.2 release 😍

nicojs (approved):

PR #2546, opened 2020-10-09
fix(concurrency): better default for low CPU count

### LINKED ISSUE #2542: Better default `--concurrency` for CI pipelines
In v4, we've changed the default for `--concurrency` (previously called `--maxConcurrentTestRunners`) to `n-1` (where `n` is the number of logical cores). 

The reason for this is that we want to reserve 1 cpu for the main Stryker process to orchestrate the workers (both test runner and checker child processes). This is helpful for dev laptops with a lot of logical cores (for example 16), but on GitHub actions, where the number of cores is 2, it is more or less pointless (the main process won't have a lot to do).

Use `n-1` unless `n <= 4`, in that case use `n`.

Use `n` when `process.env.CI === 'true'`, but I think this might be useful for all PC's.

**Additional context**
@mthmulders experienced a slow down after migrating from Stryker @3 to Stryker @4.

nicojs:
Sets the default `--concurrency` to CPU count when the number of available CPU's <= 4.

Also updates the docs (apparently, `--concurrency` was missing from the readme).

Fixes #2542

### DIFF
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -237,7 +237,7 @@
       "default": []
     },
     "concurrency": {
-      "description": "Set the concurrency of workers. Stryker will always run checkers and test runners in parallel by creating worker processes (note, not `worker_threads`). This defaults to n-1 where n is the number of cpu's available on your machine, unless CPU count is <= 4, in that case it uses the CPU count. This is a sane default for most use cases.",
+      "description": "Set the concurrency of workers. Stryker will always run checkers and test runners in parallel by creating worker processes (note, not `worker_threads`). This defaults to `n-1` where `n` is the number of logical CPU cores available on your machine, unless `n <= 4`, in that case it uses `n`. This is a sane default for most use cases.",
       "type": "number"
     },
     "maxTestRunnerReuse": {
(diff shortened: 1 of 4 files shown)


packages/api/schema/stryker-core.json
```diff
@@ -237,7 +237,7 @@
       "default": []
     },
     "concurrency": {
-      "description": "Set the concurrency of workers. Stryker will always run checkers and test runners in parallel by creating worker processes (note, not `worker_threads`). This defaults to n-1 where n is the number of cpu's available on your machine. This is a sane default for most use cases.",
+      "description": "Set the concurrency of workers. Stryker will always run checkers and test runners in parallel by creating worker processes (note, not `worker_threads`). This defaults to n-1 where n is the number of cpu's available on your machine, unless CPU count is <= 4, in that case it uses the CPU count. This is a sane default for most use cases.",
```
Lakitna, line 240: > the number of cpu's available

'number of CPU threads' would be more accurate. Calling them 'cores' or 'logical cores' works too.

nicojs: Fixed ✅

Good find

PR #2581, opened 2020-10-23
feat: add cosmiconfig for config search + parsing

### LINKED ISSUE #2371: Search for config file in parent folder
When using stryker in a monorepo it's helpful to have a common stryker config in a parent folder which can be overwritten by configs in the sub packages if necessary.
```
├── package.json
├── packages
│   ├── pack1
│   │   ├── package.json
│   │   └── stryker.conf.json
│   ├── pack2
│   │   ├── package.json
│   │   └── stryker.conf.json
│   ├── pack3
│   │   ├── package.json
│   │   └── stryker.conf.json
├── stryker.parent.conf.js
```

However, if I don't want to overwrite it I still have to include an additional package config file which only includes the parent config like this:
```
module.exports = require('../../stryker.parent.conf')
```

It would be great if stryker does automatically look for config files in parent folders if no config can be found in the current working directory.

Tummerhore:
Fixes #2371 by using the popular package cosmiconfig (for example also used by prettier) instead of an own solution for searching config files. This would also easily allow options parsing from package.json.


packages/core/package.json
```diff
@@ -61,6 +61,7 @@
     "ajv": "~6.12.0",
     "chalk": "~4.1.0",
     "commander": "~6.2.0",
+    "cosmiconfig": "^7.0.0",
```
kmdrGroch, line 64: I'd rather prefer "~" instead of "^". We have been trying to eliminate all major versions in packages for long time now ._.

packages/core/src/config/config-reader.ts
```diff
@@ -47,37 +50,45 @@ export default class ConfigReader {
...
+        this.log.info(`Using ${result.filepath}`);
+        this.cliOptions.configFile = result.filepath;
+      } else {
+        this.log.info('No config file specified. Running with command line arguments only.');
+        this.log.info('Use `stryker init` command to generate your config file.');
+      }
+    } else {
       this.log.debug(`Loading config ${this.cliOptions.configFile}`);
-      const configFile = this.resolveConfigFile(this.cliOptions.configFile);
+      const resolvedFilename = this.resolveConfigFile(String(this.cliOptions.configFile));
```
kmdrGroch, line 76: I'm not sure why you would like to use `String()` here since it wasn't used before? Any reasoning for this?

nicojs:
We're closing this PR, since we're not convinced that walking directories up in search of a stryker.conf.* file is the right approach. We'll be implementing the `--baseDir` feature in the future that should allow users to share configuration.

PR #2582, opened 2020-10-23
Add a json reporter

### LINKED ISSUE #1814: JSON Reporter
We are making a tool that requires stryker report snapshots to express mutant rating over time. Historical mutant rating.

Export a stryker report in a json file.

My first take was to use the data from the html-reporter but it was embedding the data in a file and it wasn't nice to take them out.

Tummerhore:
Fixes #1814. The implementation is relatively simple and copies most of the code from the html reporter. Probably this can be refactored at some point so that there is less duplicated logic. But it works for now.

### DIFF
--- a/packages/core/src/reporters/json-reporter.ts
+++ b/packages/core/src/reporters/json-reporter.ts
@@ -7,17 +7,18 @@ import { mutationTestReportSchema, Reporter } from '@stryker-mutator/api/report'
 
 import fileUrl = require('file-url');
 
-import * as HtmlReporterUtil from './html/html-reporter-util';
+import * as ReporterUtil from './reporter-util';
 
-const DEFAULT_BASE_FOLDER = path.normalize('reports/mutation');
-const DEFAULT_INDENTION = 0;
+const INDENTION_LEVEL = 0;
 export const RESOURCES_DIR_NAME = 'strykerResources';
 
 export default class JsonReporter implements Reporter {
-  private _baseDir!: string;
+  private readonly _filePath: string;
   private mainPromise: Promise<void> | undefined;
 
-  constructor(private readonly options: StrykerOptions, private readonly log: Logger) {}
+  constructor(private readonly options: StrykerOptions, private readonly log: Logger) {
+    this._filePath = path.resolve(this.options.jsonReporter.baseDir, this.options.jsonReporter.filename);
+  }
 
   public static readonly inject = tokens(commonTokens.options, commonTokens.logger);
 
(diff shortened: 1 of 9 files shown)


packages/core/src/reporters/json-reporter.ts
```diff
@@ -0,0 +1,62 @@
...
+import fileUrl = require('file-url');
+
+import * as HtmlReporterUtil from './html/html-reporter-util';
+
+const DEFAULT_BASE_FOLDER = path.normalize('reports/mutation');
+const DEFAULT_INDENTION = 0;
+export const RESOURCES_DIR_NAME = 'strykerResources';
+
+export default class JsonReporter implements Reporter {
+  private _baseDir!: string;
```
nicojs, line 17: Please use `_baseDir: string | undefined` here.

packages/core/src/reporters/json-reporter.ts
```diff
@@ -0,0 +1,62 @@
...
+          `No base folder configuration found (using configuration: jsonReporter: { baseDir: 'output/folder' }), using default ${DEFAULT_BASE_FOLDER}`
+        );
+        this._baseDir = DEFAULT_BASE_FOLDER;
+      }
+    }
+    return this._baseDir;
+  }
+
+  private async cleanBaseFolder(): Promise<void> {
+    await HtmlReporterUtil.deleteDir(this.baseDir);
```
nicojs, line 59: I don't think we should clean the base dir here. This is by default the directory that the HTML reporter is also writing to. 

 I think you can simply remove this here.

nicojs (changes requested):
Thanks a lot for this PR! Could you maybe also add some unit tests?

packages/core/src/reporters/json-reporter.ts
```diff
@@ -0,0 +1,62 @@
...
+  public wrapUp() {
+    return this.mainPromise;
+  }
+
+  private async generateReport(report: mutationTestReportSchema.MutationTestResult) {
+    const jsonReportFile = path.resolve(this.baseDir, 'mutation.json');
+    await this.cleanBaseFolder();
+    let indent = DEFAULT_INDENTION;
+    if (this.options.jsonReporter && this.options.jsonReporter.spacing) {
+      indent = this.options.jsonReporter.spacing;
```
nicojs, line 37: In hind sight, I think this feature is a bit expensive for what it does. Expensive in the sense that we need to document and maintain it. In practise, the Json reporter would not be used to show the report to a human (i.e. open in an editor).

Is there a specific reason you would want this feature?

Tummerhore: Do you only mean the indention option or the whole json reporter in general? I think that a json reporter should definitely be supported since there is currently no reporter for easy data exchange. If you mean the indention option only, you are probably right. I just added it because it was a very easy addition and is probably wanted. :)

Tummerhore:
Ok, finally found some time to fix the remarks: Added some unit tests, removed the indention option and adjusted some things you commented on in the analog pr #2601 which were also present in my pr. I've also added the option for a custom filename besides the custom path.

packages/core/src/reporters/json-reporter.ts
Tummerhore: I've now removed the indention option and use the default value (0 = no indention) as a static constant.

nicojs (approved):
This looks great overall! I've got one more remark, please take a look if you have some time. If not, I can merge it in and change it before doing a release. Your choice 👍

nicojs:
Thanks! 👍

dance-cmdr:
Many thanks for the feature @Tummerhore!

PR #2583, opened 2020-10-23
Inital timeout parameter

### LINKED ISSUE #2302: Paramterize timeout for initial test run
Currently the timeout for the inital test run is 5 minutes and if this time is exceeded stryker fails with a rather ugly and not very helpful error message:
```
19:44:46 (8900) ERROR InitialTestExecutor Initial test run timed out! Ran following tests before timeout:
19:44:46 (8900) ERROR StrykerCli an error occurred Error: Something went wrong in the initial test run
    at InitialTestExecutor.validateResult (...\node_modules\@stryker-mutator\core\src\process\InitialTestExecutor.js:87:19)
    at InitialTestExecutor.run (...\node_modules\@stryker-mutator\core\src\process\InitialTestExecutor.js:42:18)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Stryker.runMutationTest (...\node_modules\@stryker-mutator\core\src\Stryker.js:53:38)
```

Because a test run could possibly take longer than a hard coded time limit, I suggest introducing another optional parameter like `initialTimeoutMS`. (Note: this is different to the already existing parameters `timeoutMS` and `timeoutFactor` which are used for the timeout  during the actual mutation testing.)
Even if stryker is meant to be used with fast unit tests it should be _at least documented_ that this maximal time exists. (I didn't find this information anywhere in the docs). And in this context, it would be no big deal to have it parameterized.

BTW: Although the error message claims to show the executed tests so far, it does never show anything listed here. But I'm quite sure that some tests have already finished successfully before reaching the time limit.

Tummerhore:
Fixes #2302 (partly) by introducing a new optional timeout parameter for the initial test run.

### DIFF
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -399,8 +399,9 @@
       "default": 5000
     },
     "dryRunTimeoutMinutes": {
-      "description": "Configure an absolute timeout for the initial test run. (It can take a while.)",
+      "description": "Configure an absolute timeout for the initial test run. (It can take a while, therefore we use minutes as time unit.)",
       "type": "number",
+      "minimum": 0,
       "default": 5
     },
     "tsconfigFile": {
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -398,7 +398,7 @@
       "type": "number",
       "default": 5000
     },
-    "initialTimeoutMIN": {
+    "dryRunTimeoutMinutes": {
       "description": "Configure an absolute timeout for the initial test run. (It can take a while.)",
       "type": "number",
       "default": 5
--- a/packages/api/schema/stryker-core.json
+++ b/packages/api/schema/stryker-core.json
@@ -398,6 +398,11 @@
       "type": "number",
       "default": 5000
     },
+    "initialTimeoutMIN": {
+      "description": "Configure an absolute timeout for the initial test run. (It can take a while.)",
+      "type": "number",
+      "default": 5
+    },
     "tsconfigFile": {
       "description": "Configure the (root) tsconfig file for typescript projects. This will allow Stryker to rewrite the `extends` and `references` settings in this and related tsconfig files in your sandbox. Defaults to `tsconfig.json`. This setting is also used when you enable the `@stryker-mutator/typescript-checker plugin",
       "type": "string",
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -240,9 +240,9 @@ You can *ignore* files by adding an exclamation mark (`!`) at the start of an ex
 <a name="dryRunTimeoutMinutes"></a>
 ### `dryRunTimeoutMinutes` [`number`]
 
-Default: `5`
-Command line: `--dryRunTimeoutMinutes 5`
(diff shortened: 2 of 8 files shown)


packages/api/schema/stryker-core.json
```diff
@@ -398,6 +398,11 @@
       "type": "number",
       "default": 5000
     },
+    "initialTimeoutMIN": {
```
nicojs, line 401: Can this be renamed to `dryRunTimeoutMinutes`?

Reasoning:

* "initial timeout" can be confused with other "timeoutxxx" options.
* "Min" can be confused with "minimum" or something like that.

packages/core/src/stryker-cli.ts
```diff
@@ -86,6 +86,7 @@ export default class StrykerCli {
       )
       .option('--timeoutMS <number>', 'Tweak the absolute timeout used to wait for a test runner to complete', parseInt)
       .option('--timeoutFactor <number>', 'Tweak the standard deviation relative to the normal test run of a mutated test', parseFloat)
+      .option('--initialtimeoutMIN <number>', 'Configure an absolute timeout for the initial test run. (It can take a while.)', parseFloat)
```
nicojs, line 89: Please also add a unit test for this.

packages/core/src/process/3-dry-run-executor.ts
```diff
@@ -127,7 +127,7 @@ export class DryRunExecutor {
   private async timeDryRun(testRunner: TestRunner): Promise<{ dryRunResult: CompleteDryRunResult; timing: Timing }> {
     this.timer.mark(INITIAL_TEST_RUN_MARKER);
     this.log.info('Starting initial test run. This may take a while.');
-    const dryRunResult = await testRunner.dryRun({ timeout: INITIAL_RUN_TIMEOUT, coverageAnalysis: this.options.coverageAnalysis });
+    const dryRunResult = await testRunner.dryRun({ timeout: this.initialRunTimeout, coverageAnalysis: this.options.coverageAnalysis });
```
nicojs, line 130: Can we also add a unit test for this?

nicojs (changes requested):
Thanks! Some small remarks.

packages/core/README.md
```diff
@@ -236,6 +237,15 @@ When using the config file you can provide an array with `string`s
 
 You can *ignore* files by adding an exclamation mark (`!`) at the start of an expression.
 
+<a name="dryRunTimeoutMinutes"></a>
+### `dryRunTimeoutMinutes` [`number`]
+
+Default: `5`
```
nicojs, line 243: It is important that this line has 2 trailing whitespaces (or a `\`). See https://github.github.com/gfm/#hard-line-breaks

The same goes for the other 2 lines below.

nicojs (approved):
Looks great as is 👍. I've got a small remark about the readme. I'll merge it in tomorrow anyway, so feel free to ignore if you don't have the time for it.

packages/core/README.md
Tummerhore: Ah good to know. :) Fixed it, though it seems strange that only the new options use the line break. Maybe this can be unified at some point (didn't want to include it here since this pr is only about the timeout paramter.)

PR #2593, opened 2020-10-30
feat(html-report): reposition stryker image

nicojs:
Position Stryker image on the bottom of the report

image


nicojs:
@wijtserekker Let's fix it like this for now. If you know of better ways, please don't hesitate to implement it in a new PR.

PR #2609, opened 2020-11-06
feat(debugging): allow passing node args to the test runner

### LINKED ISSUE #2505: Improve debugging support
When trying to figure out what's happening in a mutant, I'd like to try and debug the test run

I'd like to have the child process that the TestRunners run in have an `--inspect` or `--inspect-brk` flag when specified. Here's a few things to consider:

 * [ ] Concurrency would be an issue, so we'd need to implement a `inspectTestRunner` flag that would set `concurrency` to `1` by default
 * [ ] Should we pass in any `gc` flags that were passed in to the main Stryker process?
 * [ ] We'd probably need the instrumentor to emit sourceMaps, so that whatever hooks into the debug mode can place break points at the lines that they expect to see, not knowing what the instrumented code will look like
 * [ ] Would it be a good user experience to break for every test for every mutant that hits a specific line of code, if you're only interested in targetting certain mutants?

nicojs:
Add the `--testRunnerNodeArgs` option. With this option, you can pass arguments to the node process that hosts the test runner.

For example, running stryker with `--timeoutMS 9999999 --concurrency 1 --testRunnerNodeArgs --inspect-brk` will allow you to debug the tests running in the dry run or the mutation test run.

Another use case is using `--testRunnerNodeArgs --cpu-prof`, this allows you to analyse the performance of the child process..

Fixes #2505


nicojs:
@Lakitna 
@Garethp 

Feel free to take a look if you have some time.

packages/api/schema/stryker-core.json
```diff
@@ -400,6 +400,14 @@
       "type": "string",
       "default": "command"
     },
+    "testRunnerNodeArgs": {
+      "description": "Configure arguments to be passed as exec arguments to the test runner child process. For example, running Stryker with `--timeoutMS 9999999 --concurrency 1 --testRunnerNodeArgs --inspect-brk` will allow you to debug the test runner child process. See `execArgv` of [`child_process.fork`](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options)",
```
Lakitna, line 404: It's probably a good idea to support and communicate with quotes around the test runner args:

```
--timeoutMS 9999999 --concurrency 1 --testRunnerNodeArgs "--inspect-brk"
```

This allows us to send multiple args like this:

```
--timeoutMS 9999999 --concurrency 1 --testRunnerNodeArgs "--inspect-brk --foo --bar --baz"
```

It also makes it clearer which args are for the main process and which for the test runner.

Lakitna: The latter results in `process.argv` containing:

```
[
  '--timeoutMS',
  '9999999',
  '--concurrency',
  '1',
  '--testRunnerNodeArgs',
  '--inspect-brk --foo --bar --baz'
]
```

So `commander` should be able to handle it without issues.

Lakitna (changes requested):
Don't forget to gracefully exit the child processes as I did here: https://github.com/stryker-mutator/stryker/pull/2536/files#diff-9bd013c2c0d5a7d2bcd8594a6bfb8b4a89b54b3db0d4626b61637f00fd37900d

Without it, things that trigger on `process.exit` will not trigger. The result of this is that, for example, `--cpu-prof` does not output anything.

packages/api/schema/stryker-core.json
nicojs: > The latter results in `process.argv` containing:
> 
> ```
> [
>   '--timeoutMS',
>   '9999999',
>   '--concurrency',
>   '1',
>   '--testRunnerNodeArgs',
>   '--inspect-brk --foo --bar --baz'
> ]
> ```
> 
> So `commander` should be able to handle it without issues.

Well. We currently support splitting on "," as you can see here: 

https://github.com/stryker-mutator/stryker/blob/b2e541badd08e34826bc0ac6338cfb40f2e8565d/packages/core/src/stryker-cli.ts#L23

I've added an example with 2 arguments in the readme: `--testRunnerNodeArgs --inspect-brk,--cpu-prof`. Would this be acceptable?

> Don't forget to gracefully exit the child processes as I did here: https://github.com/stryker-mutator/stryker/pull/2536/files#diff-9bd013c2c0d5a7d2bcd8594a6bfb8b4a89b54b3db0d4626b61637f00fd37900d
> 
> Without it, things that trigger on `process.exit` will not trigger. The result of this is that, for example, `--cpu-prof` does not output anything.

Hmm, interesting. I don't seem to get CPU profile files at all this way. Even with the `finally(() => process.exit());`. I'm using Linux, this kind of behavior will definitly be influenced by OS specific closing details. Needs further investigating. I'm hesitant to drag it into this PR, since it is not directly related.

nicojs:
Question. How could we support the command test runner here? Maybe we shouldn't allow it? At least log a warning would be appropriate I guess.

Lakitna:
If the command runner doesn't use child processes it's kind of moot to support it. Besides, the command runner supports passing arguments by definition.

I can see testing/debugging situations where you might temporarily switch to the command runner. So I would log it at warn for those reasons.

packages/api/schema/stryker-core.json
Lakitna: Splitting on `','` would also work, but splitting on `/\s+/` (= any whitespace of any length) would make it feel a lot more like a command to me. If I didn't read the docs I would try `--testRunnerNodeArgs "--foo --bar"` before trying `--testRunnerNodeArgs --foo,--bar`. Both options will work for me though :)

> this kind of behavior will definitly be influenced by OS specific closing details. Needs further investigating.

If you add this to the PR I'll be able to test it on Windows and MacOS. With you testing on Linux we can cover all bases. This might be out of scope for this PR though, we can also cover these bases separately. Also note that `--cpu-prof` is experimental at this time: https://nodejs.org/api/cli.html#cli_cpu_prof. There might be stuff that doesn't work on every OS because of it.

nicojs: The "problem" with splitting on whitespace characters is that there are exotic cases of command-line arguments that require spaces. Not sure about node args though.

I've chosen `,` to split, because we're already using that for other options, like `--mutate` or `--files` (spaces in file names are a bit more common). I could change that for `--testRunnerNodeArgs` to also except a space. That might make sense as you point out.

Lakitna: Yeah, you're right. There are some exotic cases that can pop up. This is generally the moment where I wonder if I can fix all of this with NPM :)

How about using something like [`string-argv`](https://www.npmjs.com/package/string-argv) or [`spawn-args`](https://www.npmjs.com/package/spawn-args) to prevent the pain of edge cases?

nicojs: I've chosen to split on `" "` (space) for now. I've updated the docs and code. It's a better user experience than splitting on a comma.

Don't want to add an extra dependency if I don't have to. I think splitting on spaces for node-args is fine (don't see a direct reason to support escaped spaces)

nicojs: Thanks for helping this feature!

nicojs:
I've implemented the warning.

```
$ npx stryker run --testRunnerNodeArgs "--inspect"
21:00:16 (7810) INFO ConfigReader Using stryker.conf.json
21:00:16 (7810) WARN OptionsValidator Using "testRunnerNodeArgs" together with the "command" test runner is not supported, these arguments will be ignored. You can add your custom arguments by setting the "commandRunner.command" option.
21:00:16 (7810) INFO InputFileResolver Found 2 of 8 file(s) to be mutated.
[...]
```

packages/api/schema/stryker-core.json
Lakitna: No problem :)

Issue #2620, opened 2020-11-16
Fix maintenance template


kmdrGroch:
**Is your maintenance request related to a problem? Please describe.**
I'm always frustrated when I have to post maintenance thing due to not having nice tool icon 🔧 

**Describe the solution you'd like**
🔧 

**Describe alternatives you've considered**
🔨

nicojs:
More emojis? Yes please! 😁 PR's welcome

PR #2621, opened 2020-11-17
main: fix maintenance template

kmdrGroch:
fixes: #2620 
I hope this will work :c


nicojs:
I think it works

image

PR #2629, opened 2020-11-20
docs: add MAINTAINERS.md

### LINKED ISSUE #2619: Discussion: Maintaining Stryker
I'm really proud of what we've achieved here with the Stryker repository. We've got a vibrant community working together to create an awesome mutation testing framework. We've also been steadily climbing in downloads, with currently over 100k downloads/month. 

With success, also the job of maintaining Stryker is growing. We don't want to fall victim to our own success here. It is time to take the next step. We should increase the number of maintainers. 

However, I don't want to just add maintainers without guidance, structure, code of conduct, description, etc. I think we should come to a consensus and document that in a MAINTAINERS.md. It should explain a couple of things:

1. Where to find the code of conduct?
1. What is a maintainer?
1. What can we expect of a maintainer?
1. How do we reach discissions?
1. How to comment on issues/PRs?
1. What is the scope of Stryker?
  1. What is the scope of Stryker for JS explicitly?
1. What is the release process? When do we release?

I'm jealously looking at [mocha's MAITNAINERS.md](https://github.com/mochajs/mocha/blob/master/MAINTAINERS.md), although we might be able to start with something smaller at first and expand as we see fit.

@kmdrGroch 
@Garethp 

What do you guys think? Since I was thinking of adding you as maintainers. Would you feel inclined to help with this? Maybe we can work together on this? @simondel also might want to pitch in.

simondel:
Fixes #2619

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,176 @@
+# Stryker maintainer handbook
+
+## Introduction
+We've written this document for:
+
+1. Active maintainers of Stryker
+1. Prospective maintainers of Stryker
+1. Anyone curious about how Stryker's maintainers maintain Stryker
+
+The purpose of this document is to _describe our processes_. We want to avoid conflicts and confusion around "unwritten rules". In our opinion, the most straightforward way to address this is to _write them down_. This _also_ happens to be the most straightforward way to change them!
+
+## Code of conduct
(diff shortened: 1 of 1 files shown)


kmdrGroch:
It really looks like mocha maintainers :P

PR #2634, opened 2020-11-26
feat(jest-runner): support coverage analysis

### LINKED ISSUE #2316: Implement coverage analysis for Jest
Since we've implemented mutation switching, we can now implement coverage analysis more easily in jest. 

We would need a way to retrieve the `__mutantCoverage__` object after the dry run. See https://github.com/facebook/jest/issues/7421

We would also need to hook into the [test environment](https://jestjs.io/docs/en/configuration#testenvironment-string)

**Additional context**
Mutation switching: #1514

nicojs:
Add support for coverage analysis `'all'` and `'perTest'` for the Jest test runner.

Mutant coverage reporting works by integrating with the jest test environment. If you're using `node`, `jsdom` or `jest-environment-jsdom-sixteen`, stryker will automatically insert it's own equivalent test environment. If you're using a custom one or override the jest environment on a file-by-file basis, you'll have to do some work yourself.

For example (when using `jsdom`):

```js
/**
 * @jest-environment @stryker-mutator/jest-runner/jest-env/jsom 
 */
```

Or for a custom test environment:

```js
// my-custom-jest-environment.js
const { mixinJestEnvironment} = require('@stryker-mutator/jest-runner');
const NodeEnvironment = require('jest-environment-node');

class MyCustomTestEnvironment extends NodeEnvironment {
  // custom magic here ✨
}

module.exports = mixinJestEnvironment(MyCustomTestEnvironment);
```

If you're doing stuff that's not supported, you'll get a clear error message pointing you in the right direction.

Test filtering is done for `"perTest"` coverage analysis using the [`--testNamePattern`](https://jestjs.io/docs/en/cli#--testnamepatternregex)  feature. Measuring the coverage per test is done by integrating with the underlying [jest test runner](https://jestjs.io/docs/en/configuration.html#testrunner-string). Both `jest-jasmine2` (default) and [`jest-circus`](https://www.npmjs.com/package/jest-circus) are supported.

This PR also adds support for `--bail` support. You can disable this using `{ "jest": { "enableBail": false } }`

Fixes #2316

Still TODO: 

- [x] Add an e2e test for jest coverage analysis
- [x] Add an e2e test for the lower bound of the supported jest versions.
- [x] Add some more unit tests
- [x] Improve error message for when users customize their `@jest-environment`
- [x] Add scenario with a custom `@jest-environment` to the e2e tests
- [x] Get performance test data
- [x] Enable --bail mode.


nicojs:
Just ran the performance tests on the lighthouse project using GitHub actions: 

It gave an improvement of 23%, so went from 3h0min to 2h17min. I still see room for improvement though, because I saw we run with `bail: false`, we should be able to run with `bail: true`.

PR #2635, opened 2020-11-27
docs(usage): add typescript test usage hint

J4Numbers:
### Summary

Add a note for users who may use Stryker with Typescript tests that run against Typescript files.

### Why

This was a gotcha that I faced for a little while when trying to use Stryker on a project using typescript tests which were testing typescript files. The issue was that the `tsconfig.json` file that was used to compile the source code normally was not being used, which led to Stryker being unable to perform an initial run of the test suite due to compilation errors that appeared oddly (I can upload an exampe repository that shows this problem in action if required).

The solution is something that I was already doing for running mocha tests within my project, but I believe that it would be useful to have this usage tip directly within the documentation for an immediate direction of what to do when errors like this occur.

### Why is it here?/Phrased like this?

First draft proposal. If there is a better place for it or it requires rewording, then I can update as required.


nicojs:
I understand, but better might be to write a guide under the `guides` section. I think this can be called "TypeScript project with tsnode". Maybe you can share more of your config there to help other newcomers.

As a side note, you will get better performance by compiling your test files with a `buildCommand`. @kmdrGroch 's BigMath as an example: https://github.com/kmdrGroch/BigMath/pull/214

kmdrGroch:
> I understand, but better might be to write a guide under the `guides` section. I think this can be called "TypeScript project with tsnode". Maybe you can share more of your config there to help other newcomers.
> 
> As a side note, you will get better performance by compiling your test files with a `buildCommand`. @kmdrGroch 's BigMath as an example: [kmdrGroch/BigMath#214](https://github.com/kmdrGroch/BigMath/pull/214)

Now I have to push these changes ;D

J4Numbers:
Second attempt at this. Moved to guides, expounded, and added an entry in the sidebar for it.

J4Numbers:
Can't really replicate the windows failure or run the suite on demand to confirm the test failure.

docs/guides/testing-typescript-with-tsnode.md
```diff
@@ -0,0 +1,39 @@
...
+The following _additional_ items may be required to add to your `stryker.conf.js` file. This guide assumes that you have already generated a Stryker configuration file, so the following properties will need either adding or changing within the existing configuration.
+
+```js
+/**
+ * @type {import('@stryker-mutator/api/core').StrykerOptions}
+ */
+module.exports = {
+  buildCommand: 'tsc --build path/to/your/tsconfig.json',
+  commandRunner: {
+    command: 'npm run test:ts',
```
nicojs, line 25: I would prefer to have a guide using the mocha test runner. Reasons are:

1. A nodejs project will probably use either jest or mocha. Jest does its own transpiling, so makes sense to focus on mocha.
2. Using a test runner plugin will dramatically enhance performance because they can use coverage analysis.

Leaving it like this would mean to create a `"scripts": { "test:ts": "..." }` script in your package.json file and this guide doesn't specify that part currently. You can look at [mocha-ts-node](https://github.com/stryker-mutator/stryker/tree/master/e2e/test/mocha-ts-node) e2e test for an example.

Maybe it makes sense to create 2 chapters in this guide? One per test runner (jest vs mocha). You might want to mention that jest doesn't need the use of ts-node at all.

docs/guides/testing-typescript-with-tsnode.md
```diff
@@ -0,0 +1,39 @@
...
+    'glob/for/your/typescript/source/files',
+    'glob/for/your/typescript/test/files',
+    'path/to/your/tsconfig.json/file',
+  ],
+};
+```
+
+In addition, you may be required to create a script in your `package.json` file to run this configuration with the following:
+
+```bash
```
nicojs, line 37: Please add the entire example. Also, add the use of use [`cross-env`](https://www.npmjs.com/package/cross-env) so that this example also works on windows.

```json
{
  "scripts": {
    "test:mutation": "cross-env TS_NODE_PROJECT=path/to/your/tsconfig.json stryker run path/to/your/stryker.conf.js"
  }
}
```

nicojs (changes requested):

docs/guides/testing-typescript-with-tsnode.md
J4Numbers: Filled in a more complete example of a `package.json` file.

docs/guides/testing-typescript-with-tsnode.md
J4Numbers: Second attempt at explaining the configuration process, now with specific references to mocha and jest.

nicojs:
@J4Numbers Sorry for the long wait. I used your text as a starting point, but then rewrote a lot of it and added a bunch of stuff. Hope you're OK with that.

If you're happy I can merge this in.

J4Numbers:
Looks fine to me. Definitely feels like a mucch more well-rounded explanation than the one I had before.

Thanks.

nicojs:
I'm thinking... maybe we can make it a general "NodeJS" guide, as I think it also covers babel projects.

nicojs (approved):

nicojs:
🎉 https://stryker-mutator.io/docs/stryker/guides/nodejs

Issue #2646, opened 2020-12-07
Don't mutate string concatenations with the aritmatic operator


nicojs:
**Is your feature request related to a problem? Please describe.**
Currently, we mutate string concatenations with the [arithmetic operator mutator](https://stryker-mutator.io/docs/mutation-testing-elements/supported-mutators#arithmetic-operator).

```js
"foo" + "bar"
```

Is mutated to:

```js
"foo" - "bar"
```

This is all in all not that helpful, as such expressions result in `NaN`. You do want to mutate these strings, as you want to make sure they are tested. However, the [string literal mutator](https://stryker-mutator.io/docs/mutation-testing-elements/supported-mutators/) already does a great job at it.

**Describe the solution you'd like**

As soon as 1 of the 2 operands is a string literal, don't mutate `+` into `-`,

![image](https://user-images.githubusercontent.com/1828233/101349027-f0b67b80-388c-11eb-8eda-0c3e80af1a1b.png)

**Describe alternatives you've considered**
-

**Additional context**
-

Garethp:
As far as I'm aware I think you only need to traverse your way down the left path of the tree. If there's a string on the right hand side, but nothing on the left hand then it could still be an arithmetic operator. For example:

`'a' + 3 + 2` will result in `a32`, but `3 + 2 + 'a'` will result in `5a`. It leaves a gap when you have instances of 

```js
const a = 'a';
const b = 'b';
console.log(a + b + 'c');
```

where you won't be able to tell if `a + b` is a string concatenation or an arithmetic operation, but it'll still help to optomise cases where we can

nicojs:
> As far as I'm aware I think you only need to traverse your way down the left path of the tree

I don't think so. What I propose here, is a 'shallow' search. In other words: `'a' + 3 + 2` would still be mutated in `'a' + 3 - 2`. 

> where you won't be able to tell if `a + b` is a string concatenation or an arithmetic operation

Yes, you are correct. This is why we originally decided not to differentiate. 

I don't want to have too much logic inside mutators. What I propose here, is  "best-effort". If it is not immediately visible if it's a string concatenation or not, then we mutate. It is at least better than what we have now, and we can add more exceptions if they occur in actual production code.

Garethp:
This should be resolve by #2648. We can re-open this if we find that solution isn't suitable

PR #2648, opened 2020-12-07
fix(instrumentor): Do not run the arithmetic operator mutator on obvious string concatenations

Garethp:
As discussed in #2646, this is a shallow "Best-effort" approach to not mutating obvious string concatenations

### DIFF
--- a/packages/instrumenter/src/mutators/arithmetic-operator-mutator.ts
+++ b/packages/instrumenter/src/mutators/arithmetic-operator-mutator.ts
@@ -34,8 +34,9 @@ export class ArithmeticOperatorMutator implements NodeMutator {
     }
 
     const stringTypes = ['StringLiteral', 'TemplateLiteral'];
+    const leftType = node.left.type === 'BinaryExpression' ? node.left.right.type : node.left.type;
 
-    if (stringTypes.includes(node.right.type) || stringTypes.includes(node.left.type)) {
+    if (stringTypes.includes(node.right.type) || stringTypes.includes(leftType)) {
       return false;
     }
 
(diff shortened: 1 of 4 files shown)


packages/instrumenter/src/mutators/arithmetic-operator-mutator.ts
```diff
@@ -28,7 +28,17 @@ export class ArithmeticOperatorMutator implements NodeMutator {
...
   }
 
-  private isSupported(operator: string): operator is keyof typeof ArithmeticOperators {
-    return Object.keys(this.operators).includes(operator);
+  private isSupported(operator: string, node: types.BinaryExpression): operator is keyof typeof ArithmeticOperators {
+    if (!Object.keys(this.operators).includes(operator)) {
+      return false;
+    }
+
+    const stringTypes = ['StringLiteral', 'TemplateLiteral'];
```
Garethp, line 36: On further thought, this should probably be a whitelist not a blacklist. It's a bit weird, but something like:

```
const log = (thingToLog) => console.log(thingToLog);
const concat = 1 + log;
console.log(concat)
```

Will result in `"1(thingToLog) => console.log(thingToLog)"`. What do you think @nicojs, should we do a whitelist of things that actually make sense for an ArithmeticOperator, or keep it simple and only make exceptions for strings and template strings?

nicojs: I think a whitelist can become big and hard(er) to maintain. I personally think this is fine to start with. It's better than what we have now.

nicojs (approved):
LGTM.

PR #2669, opened 2020-12-23
bump(deps-dev): bump eslint and its plugins

kmdrGroch:
closes: #2662


kmdrGroch:
It is a super strange issue... For some reason node_modules in these packages dont get stryker packages installed - but it seems to be lerna/typescript problem not eslint... So i have no clue why is it happening

nicojs:
Might be that we were just in the process of updating the mutation-testing-elements dependencies and that repo famously [doesn't follow semver](https://github.com/stryker-mutator/mutation-testing-elements#versioning) 😲 

I'll have a look. 👀

PR #2709, opened 2021-01-22
feat(regex-mutator): smart regex mutations

nicojs:
This PR adds support for smart regex mutations to Stryker. 👽

Some examples:

```ts
/\d{4}\s[a-Z]{2}/;
// Mutates to:
/\d\s[a-Z]{2}/; // => Quantifier removal
/\D{4}\s[a-Z]{2}/; // => Predefined Character Class Negation
/\d{4}\S[a-Z]{2}/; // => Predefined Character Class Negation
/\d{4}\s[a-Z]/; // => Quantifier removal
/\d{4}\s[^a-Z]{2}/; // => Character Class Negation
```

Stryker will identify regex literals: `/\d{4}\s[a-Z]{2}/` as well as clear regex string literals in a `RegExp` constructor: `new RegExp('\\d{4}\\s[a-Z]{2}')`. After that it uses the awesome [Weapon-regeX](https://github.com/stryker-mutator/weapon-regex/) library to generate mutants based on the regex pattern. These mutants are then placed inside the code with regular mutation switching practises.

Weapon regex supports mutation levels. Currently, Stryker only introduces regex mutants of mutation level 1 in your code. We might be making this configurable in the future.

Note: this mutator might have a significant impact on your mutation score, depend on whether or not you use regexes a lot and how well you've been testing them 😉. 

You can opt-out of using this mutator by excluding it via stryker.conf.json:

```json
{
  "mutator": {
    "excludedMutations": ["Regex"]
  }
}
```

### DIFF
--- a/packages/instrumenter/src/mutators/regex-mutator.ts
+++ b/packages/instrumenter/src/mutators/regex-mutator.ts
@@ -1,36 +1,63 @@
 import * as types from '@babel/types';
 import { NodePath } from '@babel/core';
-import { mutate } from 'weapon-regex';
+import * as weaponRegex from 'weapon-regex';
 
 import { NodeMutation } from '../mutant';
 
 import { NodeMutator } from '.';
 
-// function isObviousRegexString(path: NodePath<types.StringLiteral>){
-//   if(path.parentPath.isCallExpression() && path.parentPath.isNewExpression())
-// }
+/**
+ * Checks that a string literal is an obvious regex string literal
+ * @param path The string literal to checks
+ * @example
+ * new RegExp("\\d{4}");
+ */
+function isObviousRegexString(path: NodePath<types.StringLiteral>) {
+  return (
+    path.parentPath.isNewExpression() &&
+    types.isIdentifier(path.parentPath.node.callee) &&
+    path.parentPath.node.callee.name === RegExp.name &&
+    path.parentPath.node.arguments[0] === path.node
+  );
+}
+const weaponRegexOptions: weaponRegex.Options = { mutationLevels: [1] };
+
+function mutatePattern(pattern: string): string[] {
+  if (pattern.length) {
+    try {
+      return weaponRegex.mutate(pattern, weaponRegexOptions).map((mutant) => mutant.pattern);
+    } catch (err) {
+      console.error(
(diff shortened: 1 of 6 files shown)


wijtserekker:
Looks very clean! Nice to see Weapon regeX found a home in stryker 😄. I am a little sad that you mutate only on level 1, but I guess users otherwise drown in mutants by default.

packages/instrumenter/src/mutators/regex-mutator.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+function isObviousRegexString(path: NodePath<types.StringLiteral>) {
+  return (
+    path.parentPath.isNewExpression() &&
+    types.isIdentifier(path.parentPath.node.callee) &&
+    path.parentPath.node.callee.name === RegExp.name &&
+    path.parentPath.node.arguments[0] === path.node
+  );
+}
+const weaponRegexOptions: weaponRegex.Options = { mutationLevels: [1] };
```
Nhaajt, line 23: The mutation levels are designed so that the tests get more specific as the mutation level increases. If a test suite can kill the mutants at some level, then it should also kill all mutants at lower mutation levels. So if you want to keep it simple with only one level, then go for level 3.
Other than that, it looks awesome 🎉

nicojs: Thanks for the feedback!

I know this is true, but it would also generate a lot more mutants. That means a bigger hit on performance as well as it representing a bigger piece of the mutation score pie. This first inclusion of the `Regex` mutator is just the beginning. I want to first see how our users react first before I go all out.

In the end, I want to implement mutation levels Stryker-wide. So each mutator can use them and this is a way for us to add new mutators without a big hit on performance.  Stryker.NET already started with this, but they currently use their gut-feeling to choose which mutator provides which mutant level. I'm preparing a new graduation assignment that should investigate this further and implement this as a proof of concept.

I hope you're not disappointed. As you can see, we're definitely planning to implement level 2 and level 3, but with this first release I want to add value for users of Stryker fast and improve on that later.

Nhaajt: Not at all, it's totally understandable 😉

PR #2714, opened 2021-01-25
feat(package): restructure package internals

nicojs:
- Redirect typescript output to "dist" directories
- No longer compile "api" package root ts files (reduce complexity)
- Enable typescript `--strict` mode
- Fix 2 (small) bugs that resulted from uninitialized properties
- Use default import syntax for commonjs modules instead of wildcard import syntax (eg. ~`import * as path from 'path'`~ _`import path from 'path'_),
- Use named exports instead of a mix of default exports and named exports everywhere.
- Force named exports using an eslint rule.

If you're using deep-imports from `@stryker-mutator/core` or friends, they will have to be updated. Note: as a general rule: don't rely on deep imports from Stryker, as they might break at any moment.

```diff
- import { default as HtmlReporter } from '@stryker-mutator/core/src/reporters/html/html-reporter';
+ import { HtmlReporter } from '@stryker-mutator/core/dist/src/reporters/html/html-reporter';
```


kmdrGroch:
Is it ready to be checked @nicojs ?

nicojs:
@kmdrGroch this is ready for review now. You might want to focus on the eslint and typescript config stuff, nearly all files have changes 🤷‍♀️

packages/api/check.js
```diff
@@ -0,0 +1 @@
+module.exports = require("./dist/src/check")
```
kmdrGroch, line 1: Should JS and .D.TS files be on github tho? 🤔

packages/api/core.ts
```diff
@@ -1,11 +0,0 @@
-export { default as File } from './src/core/file';
```
kmdrGroch, line 1: typescript files are removed? 🤔

packages/api/test/unit/core/file.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { expect } from 'chai';
 import { deserialize, serialize } from 'surrial';
 
-import { File } from '../../../core';
+import { File } from '../../../src/core';
```
kmdrGroch, line 4: I don't quite like pathing to src from tests / packages

packages/core/package.json
```diff
@@ -2,13 +2,13 @@
   "name": "@stryker-mutator/core",
   "version": "4.4.0",
   "description": "The extendable JavaScript mutation testing framework",
-  "main": "src/index.js",
+  "main": "dist/src/index.js",
```
kmdrGroch, line 5: dist/src/index? so is it dist or src ;)

packages/core/src/child-proxy/child-process-proxy.ts
```diff
@@ -27,10 +27,10 @@ const BROKEN_PIPE_ERROR_CODE = 'EPIPE';
 const IPC_CHANNEL_CLOSED_ERROR_CODE = 'ERR_IPC_CHANNEL_CLOSED';
 const TIMEOUT_FOR_DISPOSE = 2000;
 
-export default class ChildProcessProxy<T> implements Disposable {
+export class ChildProcessProxy<T> implements Disposable {
   public readonly proxy: Promisified<T>;
 
-  private readonly worker: ChildProcess;
+  private readonly worker: childProcess.ChildProcess;
```
kmdrGroch, line 33: I like this change, provides more information what it is, especially in `fork`

packages/core/src/di/plugin-loader.ts
```diff
@@ -63,8 +63,8 @@ export class PluginLoader implements PluginResolver {
       if (typeof pluginExpression === 'string') {
         if (pluginExpression.includes('*')) {
           // Plugin directory is the node_modules folder of the module that installed stryker
-          // So if current __dirname is './@stryker-mutator/core/src/di' so 4 directories above
-          const pluginDirectory = path.dirname(path.resolve(__dirname, '..', '..', '..', '..', pluginExpression));
+          // So if current __dirname is './@stryker-mutator/core/dist/src/di' so 4 directories above
+          const pluginDirectory = path.dirname(path.resolve(__dirname, '..', '..', '..', '..', '..', pluginExpression));
```
kmdrGroch, line 67: even more dots 😓

packages/core/src/initializer/stryker-initializer.ts
```diff
@@ -1,17 +1,17 @@
-import * as child from 'child_process';
+import child from 'child_process';
```
kmdrGroch, line 1: can't we keep it consistent? childProcess for example

packages/core/src/reporters/progress-reporter.ts
```diff
@@ -1,10 +1,10 @@
...
 
-import ProgressBar from './progress-bar';
-import ProgressKeeper from './progress-keeper';
+import { ProgressBar } from './progress-bar';
+import { ProgressKeeper } from './progress-keeper';
 
-export default class ProgressBarReporter extends ProgressKeeper {
-  private progressBar: ProgressBar;
+export class ProgressBarReporter extends ProgressKeeper {
+  private progressBar?: ProgressBar;
```
kmdrGroch, line 7: so progressBar is optional? 🤔 potentially dangerous, unless you know what you are doing

packages/core/test/helpers/test-utils.ts
```diff
@@ -6,6 +6,6 @@ export function sleep(ms = 0) {
   });
 }
 
-export function resolveFromRoot(...pathSegments: string[]) {
-  return resolve(__dirname, '..', '..', ...pathSegments);
+export function resolveFromRoot(...pathSegments: string[]): string {
```
kmdrGroch, line 9: resolve doesn't return string by default?

packages/instrumenter/testResources/app.component.ts.out.snap
```diff
@@ -0,0 +1,34 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
kmdrGroch, line 1: what happened? 🤔

packages/jest-runner/src/messaging.ts
```diff
@@ -8,14 +8,14 @@ class State {
   }
 
   public coverageAnalysis: CoverageAnalysis = 'off';
-  private mutantCoverageHandler: MutantCoverageHandler;
+  private mutantCoverageHandler?: MutantCoverageHandler;
```
kmdrGroch, line 11: again optional? 🤔

packages/mocha-runner/test/unit/mocha-test-runner.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import * as Mocha from 'mocha';
+import Mocha from 'mocha';
```
kmdrGroch, line 2: Mocha? not mocha?

packages/test-helpers/src/test-injector.ts
```diff
@@ -1,16 +1,16 @@
...
 import { Injector, createInjector } from 'typed-inject';
 
 import * as factory from './factory';
 
 class TestInjector {
-  public pluginResolver: sinon.SinonStubbedInstance<PluginResolver>;
-  public options: StrykerOptions;
-  public logger: sinon.SinonStubbedInstance<Logger>;
-  public injector: Injector<PluginContext>;
+  public pluginResolver!: sinon.SinonStubbedInstance<PluginResolver>;
```
kmdrGroch, line 10: !? are they possibly undefined?

perf/tasks/install.ts
```diff
@@ -1,7 +1,7 @@
-import * as fs from 'fs';
-import * as minimatch from 'minimatch';
-import * as path from 'path';
-import * as execa from 'execa';
+import fs from 'fs';
+import minimatch from 'minimatch';;
+import path from 'path'
+import execa from 'execa';;
```
kmdrGroch, line 4: what happened here? 😆

perf/tasks/run-perf-tests.ts
```diff
@@ -1,9 +1,9 @@
-import * as fs from 'fs';
-import * as path from 'path';
-import * as execa from 'execa';
+import fs from 'fs';
+import path from 'path'
+import execa from 'execa';;
```
kmdrGroch, line 3: same here

packages/api/check.js
nicojs: Great question!

I would actually like to have it as a *.ts file, but that would mean we would need to compile those files _in place_ (not to the `dist` directory), because I want this import to work: `import { File } from '@stryker-mutator/api/core'` and that can only be done with a file called "core.js" inside api (on node < 12.7).

When we drop support for node < 12.7 we can start using [`exports`](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_exports) and remove these root files altogether.

I don't want to make an exception for some files to be compiled in place, because that also means making an exception in our clean script and making the output files hidden in vscode config (i don't like hidden files 😢)

Do you agree with this? In short: we should be able to remove these files once we can use `exports` in package.json.

packages/api/core.ts
nicojs: Yes, see my precious remark.

packages/api/test/unit/core/file.spec.ts
nicojs: Ah, you mean just import from `'../../../core'`? Yes that would be possible here because of the new `d.ts` files.

packages/core/package.json
nicojs: Yes. So `src` is compiled to `dist/src`  and `test` is compiled to `dist/test`. If you really want `src` to be compiled to `dist` we would need to add the tests inside the `src` directory. I don't really mind this that much.

packages/core/src/di/plugin-loader.ts
nicojs: haha yes. No way around that I fear. I don't want to rely on `process.cwd` because one could run Stryker from a different directory. The only other thing we can use is `__dirname` and this is a deep file.

Luckily our e2e tests cover this, so it is tested.

packages/core/src/initializer/stryker-initializer.ts
nicojs: yes definitely. Good point.

packages/core/src/reporters/progress-reporter.ts
nicojs: True, it is dangerous, but your reasoning the wrong way around. It was always optional because we didn't set the value in the constructor or inline here. However, we never got warnings when we used it without a null-check because typescript didn't warn us. 

By enabling [`--strictPropertyInitialization`](https://www.typescriptlang.org/tsconfig#strictPropertyInitialization) I now got a compile error that I fixed by adding the `?`. This triggered further type errors in the places we used it without a null-check, that's how I fixed a small bug :).

packages/core/test/helpers/test-utils.ts
nicojs: Yes it does, I wanted to make it a bit more explicit I guess.

packages/instrumenter/testResources/app.component.ts.out.snap
nicojs: Good point, snapshots shouldn't change. Will revert.

packages/jest-runner/src/messaging.ts
nicojs: Yeah, the same reasoning as before. This is because I enabled `--strictPropertyInitialization`.

packages/mocha-runner/test/unit/mocha-test-runner.spec.ts
nicojs: It's a constructor function (or `class`) 🤷‍♂️

packages/test-helpers/src/test-injector.ts
nicojs: Yes, this is because of `--strictPropertyInitialization` again. But this time I used a `!` (not null assertion) instead of a `?`, because I know for sure this can only _theoretically_ be undefined and will always be defined in practice because I call `reset` directly inside this file (and it's a singleton).

perf/tasks/install.ts
nicojs: Midnight hacking 😅 

Will fix

nicojs:
Thanks for your great feedback @kmdrGroch . I've commented on each remark and fixed some issues.

packages/core/package.json
kmdrGroch: I haven't noticed we have src and test in dist. If it is like this, it is ok i believe :)

packages/core/test/helpers/test-utils.ts
kmdrGroch: In my experience declaring return type can be harmful and dangerous. This time it is a simple one liner with return, but even in this case people can break it... believe me...

kmdrGroch:
The only thing thats left is just making it green again :D

e2e/test/exit-prematurely-dry-run-fails/verify/verify.ts
```diff
@@ -1,6 +1,5 @@
-import { promises as fs } from 'fs';
+import { promises as fsPromises } from 'fs';
 import chai, { expect } from 'chai';
-import { it } from 'mocha';
```
kmdrGroch, line 3: I think thats an issue

nicojs:
@kmdrGroch I had to revert the imports in the api tests.

```ts
import { TestStatus } from '../../../src/test-runner'; // -> this works
import { TestStatus } from '../../../../test-runner'; // -> this compiles, but at runtime won't resolve, because the root dir is one directory more up
import { TestStatus } from '../../../../../test-runner'; // -> this would resolve at runtime, but doesn't compile 🤷‍♂️
```

packages/api/test/unit/core/file.spec.ts
nicojs: sorry, that doesn't work. See my latest comment

packages/core/package.json
nicojs: What would be consistent?

e2e/test/exit-prematurely-dry-run-fails/verify/verify.ts
nicojs: Importing `it` from mocha is optional. Mocha will make sure `it` is globally available. I've removed it since we're never importing `it`.

nicojs:
Good find with the snapshot file. Snapshot remapping broke somehow, I've implemented it with some clever sourcemap magic.

kmdrGroch:
If you want I can review it once more, but probably tomorrow since today I am busy. Or you could merge it and hope for the best 😅

nicojs:
Living on the edge 😉

If we something after the fact, we will open a new PR.

PR #2734, opened 2021-02-05
Create node_modules symlink in subdirectories, make monorepo test easier

wwwzbwcom:
Hello, in this PR I make `symlinkNodeModulesIfNeeded` flag also create symlink for `node_modules` in subdirectories to make testing for monorepo easier.


packages/core/src/utils/file-utils.ts
```diff
@@ -78,18 +77,44 @@ export function symlinkJunction(to: string, from: string) {
...
+  const dirBfsQueue: string[] = (await fs.promises.readdir(basePath)) ?? [];
+
+  while (dirBfsQueue.length > 0) {
+    const dir = dirBfsQueue.pop();
+
+    if (!dir) {
+      continue;
     }
+
+    if (dir && path.basename(dir) === '.stryker-tmps') {
```
nicojs, line 98: I don't think this exception is needed. 

First of all, the default stryker-tmp name is `.stryker-tmp` (minus the `s`) and second of all, it is overridable by the user (for example `--tempDirName tmp`). But I don't think it is really needed, as long as you make sure to not symlink symlinks.

packages/core/src/utils/file-utils.ts
```diff
@@ -78,18 +77,44 @@ export function symlinkJunction(to: string, from: string) {
...
-    if (parent === basePath) {
-      return null;
-    } else {
-      return findNodeModules(path.dirname(basePath));
+  const nodeModulesList: string[] = [];
+
+  const dirBfsQueue: string[] = (await fs.promises.readdir(basePath)) ?? [];
+
+  while (dirBfsQueue.length > 0) {
+    const dir = dirBfsQueue.pop();
```
nicojs, line 92: To get rid of the "not null assertions" below (`dir && ...`), you can rewrite this while statement:

```ts
let dir;
while (dir = dirs.pop()) {
   // [...]
}
```

packages/core/src/utils/file-utils.ts
```diff
@@ -78,18 +77,44 @@ export function symlinkJunction(to: string, from: string) {
...
-    return nodeModules;
-  } catch (e) {
-    const parent = path.dirname(basePath);
-    if (parent === basePath) {
-      return null;
-    } else {
-      return findNodeModules(path.dirname(basePath));
+  const nodeModulesList: string[] = [];
+
+  const dirBfsQueue: string[] = (await fs.promises.readdir(basePath)) ?? [];
```
nicojs, line 89: I think we should make sure that the `node_modules` is a directory and that it is not already a symlink. We can probably use `withFileTypes: true` to get this information all in one go:

https://nodejs.org/api/fs.html#fs_fspromises_readdir_path_options

packages/core/src/utils/file-utils.ts
```diff
@@ -78,18 +77,44 @@ export function symlinkJunction(to: string, from: string) {
...
+  if (nodeModulesList.length === 0) {
+    const parentDir = path.dirname(basePath);
+    if (parentDir !== basePath) {
+      return findNodeModulesList(parentDir);
+    }
+  }
+
+  return {
+    nodeModulesList,
+    basePath,
```
nicojs, line 118: Why is it needed to return the `basePath` again?

packages/core/test/unit/utils/file-utils.spec.ts
```diff
@@ -22,31 +22,46 @@ describe('fileUtils', () => {
...
+      readdirStub.withArgs(path.dirname(path.resolve(basePath))).resolves([expectedNodeModules]);
+      const actual = await fileUtils.findNodeModulesList(basePath);
+      expect(actual.nodeModulesList[0]).eq(expectedNodeModules);
     });
 
-    it('should return null if no node_modules exist in basePath or parent directories', async () => {
+    it('should return node_modules array in subDirectory of `basePath`', async () => {
+      const basePath = path.resolve('a');
+      const expectedNodeModulesList = [path.resolve('a', 'b', 'node_modules'), path.resolve('a', 'b', 'c', 'node_modules')];
+      readdirStub.withArgs(path.resolve(basePath)).resolves([path.resolve(basePath, 'b')]);
```
nicojs, line 45: `readdir` won't ever return a resolved path. It is probably better to let it resolve `'b'` instead.

nicojs (changes requested):
@wwwzbwcom thanks a lot for this change. This was actually on my wishlist for a while!

I don't know exactly why the e2e tests are failing, would you mind taking a look, it should have to do with your changes.

I also have some requests for changes, please take a look.

HonzaMac:
Would it make sense to add e2e test to validate this sub-sub-directories linking works?

wwwzbwcom:
> Would it make sense to add e2e test to validate this sub-sub-directories linking works?

Ok, let me check them. thks for your review

wwwzbwcom:
> @wwwzbwcom thanks a lot for this change. This was actually on my wishlist for a while!
> 
> I don't know exactly why the e2e tests are failing, would you mind taking a look, it should have to do with your changes.
> 
> I also have some requests for changes, please take a look.

Fixing problems now, thks for your review!

packages/core/src/utils/file-utils.ts
wwwzbwcom: eslint disallow me to write like that, but I switch to a shortter way to do this instaed of "not null assertions"

packages/core/src/utils/file-utils.ts
wwwzbwcom: Yes, thank you for notice me. infact, readdir on file cause the error on e2e test, this is fixed now

packages/core/src/utils/file-utils.ts
wwwzbwcom: I want to get the relative path to project base dir, (because the basePath can change to parent dir of cwd in this function logic).

Now this is not necessary, for stryker run actually can't work in project subdir, and I disallow basePath change to parent (if we change the base path to parent recursively, we might treat `/` as basePath and scan the whole disk for node_modules, which is a huge performance issue)

packages/core/test/unit/utils/file-utils.spec.ts
wwwzbwcom: Yes, sorry for this mistake, I fix this now

packages/core/src/utils/file-utils.ts
wwwzbwcom: Fixed, using the this.options.tempDirName

nicojs:
Thanks a lot! I've cleaned it up a bit. This is allowed by our eslint rules. The problem was that you are not allowed to assign a `string | undefined` to a `string`, since that is shady business 🕶.

```ts
let dir: string | undefined;
while ((dir = dirBfsQueue.pop())) {
}
```

nicojs (approved):

nicojs:
I'm approving this, but I still want to add an e2e test (good point @HonzaMac ) I';; try to do that tonight.

nicojs:
I've added a _very simple_ mono repo example:

```
├── node_modules
├── package-lock.json
├── package.json
├── packages
│   └── app
│       ├── node_modules
│       │   └── foo
│       │       └── index.js
│       ├── src
│       │   └── app.js
│       └── test
│           └── app.spec.js
├── stryker.conf.json
└── tasks
    └── bootstrap.js
```

It fails without this fix and succeeds with the fix 👍

nicojs:
Thanks @wwwzbwcom 🎉

PR #2751, opened 2021-02-16
feature(instrumentor): Allow only running specific mutants

Garethp:
This is a PR to accompany issue #2751. It allows you to define specific mutants through the `--mutate` flag by adding a non-inclusive range for the mutant you want to run. For example, `--mutate src/index.ts:1:3:1:5` would only instrument mutants that start on line 1 column 3 and end on line 1 column 5 (non-inclusive)


nicojs (commented):
I've started a review at my own risk (since you're still busy with it) and got some remarks. Mainly I think `file.js:5:0:20:0` should mutate line 5 up and including line 20. So you declare a range of mutants. I think the way you see it is that it will only mutate that specific mutant itself.

Garethp:
> I've started a review at my own risk (since you're still busy with it) and got some remarks.
It's fine, I was thinking of doing the `--specificMutantsJSON` flag as part of a different PR anyway and move this from WIP to ready for review.

> Mainly I think file.js:5:0:20:0 should mutate line 5 up and including line 20. So you declare a range of mutants. I think the way you see it is that it will only mutate that specific mutant itself.
Only mutating the specific mutant defined and not any inclusive mutants is what I laid out in the issue and was what I thought would be good here. What's the advantage of having an inclusive range rather than a specific match? As long as I can have **some** way to define specific non-inclusive mutant ranges then I'm not bothered. I can move that functionality to `--specificMutantJSON`, but I get the feeling that you're not super keen on that solution either

packages/core/src/config/options-validator.ts
```diff
@@ -97,6 +100,22 @@ export class OptionsValidator {
     }
   }
 
+  private specificMutantsValidation(options: StrykerOptions) {
+    options.mutator.specificMutants = options.mutate
+      .filter((fileToMutate) => RegExp('(:\\d+){4}$').exec(fileToMutate))
+      .map((fileToMutate) => {
+        const fileName = fileToMutate.replace(RegExp('(:\\d+){4}$'), '');
```
kmdrGroch, line 107: consider using `/(:\d+){4}$/` instead. It's equivalent but much faster.

packages/core/src/input/input-file-resolver.ts
```diff
@@ -117,6 +119,14 @@ export class InputFileResolver {
   }
 
   private async expandPattern(globbingExpression: string, logAboutUselessPatterns: boolean): Promise<string[]> {
+    if (RegExp('(:\\d+){4}$').exec(globbingExpression)) {
```
kmdrGroch, line 122: same here. Maybe consider putting this regex in a variable somewhere since it's used everywhere 🤔 easier to maintain in the future i guess

kmdrGroch (changes requested):

packages/api/schema/stryker-core.json
```diff
@@ -204,6 +204,47 @@
             "type": "string"
           },
           "default": []
+        },
+        "specificMutants": {
```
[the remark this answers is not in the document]
Garethp: The instrumentor gets passed the `InstrumenterOptions` section of `StrykerOptions`, rather than the whole thing. I thought it would be better to have the options validator take the data from the `mutate` property and transform it into a well-defined object that gets passed into the instrumentor rather than change it so that the instrumentor gets passed the entire `StrykerOptions` object just so that the instrumtor can try and parse it out of `mutate`.

packages/core/src/input/input-file-resolver.ts
```diff
@@ -119,7 +119,7 @@ export class InputFileResolver {
   }
 
   private async expandPattern(globbingExpression: string, logAboutUselessPatterns: boolean): Promise<string[]> {
-    if (RegExp('(:\\d+){4}$').exec(globbingExpression)) {
+    if (RegExp(/(:\d+){4}/).exec(globbingExpression)) {
```
kmdrGroch, line 122: `/(:\d+){4}/` not `RegExp(/(:\d+){4}/)` ;)

Garethp: Using `globExpression.match(/(:\d+){4}/)` causes an `eslint` error telling me to use `RegExp` instead because `RegExp` is apparently faster?

kmdrGroch: isn't it saying to use /regex/.exec(words)  instead of (words).match(regex)?

Garethp: Done :)

packages/core/src/input/input-file-resolver.ts
```diff
@@ -119,12 +119,12 @@ export class InputFileResolver {
   }
 
   private async expandPattern(globbingExpression: string, logAboutUselessPatterns: boolean): Promise<string[]> {
-    if (RegExp(/(:\d+){4}/).exec(globbingExpression)) {
+    if (/(:\d+){4}$/.exec(globbingExpression)) {
```
kmdrGroch, line 122: is this change intentional?

Garethp: Yup. The specific mutation definition should only occur at the end of the filename/glob expression

packages/api/schema/stryker-core.json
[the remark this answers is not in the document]
nicojs: Sorry, but I would have to disagree with adding `specificMutants` here. Adding it here would mean users can provide specific mutants both in `--mutate` as well as here. This increases our API surface, maintenance, documentation, etc.

I looked into the code, but it seems that the `InstrumenterOptions` are already separate from the `MutatorDescriptor`. In other words, you could provide it to the instrumentor with something like this:

```js
// Instrument files in-memory
const instrumentResult = await instrumenter.instrument(this.inputFiles.filesToMutate, {
  ...this.options.mutator,
  mutationRange: this.inputFiles.mutationRange
});
```

(I think `mutationRange` might be a better word in this context, leaving "specificMutants" to be used as an array of id's in the future).

It would mean adding the `mutationRange` to the `InputFileCollection`, which kind of makes sense.

We could also think a bit more outside the box and specify a class that inherits from `File` the `mutationRange`, but that would mean doing something with object-orientation, which we've kind of moved away from in recent years.

packages/core/src/config/options-validator.ts
```diff
@@ -97,6 +100,22 @@ export class OptionsValidator {
     }
   }
 
+  private specificMutantsValidation(options: StrykerOptions) {
+    options.mutator.specificMutants = options.mutate
```
nicojs, line 104: I can see what you're doing here, but I don't think the parsing should be done here. The validation can move to the `additionalValidation` function, while the parsing should be part of the `InputFileResolver`, where it can provide the range as `mutationRange`.

packages/core/src/input/input-file-resolver.ts
```diff
@@ -117,6 +119,14 @@ export class InputFileResolver {
   }
 
   private async expandPattern(globbingExpression: string, logAboutUselessPatterns: boolean): Promise<string[]> {
+    if (/(:\d+){4}$/.exec(globbingExpression)) {
+      if (hasMagic(globbingExpression)) {
+        throw new StrykerError('Do not use glob patterns with specific mutants on the glob pattern');
```
nicojs, line 124: As mentioned, this could be part of the `OptionsValidator`. Throwing a `ConfigError` is better than a `StrykerError`, because it will show a better readable error message to the end-user, without a stack trace.

packages/instrumenter/src/transformers/babel-transformer.ts
```diff
@@ -22,6 +22,18 @@ export const transformBabel: AstTransformer<AstFormat.JS | AstFormat.TS> = ({ ro
         // Don't mutate type declarations or import statements
         path.skip();
       } else {
+        if (
+          options.specificMutants?.length &&
```
nicojs, line 26: We could optimize performance a bit here by `skip`-ing parts of the AST that fall outside of the mutant range. But that is definitely a nice-to-have

nicojs (changes requested):
Thanks, @Garethp I've taken a more in-depth look. 

Sorry for the delay 😔, There is a lot of stuff on my plate and I sometimes need a trigger, so feel free to do that any time on Slack (as you now did).

I could also pick up the remarks in this PR.  I have some time this week, I planned an open-source day this Friday and will have some time in this weekend as well.

e2e/test/mutate-specific-lines/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "mutate-specific-lines",
+  "description": "A test project for only running specific mutants, rather than all mutants",
+  "scripts": {
+    "test": "stryker run -m src/index.js:3:18:3:26,src/index.js:4:18:4:25",
+    "test:unit": "jest",
+    "posttest": "mocha --require ../../tasks/ts-node-register.js verify/verify.ts"
+  },
+  "devDependencies": {
+    "@types/jest": "~25.2.1",
```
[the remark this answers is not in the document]
Garethp: I don't know at this point. Will they be pulled in automatically? I assumed I needed them because this specific test uses jest, but perhaps I don't understand what dependencies are pulled in where in the e2e tests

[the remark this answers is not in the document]
nicojs: The e2e parent project itself has a big list of dependencies. Specific e2e projects only need their own dependencies to test special cases. For example angular-project, vue-cli project, etc. They generally don't play nice with "hoisted" dependencies (sigh...).

For e2e tests testing stuff within Stryker itself (coverage analysis, reporters, and I think this mutant range feature as well) you can probably get away with a simple project that relies on the parent's dependencies.

Garethp:
I've gone ahead and moved the parsing to happen only in the `input-file-resolver`, and the validation to only occur in the `options-validator`. I haven't implemented `skipping` any code that lies outside of the mutation range, but I'm not even sure how I'd test that.

We'd have to not `skip` over nodes that surround the ranges, so we'd need to keep the current condition and just have another one that does `skip` for anything where **none** of the node falls inside that range, which is what I think you were thinking about? But in that case, the observable output of the new condition wouldn't change, and it would be a purely performance improvement with no way to write a test against it, right?

I think it's a great idea, I'm just not sure how you want to approach it

packages/instrumenter/src/transformers/babel-transformer.ts
Garethp: Great idea, but I'm not sure how I'd go about writing a test for that since the observable output would still the same with or without it

nicojs:
> We'd have to not `skip` over nodes that surround the ranges, so we'd need to keep the current condition and just have another one that does `skip` for anything where **none** of the node falls inside that range, which is what I think you were thinking about? But in that case, the observable output of the new condition wouldn't change, and it would be a purely performance improvement with no way to write a test against it, right?

Exactly right. We should have tests for all those cases you're describing, make sure our code works, then we can introduce the performance enhancement and see if the tests still pass. The performance enhancement doesn't have an observable output, but that's fine, as long as the tests still pass, nothing broke 🤷‍♂️. That's enough for me. It would result in equivalent mutants, which is why I wouldn't go for 100% mutation score.

I actually found a bug... I think:

```ts
              path.node.loc!.start.line >= mutant.start.line &&
              path.node.loc!.start.column >= mutant.start.column &&
              path.node.loc!.end.line <= mutant.end.line &&
              path.node.loc!.end.column <= mutant.end.column
```

This would mean that the range from line 1, column 1, to line 1 column 2 would not fall into the range of line 0, column 0, line 100, column 0.

I'll go ahead and add tests for this and fix it. As well as review your work. Thanks for the hard work and sorry again for my late responses

nicojs:
Ok I think this should be it 😅. This is what I changed:

* Rename `filename` to `fileName`. This is a hot topic within our code base, but since we're using `fileName` currently everywhere I went ahead and changed it.
* Rename `mutationRange` to `mutationRanges`, since it's an array.
* Implement the new mutation range matching algorithm using `locationIncluded` and `locationOverlaps`. I think it looks quite nice 🎨
* Change the mutation range functionality a bit. Stryker will now also files that don't have mutation ranges. I thought this would make the most sense. If you run `stryker run -m "foo.js:0:0:1:0,bar.js" I would personally expect `bar.js` to also be mutated. Thoughts?
* Refactor the regexes to reduce them to 1 uber regex. It uses regex groups to capture the file name, lines and columns: ` /(.*):(\d+):(\d+):(\d+):(\d+)$/`. It's exported and reused by the options validator.
* Changed the e2e test a bit so it also verifies that files without ranges are mutated and files that are not listed at all are not mutated.
* I've documented mutation ranges in `configuration.md`

Some final thoughts: 
* 🤔 should we change the syntax to `foo.js:1:0-10:0`?  That might feel more natural for some users
* 🤔 should we allow for a short range syntax: `foo.js:1-10`? This would be pretty simple to implement:
  * `1-10` is equivalent to `1:0-11:0` (it includes everything though line 10 _excluding_ line 11)
  * `1:4-10` is equivalent to `1:4-11:0`

nicojs (approved):

Garethp:
Thanks for your help, the changes you made are great!

> should we change the syntax to foo.js:1:0-10:0? That might feel more natural for some users
> should we allow for a short range syntax: foo.js:1-10? This would be pretty simple to implement

Both of these ideas are fantastic. I'll try to find some time to do that

Garethp:
@nicojs I've updated the regex to accept the new format as well as optional columns. The documentation should also be updated to reflect it

packages/core/test/unit/input/input-file-resolver.spec.ts
```diff
@@ -190,8 +190,30 @@ describe(InputFileResolver.name, () => {
...
+            column: 2,
+            line: 0,
+          },
+        },
+      ];
+      expect(result.mutationRanges).to.deep.equal(expectedRanges);
+    });
+
+    it('should default column numbers to 0 if not present', async () => {
+      testInjector.options.mutate = ['mute1:0-1'];
```
nicojs, line 216: This should create a range from `0:0` to `2:0` (so I would want to include line 1 here)

nicojs: I've implemented this.

nicojs:
I've changed the implementation a bit. Now the to-line is inclusive when using `a-b` syntax. I'm using `Number.MAX_SAFE_INTEGER` as a default for the column number, sounds like a big enough number to me.

Issue #2574, opened 2020-10-22
Stryker with multiple Jest projects


giacomoforlani:
**Question**
I created an application with multiple Jest projects because I need to tests node and jsdom environments.
When I launch `stryker run`, I encountered this exception:
```
ERROR DryRunExecutor One or more tests resulted in an error:
        Test runner crashed. Tried twice to restart it without any luck. Last time the error message was: Error: Error: Jest: Cannot use configuration as an object without a file path.
Error: Jest: Cannot use configuration as an object without a file path.
    at readConfig (/Users/user/Examples/next-test/node_modules/jest-config/build/index.js:188:13)
```

I saw that problem is due by line `/Users/user/Examples/next-test/node_modules/jest-config/build/index.js:406:32`
`const parsedConfig = await readConfig(argv, projects[0]);`

`readConfig` function expects a string as the second argument but, in my case, `projects` is an array of objects instead of an array of strings.

Anyone have an idea to fix or bypass this exception?

**Environment**
```
 "@stryker-mutator/core": "^4.0.0",
 "@stryker-mutator/jest-runner": "^4.0.0",
 "@types/jest": "^26.0.15",
 "jest": "^26.6.0",
```

**stryker.config.js**
```
/**
 * @type {import('@stryker-mutator/api/core').StrykerOptions}
 */
module.exports = {
  mutate: [
    "**/*.ts",
    "**/*.tsx",
    "!**/*.spec.*",
  ],
  packageManager: "yarn",
  reporters: ["html", "clear-text", "progress", "dashboard"],
  testRunner: "jest",
  coverageAnalysis: "all",
};

```

**jest.config.js**
```
const testPathIgnorePatternsDefault = [
  '<rootDir>[/\\\\](build|docs|node_modules|.next|.stryker-tmp)[/\\\\]'
];

module.exports = {
  roots: [
    '<rootDir>'
  ],
  moduleFileExtensions: [
    'js',
    'ts',
    'tsx',
    'json',
  ],
  transform: {
    '^.+\\.(ts|tsx)$': 'babel-jest',
  },
  transformIgnorePatterns: [
    '[/\\\\]node_modules[/\\\\].+\\.(ts|tsx)$',
  ],
  projects: [
    {
      name: 'jsdom',
      testPathIgnorePatterns: [
        ...testPathIgnorePatternsDefault,
        '/__tests__/node/',
      ],
      testRegex: '.*.spec.*',
      testEnvironment: 'jsdom',
[...]

PR #2780, opened 2021-03-05
fix(jest-runner): Support multi project jest-configs

swist:
This mimics behaviour of jest - if you want to use projects,
you need to pass the config by file. This teaches stryker to do
the same - if you set explicit config path in jest config,
that gets passed into jest and voila, multi project jest works :)

https://github.com/swist/stryker-test is the minimum reproducible
example repo for this

Gives a sensible-ish workaround for #2574 until jest fix their stuff (and it doesn't look like they will, the issue was open for a couple of years, so would be great to merge this instead!)


swist:
This looks like it fell victim to unrelated test flake, jasmine runner tests have sometimes been running out of heap on my local host too.

nicojs:
Thanks, @swist 🐱‍👤, I will need some time to look into this in detail.

swist:
TL;DR - jest needs to read config from file to behave correctly with multi project. This takes stryker modified jest-config and writes it to a specially suffixed `.json` file so that jest can also read the modifications. It doesn't teach stryker how to modify each config yet, I figured this could land in a separate PR

nicojs:
Ok, thanks for the explanation. I've started looking into this. Am I allowed to say: I hate jest for having this `--projects` feature 😔

I'm trying to first get a simple e2e test project with `--projects` working. I was trying to base it on your project, but there seems to be a problem. When I run `npm t`, but I get this error:

```
 ~/github/swist-stryker-test (master)$ npm t

> starter-kit@0.0.1 test /home/nicojs/github/swist-stryker-test
> jest

 PASS   server  server/app.test.js
GET /api 200 2.153 ms - 27
GET / 200 3.245 ms - 167
GET /main.js 200 0.835 ms - 47
GET /foobar 200 1.527 ms - 167
POST /foobar 404 1.283 ms - 146
 PASS   client  client/src/pages/Home.test.js
  ● Console

    console.error
      Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
          at Home (/home/nicojs/github/swist-stryker-test/client/src/pages/Home.js:8:32)
          at Router (/home/nicojs/github/swist-stryker-test/node_modules/react-router/cjs/react-router.js:99:30)

      17 |                      })
      18 |                      .then((body) => {
    > 19 |                              setMessage(body.message);
         |                              ^
      20 |                      })
      21 |                      .catch((err) => {
      22 |                              console.error(err);

      at console.error (node_modules/@testing-library/react/dist/act-compat.js:53:34)
      at printWarning (node_modules/react-dom/cjs/react-dom.development.js:67:30)
      at error (node_modules/react-dom/cjs/react-dom.development.js:43:5)
      at warnAboutUpdateOnUnmountedFiberInDEV (node_modules/react-dom/cjs/react-dom.development.js:23914:9)
      at scheduleUpdateOnFiber (node_modules/react-dom/cjs/react-dom.development.js:21840:5)
      at dispatchAction (node_modules/react-dom/cjs/react-dom.development.js:16139:5)
      at setMessage (client/src/pages/Home.js:19:5)

A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.

Test Suites: 2 passed, 2 total
Tests:       7 passed, 7 total
Snapshots:   0 total
Time:        5.796 s
Ran all test suites in 2 projects.
```

nicojs:
Ok, I've added an e2e test. 

As for the solution direction, I don't like test runners just writing files to the sandbox. This is programmatic for a couple of reasons:
1. Test runners run in parallel on _the same sandbox_. So 2 test runners might be trying to access the same file and get an error.
2. Since test runners share the sandbox, they might be writing different configuration to the file resulting in race conditions (this is currently not the case because `dryRun` and `mutantRun` run in different phases of the mutation testing process, but we might want to change this in the future).
3. When running with `--inPlace` mode, the file will linger after Stryker is done.

I've looked through jest's `readConfig` source code and I don't see an easy fix. It convinced me that it is actually a bug in Jest, rather than a feature. Maybe we can open an issue on their end?

nicojs:
I've opened https://github.com/facebook/jest/issues/11213

swist:
> Ok, I've added an e2e test.
> 
> As for the solution direction, I don't like test runners just writing files to the sandbox. 
Agreed and I thought I had a solution that didn't need this (just pass the original config) but then it turned out that striker sometimes rewrites the jest configs :(

>This is problematic for a couple of reasons:
> 
> 1. Test runners run in parallel on _the same sandbox_. So 2 test runners might be trying to access the same file and get an error.
Is this strictly true? In my testing I get ~(# cpu cores/2)-1 sandboxes. Because the stryker modifications to the config are idempotent I figured it's just fine to timestamp them per sandbox instance

> 2. Since test runners share the sandbox, they might be writing different configuration to the file resulting in race conditions (this is currently not the case because `dryRun` and `mutantRun` run in different phases of the mutation testing process, but we might want to change this in the future).
I guess the solution would be to namespace this then

> 3. When running with `--inPlace` mode, the file will linger after Stryker is done.

Is there any way to add a clean-up hook after each run? Stryker already leaves the sandboxes behind, so from my perspective it wasn't a big deal

> I've looked through jest's `readConfig` source code and I don't see an easy fix. It convinced me that it is actually a bug in Jest, rather than a feature. Maybe we can open an issue on their end?

The issue at jest has been open for a couple of years now, unlikely they will fix it :)

nicojs:
> The issue at jest has been open for a couple of years now, unlikely they will fix it :)

Only 28 days, right? 😉 I've opened a PR, let's see what they do with it.

I agree that it is a small hack, but in my experience, hacks like these tend to linger and pile up. With all respect, but once it is in we will have to maintain it for as long as Jest is being used. Let's hope they agree to merge the PR.

> Stryker already leaves the sandboxes behind

It shouldn't. After a successful run, it is cleaned. Also with `--inPlace` mode Stryker tries its best. See https://stryker-mutator.io/docs/stryker-js/configuration#cleantempdir-boolean

swist:
Hmm, maybe it's the failed runs that leave sandboxes behind. Jest have known about the problem since 2018: https://github.com/facebook/jest/issues/7415

It's linked from stryker already (#2574)

nicojs:
Cool, all the more reasons we should be able to get this merged in jest 💪

nicojs:
My PR fix just got merged in Jest 🃏 https://github.com/facebook/jest/pull/11307

So it should simply work out-of-the-box from the next release. Closing this PR

MishaTyupa:
Hey @nicojs . Thanks for contributing to this project. 
It turns out that stryker js 5.6.1 version and jest runner 5.6.1 are not working with **jest** test path wildcards if you need to match, for instance, ~2k tests.
Here is my package.json
  **"devDependencies": {
    "@stryker-mutator/core": "^5.6.1",
    "@stryker-mutator/jest-runner": "^5.6.1"
  },**
  
  and what I have in test path in my jest.config:
      **testMatch: [
        '<rootDir>/**/test.js',
        '<rootDir>/**/**/*spec.js',
        '<rootDir>/**/**/tests/*spec.js'
    ]**
    
 It injects mutants to all source files successfully and then rin the tests and fails with the following error:
 **Test runner crashed. Tried twice to restart it without any luck. Last time the error message was: Error: Child process [pid 19908] exited unexpectedly with exit code 1 (without signal).**
 
Could you set up a similar project and try to debug and fix such an issue?
Thank you man for all your work on this project,
Best regards,
Mike

nicojs:
Thanks for coming back with this. I'll reopen this issue. We should add an e2e test for jest with a `projects` setup.

nicojs:
Ooops, reopened the PR instead of the issue 😅

PR #2781, opened 2021-03-06
feat(html): update html reporter

nicojs:
Update `mutation-testing-elements` to 1.6.2. Replaces popover and modal dialog with a drawer, see https://github.com/stryker-mutator/mutation-testing-elements/pull/871 for more information.


nicojs:
We can't merge this until we drop support for node 10 🤷‍♂️

PR #2783, opened 2021-03-07
fix(babel-transformer): respect top of the file comments/pragma

### LINKED ISSUE #2704: Babel parse error on flow syntax
**Summary**

I am getting a `BABEL_PARSE_ERROR` on flow type syntax, and I _think_ it's because stryker is adding instrumentation to the top of my files, which means that the `// @flow` pragma is no longer at the top of those files, so the [babel flow plugin](https://babeljs.io/docs/en/babel-plugin-transform-flow-strip-types) no longer runs against those files.

**Stryker config**

```js
{
  packageManager: 'yarn',
  reporters: ['html', 'clear-text', 'progress', 'dashboard'],
  testRunner: 'jest',
  coverageAnalysis: 'perTest',
  disableTypeChecks: false,

  sandbox: {
    fileHeaders: {
      // I think this is ignored?
      '**/*.js': '// @flow\n\n',
    },
  }
}
```

**Stryker environment**

```
├─ @stryker-mutator/api@4.3.1
├─ @stryker-mutator/core@4.3.1
│  ├─ @stryker-mutator/api@4.3.1
│  ├─ @stryker-mutator/instrumenter@4.3.1
│  ├─ @stryker-mutator/util@4.3.1
├─ @stryker-mutator/instrumenter@4.3.1
│  ├─ @stryker-mutator/api@4.3.1
│  ├─ @stryker-mutator/util@4.3.1
├─ @stryker-mutator/jest-runner@4.3.1
│  ├─ @stryker-mutator/api@4.3.1
│  ├─ @stryker-mutator/util@4.3.1
├─ @stryker-mutator/util@4.3.1
```

```
├─ @jest/core@25.4.0
```

**Your Environment**

| ---------------- | -------
| node             | 14
| npm              | 6
| Operating System | macos

**Add stryker.log**

> Will add if it's absolutely necessary

swist:
This means that the stryker mutator hooks are no longer just
prepended to the beginning of file. Instead we also add the comments
which means things like flow should behave correctly

Fixes #2704

### DIFF
--- a/e2e/package-lock.json
+++ b/e2e/package-lock.json
@@ -515,6 +515,23 @@
         "@babel/helper-plugin-utils": "^7.8.0"
       }
     },
+    "@babel/plugin-syntax-flow": {
+      "version": "7.12.13",
+      "resolved": "https://registry.npmjs.org/@babel/plugin-syntax-flow/-/plugin-syntax-flow-7.12.13.tgz",
+      "integrity": "sha512-J/RYxnlSLXZLVR7wTRsozxKT8qbsx1mNKJzXEEjQ0Kjx1ZACcyHgbanNWNCFtc36IzuWhYWPpvJFFoexoOWFmA==",
+      "dev": true,
+      "requires": {
+        "@babel/helper-plugin-utils": "^7.12.13"
+      },
+      "dependencies": {
+        "@babel/helper-plugin-utils": {
+          "version": "7.13.0",
+          "resolved": "https://registry.npmjs.org/@babel/helper-plugin-utils/-/helper-plugin-utils-7.13.0.tgz",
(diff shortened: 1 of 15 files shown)


nicojs:
@swist thanks for your work! I had to take a closer look, so I checked out the code and I ended up fixing an issue. hope you don't mind. 

The issue: you were mutating (no pun intended) the `instrumentationBabelHeader`. That means you changed it for all other files as well (as well as any other mutation test runs inside the same nodejs process). This is possible because the babel AST is a mutable AST by default and the `instrumentationBabelHeader` variable itself wasn't frozen. 

I already burned myself a couple of times on this as well, so I've changed this by using the `deepFreeze` helper method to be sure it will never be mutable again. I've changed your implementation in order to accommodate this change. I also revived the original "should add the global js header on top" test to make sure that scenario still works. 

I found out that all tests pass even without the `root.innerComments`. Any idea what that is all about? Do you agree with removing this?

nicojs:
> I found out that all tests pass even without the `root.innerComments`. Any idea what that is all about? Do you agree with removing this?

If not, please open a new issue or a new merge request. I want to create a new release today and wanted to have this in 😎

PR #2790, opened 2021-03-09
fix(instrumenter): offset location in mutant collector instead of html parser

### LINKED ISSUE #2775: Incorrect results with vue SFC
**Summary**

If I run stryker mutations on vue-file, it returns me completely incorrect results.

Looks like something wrong with line counting because of template block

**Stryker config**

```js
module.exports = {
    plugins: [
        '@stryker-mutator/jest-runner',
    ],
    reporters: [
        'html',
        'progress',
        'event-recorder',
    ],
    testRunner: 'jest',
    coverageAnalysis: 'perTest',
    disableTypeChecks: false,
    mutate: [
        'src/components/core/Button.vue',
    ],
};
```

**Stryker environment**

```
+-- @stryker-mutator/core@4.4.1
+-- @stryker-mutator/jest-runner@4.4.1
```

```
+-- jest@26.6.3
```

Here you can see report files generated by version 3 and version 4. In version 3 it works good
[v3.zip](https://github.com/stryker-mutator/stryker/files/6068465/v3.zip)
[v4.zip](https://github.com/stryker-mutator/stryker/files/6068469/v4.zip)

Djaler:
Fixes #2775


Djaler:
So, here I am, @nicojs

packages/instrumenter/test/unit/parsers/html-parser.spec.ts
```diff
@@ -74,11 +74,6 @@ describe('html-parser', () => {
       const parsed = await parse(html, 'index.html', contextStub as ParserContext);
 
       // Assert
-      const { start, end } = parsed.root.scripts[0].root.loc!;
```
nicojs, line 77: Can we add an assert here for the new `lineOffset`?

packages/instrumenter/src/transformers/mutant-collector.ts
```diff
@@ -16,9 +16,13 @@ export class MutantCollector {
    * @param mutationSpecs the named node mutation to be added
    * @returns The mutant (for testability)
    */
-  public add(fileName: string, mutationSpecs: NamedNodeMutation): Mutant {
+  public add(fileName: string, mutationSpecs: NamedNodeMutation, lineOffset = 0): Mutant {
     mutationSpecs.replacement.end = mutationSpecs.original.end;
     mutationSpecs.replacement.start = mutationSpecs.original.start;
+    if (mutationSpecs.original.loc) {
```
nicojs, line 22: Can we add a unit test for this in `mutant-collector.spec.ts`?

nicojs (changes requested):
Thanks a lot @Djaler. I've got 2 small remarks.

packages/instrumenter/src/transformers/mutant-collector.ts
```diff
@@ -16,9 +16,13 @@ export class MutantCollector {
...
+      mutationSpecs.original.loc.start.line += lineOffset;
```
nicojs, line 23: Wait... we're still mutating the original node here, right? So this is still the same issue, no?  It's probably better to push this `lineOffset` as a property on the `Mutant` ("packages/instrumenter/src/mutant.ts") class and use that value in `toApiMutant` to offset the location.

Also, the `Mutant` has a `range` property. It basically is used as another way to identify the mutant location based on the character index in the file. That also needs to be offset correctly. For example, the `TypeScriptChecker` plugin uses range to place mutants 1-by-1 in the code to validate them on compile errors. If we want to keep supporting `.vue` files with a `<script lang="ts">`, we will need to offset the range correctly as well.

nicojs (changes requested):
Thanks for adding the tests. I've found 2 more issues while looking more closely. Would you mind taking a look if you have the time? ❤

packages/instrumenter/src/transformers/mutant-collector.ts
Djaler: About first part of your message - looks like it works because we mutate exact node, not all

nicojs:
Thanks! I'll have another look tomorrow

packages/instrumenter/src/transformers/mutant-collector.ts
nicojs: Thanks, this looks great.

> About first part of your message - looks like it works because we mutate exact node, not all

Yes, but if 1 node is mutated twice (by 2 mutators for example), the offset would still be applied twice I think.

nicojs (approved):
Code looks great and the issue seems resolved 😎

image

PR #2797, opened 2021-03-11
test: make garbage collector tests more robust

nicojs:


nicojs:
Ran the tests now 5 times and green every time 🍏

PR #2848, opened 2021-04-21
feat(ignore patterns): add "ignorePatterns" config option

### LINKED ISSUE #1593: Add the ability to add gitignored files inside the config: files
### Summary

Here: 
https://github.com/stryker-mutator/stryker/tree/master/packages/core#files-string

I would like to be able to add my translation files that are gitignored. I run a command to update them.

### LINKED ISSUE #2739: Add an option to include additional files

Related to: #2736

When some files ignores by git are needed to run Stryker, there are currently two workarounds:
- use the `inPlace` option to avoid using the sandbox
- use the `files` option to list **all** files that should be copied to sandbox

Another nice option would be to have an option similar to `files` but allowing to specify only **additional** files.

nicojs:
Add the ability to specify `ignorePatterns` in your Stryker configuration. This replaces `files` option which is now deprecated.

By specifying `ignorePatterns` you choose which files should not be copied to the Sandbox directory (inside the `.stryker-tmp` directory). This effectively excludes them from being used during mutation testing.

For example:

```json
{
  "ignorePatterns": ["dist"]
}
```

This will discover all files in your current working directory, _except_ for files matching the `"dist"` pattern. Stryker uses [.gitignore](https://git-scm.com/docs/gitignore) rules to resolve these patterns.

Stryker always adds `['.git', 'node_modules', '/reports', 'stryker.log', '.stryker-tmp']` when resolving these patterns. If you really want them in your sandbox, you can un-ignore them using a `!` prefix in your pattern, for example:

```json
{
  "ignorePatterns": ["!node_modules"]
}
```

The `ignorePatterns` is now also the default way files are discovered. Previously, Stryker used `git ls-files --others --exclude-standard --cached --exclude .stryker-tmp` to discover files by default. This had some limitations:

* ❌ It only worked inside a git repository and you had to have `git` installed. If not, you got an error and had to use `files`
* ❌ If a `.gitignore`'d file is needed to run your tests, you needed to fall back on `files`. 

Specifying `files` is now deprecated, but still works for the time being. Stryker will rewrite your file patterns internally to ignore patterns.

Closes #1593
Closes #2739

BREAKING CHANGE: Stryker will no longer use a git command to determine which files belong to your project. Instead, it will rely on sane defaults. You can change this behavior by defining [`ignorePatterns`](https://stryker-mutator.io/docs/stryker-js/configuration/#ignorepatterns-string).

BREAKING CHANGE: The `files` configuration option is deprecated and will be removed in a future release. Please use [`ignorePatterns`](https://stryker-mutator.io/docs/stryker-js/configuration/#ignorepatterns-string) instead.

This:

```json
{
  "files": ["foo.js"]
}
```

Is equivalent to:

```json
{
  "ignorePatterns": ["**", "!foo.js"]
}
```


nicojs:
@stryker-mutator/stryker-js-and-friends-maintainers What do you think about this? A couple of thoughts:

* I've decided to _not_ go with a `.strykerignore` file in the end. It would take us a lot of time and code to maintain, which I think isn't really worth it for this niche feature (who is going to override `ignorePatterns` anyway? 🤷‍♂️)
* I've chosen the name `ignorePatterns`, which is in line with how [eslint calls it](https://eslint.org/docs/user-guide/configuring/ignoring-code#ignorepatterns-in-config-files), but this could potentially be confusing, since users might expect these patterns to be ignored for mutating... ?
* Should we deprecate "files"? I don't see a reason for users to really use a `files` array. However, since this was the only way to include non-gitignored files in the past, it might be a bit harsh to out-right deprecate it.

nicojs:
Maybe we should rename it to `ignoreFiles`, so it's more in line with `files` itself. However, you might think you can specify a `.strykerignore` or `.gitignore` file there 🤷‍♂️

nicojs:
Ok, I think I know what I want to do.

I think I'll _deprecate_ `files`. If a user defines `files`, I'll log a warning and _rewrite_ them to `ignorePatterns` rules:

```json
{  
  "files": [ "src/**/*.ts", "dist/**/*.js"]
}
```

wordt
```
{
  "ignorePatterns": ["**", "!src/**/*ts", "!dist/**/*.js"]
}
```

simondel:
Changing 
```
{  
  "files": [ "src/**/*.ts", "dist/**/*.js"]
}
```
into
```
{
  "ignorePatterns": ["**", "!src/**/*ts", "!dist/**/*.js"]
}
```
will work, but removing the `ignorePatterns` all together may also work. Should we let the user know that it might be smart to try running without the suggested change?

docs/configuration.md
```diff
@@ -126,21 +126,25 @@ Set the log level that Stryker uses to write to the "stryker.log" file. Possible
...
 
-If you do need to override `files` (for example: when your project does not live in a git repository),
-you can override the files here.
+### `ignorePatterns` [`string[]`]
 
-When using the command line, the list can only contain a comma separated list of globbing expressions.
-When using the config file you can provide an array with `string`s
+Default: `[]`<br />
+Command line: `--ignorePatterns dist,dist-test`<br />
+Config file: `"ignorePatterns": ["dist", "dist-test"]`<br />
```
simondel, line 139: Perhaps we should also add a `coverage` or `reports` folder to this example

docs/configuration.md
```diff
@@ -126,21 +126,25 @@ Set the log level that Stryker uses to write to the "stryker.log" file. Possible
...
+
+Specify the patterns to all files or directories that are not used to run your tests and thus should _not be copied_ to the sandbox directory for mutation testing. Each patterns in this array should be a [`.gitignore`-style glob pattern](https://git-scm.com/docs/gitignore#_pattern_format). 
 
-You can *ignore* files by adding an exclamation mark (`!`) at the start of an expression.
+These patterns are **always ignored**: `['node_modules', '.git', '/reports', '/stryker.log', '.stryker-tmp']`. Because Stryker always ignores these, you should rarely have to adjust the `"ignorePatterns"` setting at all. If you want to undo one of these ignore patterns, you can use the `!` prefix, for example: `['!node_modules']`.
```
simondel, line 143: Why do you specify `reports` and `stryker.log` with a `/` at the front, but not for other files/folders?

packages/api/schema/stryker-core.json
```diff
@@ -282,18 +282,19 @@
...
+        "type": "string"
+      },
+      "default": []
+    },
     "fileLogLevel": {
       "description": "Set the log level that Stryker uses to write to the \"stryker.log\" file",
       "$ref": "#/definitions/logLevel",
       "default": "off"
     },
-    "files": {
```
simondel, line 290: Shouldn't this still be in the schema, but deprecated?

packages/core/package.json
```diff
@@ -66,6 +66,7 @@
     "file-url": "~3.0.0",
     "get-port": "~5.0.0",
     "glob": "~7.1.2",
+    "ignore": "^5.1.8",
```
simondel, line 69: Was the `^` intentional here? Or did you mean `~`?

packages/core/src/input/input-file-resolver.ts
```diff
@@ -119,58 +74,92 @@ export class InputFileResolver {
...
-        is located inside a git repository`
-        ),
-        error
+  private async resolveInputFiles(): Promise<string[]> {
+    const ignoreRules = this.ignoreRules.map((pattern) => new Minimatch(pattern, { dot: true, flipNegate: true, nocase: true }));
+
+    // Inspired by https://github.com/npm/ignore-walk/blob/0e4f87adccb3e16f526d2e960ed04bdc77fd6cca/index.js#L124
+    const matchesDirectory = (entryName: string, entryPath: string, rule: IMinimatch) => {
+      return (
+        matchesFile(entryName, entryPath, rule) ||
```
simondel, line 117: This is quite unreadable as you're combining a lot of matches. It might be more readable if the sets are more on one line (for example `rule.negate && (rule.match(`/${entryPath}`, true) || rule.match(entryPath, true))`

It might also be smart to place the checks in separate variables so you can actually name what's going on

packages/core/package.json
nicojs: Wow, not using it. Will remove 😅

nicojs:
I've added a remark about completely removing "files":

> DEPRECATED. Use of "files" is deprecated, please use "ignorePatterns" instead (or remove "files" altogether will probably work as well). For now, rewriting them as ["**","!src/**/*.js","src/index.js"]. See https://stryker-mutator.io/docs/stryker-js/configuration/#ignorepatterns-string

docs/configuration.md
nicojs: Yeah, good one

docs/configuration.md
nicojs: Because I know for a fact that Stryker places them at the root of your project. For example. I would like the `src/reports` directory that someone might have to be included. Starting with a `/` means to only match when it exists in the root of the cwd.

packages/api/schema/stryker-core.json
nicojs: As explained during the community meeting, I don't want people to use it. The rewrite to `ignorePatterns` is being done _before_ the options are validated. This is in line with other deprecated options.

packages/core/src/input/input-file-resolver.ts
nicojs: Since we're using `prettier` I can't play around with the formatting. Separate variables would work, but it would technically cost performance since all matches would always be done (instead of stopping at the first failing match).

I've added a third `matchesXxx` helper function:

```js
const matchesDirectoryPartially = (entryPath: string, rule: IMinimatch) => {
      // @ts-expect-error Missing overload in type definitions. See https://github.com/isaacs/minimatch/issues/134
      return rule.match(`/${entryPath}`, true) || rule.match(entryPath, true);
    };
```

This helps a lot I think.

simondel:
LGTM!

PR #2868, opened 2021-05-03
feat(report): report test states

### LINKED ISSUE #2464: For a failure case in the final report, have a way to identify the tests that covered that code
In order to know what tests need to be improved, it would be useful to know what tests actually covered a particular mutation.

### LINKED ISSUE #2636: Backtrack tests which test concrete part
**Question**

Stryker has functionality to do test coverage analysis for tests and code.
Would it be possible to get information which line of code is tested by which tests?

My problem is, that I have more than 2000 tests and a lot of them test same parts.
I would like to reduce/delete those tests to gain speed and better maintainability of whole test suite.

nicojs:
This PR adds test reporting to both the `clear-text` and `html` reporter (it was already supported by the JSON reporter). 

<table>
<tr>
<td>

</td>
<td>

</td>
</table>

image

There are a couple of things to keep in mind:

* Stryker will only be able to produce the [`Covering` test state](https://stryker-mutator.io/docs/mutation-testing-elements/mutant-states-and-metrics/#test-states-and-metrics) with `"coverageAnalysis": "perTest"`
  image
* Currently, only the `@stryker-mutator/jest-runner` supports marking the test files for each test. This means that all other test runners will simply produce a big list of tests:
  image

Closes #2464
Closes #2636


nicojs:
@HonzaMac what do you think? 

We, unfortunately, don't have the information about the layering of the test suites, Stryker uses a flat test name structure. I personally don't think it is _that_ big of a deal though.

HonzaMac:
@nicojs Thank you for inplementing this. It really works! I was able to delete useless test, so running stryker is faster thanks removal of redundant.

Looking forward to try it on bigger suite on another project.

PR #2881, opened 2021-05-07
feat(options): make "perTest" the default for "coverageAnalysis"

### LINKED ISSUE #2872: Make coverageAnalysis default to "perTest" when command runner is not "command".
Since we now have support for `coverageAnalysis` in all supported test runners, we can finally make `perTest` the default. The `command` test runner does not (and will never) support coverage analysis, so we should make an exception for that one.

Make `perTest` the default, but downgrade to `off` when the `command` is selected (can be done in options-validator.ts). We should do this in the 5.0 release since it technically is a breaking change.

Keeping it default "off", but since a lot of users don't touch the coverage analysis setting, they would miss the awesome performance benefits it brings.

**Additional context**
Should be done in 5.0

nicojs:
Closes #2872

BREAKING CHANGE: `"perTest"` is now the default value for "coverageAnalysis" when the configured test runner is not "command". Explicitly set `"coverageAnalysis": "off"` manually to opt-out of this behavior.


kmdrGroch:
Finally made Jest working with perTest?!

PR #2886, opened 2021-05-09
feat(mutator): add the optional chaining mutator

nicojs:
Add the "OptionalChaining" mutator:

* `foo?.bar` -> `foo.bar // 👽`
* `foo?.[1]` -> `foo[1] // 👽`
* `foo?.()` -> `foo() // 👽`

If you don't want to use this mutator, you can exclude it with:

```json
{
  "mutator": {
    "excludedMutations": ["OptionalChaining"]
  }
}
```


nicojs:
We have a small challenge with placing these mutants. Take this example from `utils`:

```diff
- findUnserializables(child)?.map((description) => {
-   description.path.unshift(index.toString());
-   return description;
-   })
- )
- .filter(notEmpty);
+const things = Object.entries(thing).flatMap(stryMutAct_9fa48("128") ? () => undefined : (stryCov_9fa48("128"), ([key, val]) => stryMutAct_9fa48("129") ? findUnserializables(val).map(description => {
+  if (stryMutAct_9fa48("130")) {
+    {}
+  } else {
+    stryCov_9fa48("130");
+    description.path.unshift(key);
+    return description;
+  }
+}) : (stryCov_9fa48("129"), findUnserializables(val)?.map(description => {
+  if (stryMutAct_9fa48("130")) {
+    {}
+  } else {
+    stryCov_9fa48("130");
+    description.path.unshift(key);
+    return description;
+  }
+})))).filter(notEmpty);
```

As you can see, in this case, we'll have to duplicate the _call expression_ `?.map()`, because it is part of an optional chain. Not only that, but we're also duplicating the entire _mutated_ AST. This is a side effect that results from the somewhat naive implementation we currently have. It is not needed, since only 1 mutant can be active at any one time.

This would blow up a more serious code block. Just think about what would happen if we mutate something like this:

```js
foo?.find(bar => {
  console.log('lots');
  console.log('of');
  console.log('code');
  return bar?.find(baz => {
    console.log('lots');
    console.log('of');
    console.log('code');
    return baz?.find(qux => {
      console.log('lots');
      console.log('of');
      console.log('code');
      return qux?.find(quux => {
         console.log('lots');
         console.log('of');
         console.log('code');
         // ...
         return;
      });
    });
  });
});
```

It results in 386 lines of javascript with _only_ 20 unique mutants in it.

All this being said, I think we should move this feature to v5.1 for now.

simondel:
I think that in reality, you won't often have big blocks of code inside of an anonymous function that is called using optional chaining so I don't think the increase in code will be all that important.

nicojs:
Yeah that's a valid point. However, this is a more fundamental oversight of the current mutant placement strategy. For example, we also want to add the method expression mutator which would run into the same problem.

nicojs:
Mutation placing was fixed in #2964 , so I'm proceeding with this

PR #2902, opened 2021-05-14
Two new issues troubleshooting added

dsibinski:
Added troubleshooting info for TS compilation and our of memory issue.
Discussed on Slack.


nicojs:
Thanks a lot! 👍

PR #2908, opened 2021-05-20
fix typo in readme

rouke-broersma:


nicojs:
Thanks! You're now a StrykerJS contributor. How does that feel? 😊

rouke-broersma:
Brrrr 😣

PR #2964, opened 2021-06-22
Feat/new mutant placing

### LINKED ISSUE #2968: Don't place the same mutant twice
See https://twitter.com/_nicojs/status/1407351342428045314 👀

nicojs:
Implements a new mutant placing algorithm while also refactoring the instrumenter package.

**🔢 New mutant placing algorithm**

On the way down: 👇
* Treat the tree as immutable!
* Identify the nodes that can be used to place mutants on in the `placementMap`. I call these AST nodes "placement nodes".
   * Example: A node that is a `Statement`
* Generate the mutants on each node.
   * When a node generated mutants, do a short walk back up to the nearest node where the mutant can be placed, register them in the placement map
   * Call the new `applied` method on each `Mutant`, providing the placement node. This way the mutant will capture the AST with the single mutation from the placement node down.
  
On the way back up 👆
* If this node has mutants in the `placementMap`, place them in the AST. 

**🎨 Refactoring**
* A `Mutant` now has an `applied` method. Note: passed tense, because it treats the input as immutable.
* Split each `MutantPlacer`'s functionality in a `canPlace` method and a `place` method. The transformer uses `canPlace` to identity the placement nodes.
* Mutators no longer return an `array` of mutants. Instead, it returns an `Iterable` of replacement nodes. 
* Mutators are now implemented as objects and pure functions (instead of classes). I think this programming paradigm fits better to the use case. 
* Only the conditional expression mutator returned replacements for something other than the provided node. I've refactored that one mutator to be in line with the rest. 
* The mutate functions now return an `Iterable`, since this allows us to use [generator functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*). You can simply `yield` mutants when needed. This cleans up the code nicely
* Split up `babel-transformer` into multiple methods
* The `babel-transformer` unit tests now use test doubles instead of sinon stubs. This makes for a more robust - black box - unit test experience.

Fixes #2968


nicojs:
@Garethp feel free to review

nicojs:
@Garethp I've merged it, but feel free to still review it later and add issues/ or a new PR later.

PR #2987, opened 2021-07-02
fix/cucumber off by one error

nicojs:


nicojs:
Fixes off-by-one error in the report 
image

PR #2988, opened 2021-07-02
feat(mutator): add the optional chaining mutator 

nicojs:
Add the "OptionalChaining" mutator:

* `foo?.bar` -> `foo.bar // 👽`
* `foo?.[1]` -> `foo[1] // 👽`
* `foo?.()` -> `foo() // 👽`

If you don't want to use this mutator, you can exclude it with:

```json
{
  "mutator": {
    "excludedMutations": ["OptionalChaining"]
  }
}
```


nicojs:
This is attempt 2. Previous PR was on the wrong branch 🙈

PR #3020, opened 2021-07-23
fix(karma runner): restart a browser on disconnect error

### LINKED ISSUE #2847: karma-runner: stop using raw cli options
The current way of running `Karma` is deprecated. When running the integration tests you can see it:

```
 21 04 2021 17:32:20.977:WARN [runner]: Passing raw CLI options to `runner(config, done)` is deprecated. Use `parseConfig(configFilePath, cliOptions, {promiseConfig: true, throwErrors: true})` to prepare a processed `Config` instance and pass that as the `config` argument instead.
```

Let's do exactly what they specify in this warning to fix this.

Note: We might need to still send raw options for older versions of Karma. We'll have to see. We can start with the new implementation, since we have an e2e test for the oldest supported karma version.

### LINKED ISSUE #2989: Mutation scores differ on same file when run with other files
**Question**
We can run a mutation test on file A and get 50% score. We can then run a test on file A and file B and the mutation score for file A will be DIFFERENT than 50%. 

**Stryker environment**
    "@stryker-mutator/core": "^4.6.0",
    "@stryker-mutator/karma-runner": "^4.6.0",
    "@stryker-mutator/typescript-checker": "^4.6.0",

    "@stryker-mutator/karma-runner": "^4.6.0",

**Additional context**

nicojs:
* Restart a browser when it disconnects. This allows us to recover from an infinite loop (as long as the timeout allows for it).
* Refactor the karma runner to allow restarting of the browser
* Refactor to no longer [pass raw options to `karma.runner.run`](http://karma-runner.github.io/6.3/dev/public-api.html#karmarunner)
* Implement `dispose` in the karma runner. It now closes the browser karma server. Useful during integration testing.

Fixes #2989
Fixes #2847


packages/karma-runner/src/karma-test-runner.ts
```diff
@@ -1,77 +1,69 @@
...
-    this.listenToError();
   }
 
   public async init(): Promise<void> {
-    return new Promise((res, rej) => {
-      StrykerReporter.instance.once('browsers_ready', () => res());
-      this.starter.start().catch(rej);
+    const browsersReadyPromise = StrykerReporter.instance.whenBrowsersReady();
+    const { exitPromise } = await this.starter.start();
+    this.exitPromise = exitPromise;
```
nicojs, line 31: We should handle premature exits. It was done (in a dirty way) before, but it got lost in translation during the refactor. We should be able to add an integration test for this. For example, add a framework that crashes stuff 🤷‍♂️

PR #3031, opened 2021-07-29
feat(hit limit): infinite loop prevention in karma-runner

### LINKED ISSUE #3023: Proposal: break out of infinite loops using a hit counter

Stryker might be mutating your code to be an infinite loop. There is no way around that, because of something called the [Halting problem](https://en.wikipedia.org/wiki/Halting_problem). For example:

```ts
let goOn = true;
let i = 0;
while(goOn) {
  if(activeMutant(1)) {
    i--;   // 👈 mutant that creates an infinite loop
  }
  else { 
    i++; 
  }
  if(i >10){
    goOn = false;
  }
}
```

Currently, StrykerJS's way to handle infinite loops is to kill the worker process and start a new one. This works, but it can be very expensive, especially when webpack or other JIT build steps are involved. Take an Angular project for example. Having a new test runner worker process take a minute to start is quite 'normal' for larger projects (besides the fact that a new browser is started as well).

When StrykerJS is instrumenting the code, we could add a hit counter that counts the number of times that mutant is executed. If an absolute limit is reached, we can set a "hitCountLimitReached" flag and throw an error. Then in the test runner, we can check for the presence of the flag and report the mutant as a timeout result. 

Pseudo code:

```ts
function activeMutant(id) {
  if (id === global.__stryker__.activeMutant) {
    if (global.__stryker__.hitCount  !== undefined || ++global.__stryker__.hitCount > global.__stryker__hitCountLimit) {
      throw new Error('Mutant hit count reached');
    }
    return true;
  }
  return false;
}
```

As you can see, Stryker will throw an error when the `hitCount` of a mutant exceeds a predetermined limit. The predetermined limit can be based on the number of hits during normal (unmutated) execution, which can be determined during the dry run. 

```
hitLimit(mutant) = hits(mutant) & hitLimitFactor
```

The higher the `hitLimitFactor`, the longer it takes to break out of an infinite loop, but setting the `hitLimitFactor` too low might result in false negatives. We could start with a safe factory (i.e. 100) and see what that gives us.

We could also add a `reason` field to our `TimeouMutantRunResult`. The reason can be provided by the test runner as "Hit limit of x reached" (or something like that). That way you can identify these results in the HTML report.

There are some disadvantages to this approach:

* It is difficult to determine a universal "hard limit". For some tasks (image processing, performance tests, etc) a hard limit of 1000 would be far too low. ~~I think therefore this limit should be configurable: `--mutantHitLimit`?  Setting this value to `0` could mean turning the hit limit off.~~ @hcoles came up with an excellent idea in https://github.com/stryker-mutator/stryker-js/issues/3023#issuecomment-887301263 . The limit will be determined based on the _normal_ hit count, which can be calculated during the dry run (initial test run). This removes the need for `--mutantHitLimit` entirely. We might want to introduce a configurable `--mutantHitLimitFactory` in the future, but for now, we can use a hard factory of 100 for example (to be on the safe side)
* The error might not be handled properly by the test. It might result in an unhandled exception or unhandled promise rejection. We might want to think about adding handlers for those, or investigate if each test runner properly handles that already.
* The error might be caught by your code. This can have unexpected consequences
  * If the error is caught inside the loop, the loop won't be broken and remain infinite.
  * If the error is caught outside of the loop, where a different error was expected to be handled. This probably will result in a failing test (which should be converted to a timeout by the test runner), but still: unexpected behavior. 
* Test runners will get more complicated (although we could move some of this code to `@stryker-mutator/utils`):
    * decide when there is a timeout based on the global "hitCountLimitReached" flag.
    * Reset both the `hitCount` and the `hitCountLimit` variables. If we forget to reset, false negatives will occur.

As you can see, it can only ever be a best-effort approach and Stryker will always need its current behavior to handle infinite loops in addition. 

Note that there are still timeouts that are unhandled by this implementation, especially with regards to using callbacks in your test logic. For example: 

```ts
// foo.js
function maybeEmitFooEvent(a, b) {
  if (activeMutant(1) ? a > b : a < b) {
     emitter.emit('foo');
  }
}

// foo.spec.js
it('should raise event', (done) => {
  emitter.on('foo', done);
  maybeEmitFooEvent(1, 2);
});
```

There is no real alternative that I can think of, other than what we're doing now.

**Additional context**

Karma has difficulties with infinite loops: https://github.com/stryker-mutator/stryker-js/issues/2989#issuecomment-885461482

nicojs:
Add infinite loop prevention using a hit counter in the core and the karma runner. 

image

I want to add this functionality to the other test runners as well, but first want to see how it behaves in the wild. The karma runner is also the most expensive one to restart, so it has the most to gain with this functionality.

During mutation testing, Stryker might create an infinite loop. For example:

```ts
let i = 0;
while (i < 10){
  if(activeMutant("1")) {
    i--;   // 👈 mutant that creates an infinite loop
  } else {
    i++;
  }
}
```

In the past, Stryker would eventually report a timeout. This is expensive because Stryker has to kill the test runner process and start a new one. This can take anywhere from less than a second to multiple minutes, depending on the test runner.

With this PR, infinite-loop-prevention code is generated during code instrumentation:

```ts
function isActive(id) {
  if (ns.activeMutant === id) {
    if (ns.hitCount !== void 0 && ++ns.hitCount > ns.hitLimit) {
      throw new Error('Stryker: Hit count limit reached (' + ns.hitCount + ')');
    }
    return true;
  }
  return false;
}
```

As you can see, an error is thrown when a limit is reached. This hopefully breaks the infinite loop _but it might not_, since the error can be caught and handled by user-code inside the loop, making it infinite once again. I don't see a way around this, since [js has no `goto` statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/label) and I don't know of another way to "break out of infinity". In such cases, the timeout will still occur and the test runner process will be killed.

The job of preventing infinite loops is a shared responsibility between Stryker and the test runner plugin. 
* Stryker is responsible for adding the infinite loop prevention code and providing the hit limit to the test runner.
* The test runner plugin is responsible to set the `hitLimit` and reset the `hitCount` on the `__stryker__` variable and determining whether or not the hit limit was reached.

Important to note is that this functionality is optional. If the test runner doesn't set the `__stryker__.hitCount` and `__stryker.hitLimit` variables, the functionality will simply be ignored.

The hit limit is provided to the test runner in the `MutantRunOptions`. Stryker calculates it using the `hitCount` from the dry run:

```
hitLimit(m) = hitCount(m) * hitLimitFactor
```

Choosing a good `hitLimitFactor` here is crucial. If the value is too low, it might increase false positives. If the value is too high, it might take longer to complete. However, I've chosen to start with a factor of `100`, to lean to the save side. We might want to make this factor configurable in the future.

Note that there are theoretical cases where this functionality results in false positives, no matter how high we make the `hitLimitFactor`. For example:

```ts
let i = 0;
while (i < 10 && i > -9999){
 if(activeMutant("1")) {
    i--;   // 👈 mutant that will "falsly" be reported as a timeout
  } else {
    i++;
  }
}
```

There is one case I can think of that isn't theoretical. Namely when someone is using a property testing library, like [jsverify](https://jsverify.github.io/). During property testing, a failing scenario might be executed a lot of times to determine the [smallest possible counterexample](https://jsverify.github.io/#minimal-counterexample). I haven't tested this yet, but I am planning to do that.

List of changes:

* API
  * Add `hitLimit` to the `MutantRunOptions`
  * Add `hitCount` and `hitLimit` to `InstrumenterContext`
  * Add `hitCount` to `MutantTestCoverage` result.
  * Add optional `reason` field to `Timeout` test runner results.
  * Add `determineHitLimitReached` helper function to determine wether or not a hit limit is reached
* Instrumenter
  * Add instrumentation for hit limit and counting hits.
* Core
  * Determine the `hitCount` based on the mutant coverage.
  * Determine the `hitLimit` based on the `hitCount` and provide it in `MutantRunOptions`
* Karma runner
  * Grap `hitLimit` and configure it using the test hooks middleware.
  * Determine if a timeout occurred using the `determineHitLimitReached` helper.
  * Add integration test for this new behavior
* e2e
  * Add new `hit-limit` e2e test.

Closes #3023


nicojs:
@hcoles I've decided to go with your approach of using the dry run results to determine the 'normal' hit count as you can see in the description. Thanks again for that great insight.

hcoles:
@nicojs Glad someone is going to implement it.

packages/api/test/unit/test_runner/run-result-helpers.spec.ts
```diff
@@ -1,12 +1,30 @@
...
+  MutantRunStatus,
+  determineHitLimitReached,
+  TimeoutDryRunResult,
+} from '../../../src/test-runner';
 
 describe('runResultHelpers', () => {
+  describe(determineHitLimitReached.name, () => {
+    it('should determine a timeout result when hit count is higher than limit', () => {
+      const expected: TimeoutDryRunResult = { status: DryRunStatus.Timeout, reason: 'Hit limit reached (10/9)' };
+      expect(determineHitLimitReached(10, 9)).deep.eq(expected);
```
simondel, line 17: This is fine, but I would prefer tests to split the _act_ and _assert_ steps:
```
const result = determineHitLimitReached(10, 9);
expect(result).deep.eq(expected);
```

packages/core/test/unit/mutants/find-mutant-test-coverage.spec.ts
```diff
@@ -146,10 +148,26 @@ describe(sut.name, () => {
...
       expect(result).deep.eq(expected);
     });
 
+    it('should calculate the hitCount based on total hits (perTest and static)', () => {
+      // Arrange
+      const mutant = factory.mutant({ id: '1' });
+      const mutants = [mutant];
+      const dryRunResult = factory.completeDryRunResult({
+        tests: [factory.successTestResult({ id: 'spec1', timeSpentMs: 0 })],
+        mutantCoverage: { static: { 1: 1 }, perTest: { 1: { 1: 2, 2: 100 }, 2: { 2: 100 }, 3: { 1: 3 } } },
```
simondel, line 161: I would find this more readable:
```
mutantCoverage: { 
    static: { 1: 1 },
    perTest: { 
        1: { 1: 2, 2: 100 },
        2: { 2: 100 },
        3: { 1: 3 } 
    }
},
```

In the current form, it's a lot of similar looking numbers on one line which can be confusing

packages/core/test/unit/process/4-mutation-test-executor.spec.ts
```diff
@@ -132,6 +132,29 @@ describe(MutationTestExecutor.name, () => {
...
+
+  it('should set the hit limit to undefined when there was no hit count', async () => {
+    arrangeScenario();
+    mutants.push(factory.mutantTestCoverage({ hitCount: undefined, static: true }));
+
+    // Act
+    await sut.execute();
+
+    // Assert
+    const expected: Partial<MutantRunOptions> = { hitLimit: undefined };
```
simondel, line 155: In other tests you've placed the `expected` in the _arrange_ part of your tets

packages/karma-runner/src/karma-plugins/stryker-mutant-coverage-adapter.ts
```diff
@@ -4,13 +4,20 @@
...
+  function firstArg() {
     if (!args.length) {
       args.push({});
     }
-    // @ts-expect-error
-    args[0].mutantCoverage = window.__stryker__?.mutantCoverage;
+    return args[0];
+  }
+  // @ts-expect-error
+  const ns = window.__stryker__ || {};
```
simondel, line 14: What's `ns`?

packages/karma-runner/src/karma-plugins/stryker-mutant-coverage-adapter.ts
```diff
@@ -4,13 +4,20 @@
...
-  if (window.__strykerShouldReportCoverage__) {
+  function firstArg() {
     if (!args.length) {
       args.push({});
     }
-    // @ts-expect-error
-    args[0].mutantCoverage = window.__stryker__?.mutantCoverage;
+    return args[0];
+  }
+  // @ts-expect-error
```
simondel, line 13: What kind of error do you expect?

packages/karma-runner/src/karma-plugins/stryker-reporter.ts
```diff
@@ -56,6 +58,11 @@ export class StrykerReporter implements karma.Reporter {
     this.runTask?.resolve(this.collectRunResult());
   };
 
+  public configureHitLimit(hitLimit: number | undefined): void {
+    this.hitLimit = hitLimit;
+    this.hitCount = undefined;
```
simondel, line 63: Wouldn't it be more logical if the `hitCount` defaults to 0? Or would you prefer `undefined` because it simply may not have been set?

simondel (approved):

packages/core/test/unit/mutants/find-mutant-test-coverage.spec.ts
nicojs: Sorry, no can do 😟 Prettier doesn't allow it

packages/karma-runner/src/karma-plugins/stryker-mutant-coverage-adapter.ts
nicojs: namespace

nicojs: I'll rename to `context`

packages/karma-runner/src/karma-plugins/stryker-mutant-coverage-adapter.ts
nicojs: > `window` is not defined

I'll update it

packages/karma-runner/src/karma-plugins/stryker-reporter.ts
nicojs: Yeah, I guess `0` would work, but want to keep it `undefined`. I think that makes more sense.

simondel:
Looks good!

PR #3058, opened 2021-08-06
maintenance: clean up types

kmdrGroch:
Partially fixed `any` types, instead giving them better type or `unknown`

### DIFF
--- a/packages/api/src/logging/logger.ts
+++ b/packages/api/src/logging/logger.ts
@@ -6,10 +6,10 @@ export interface Logger {
   isErrorEnabled(): boolean;
   isFatalEnabled(): boolean;
 
-  trace(message: string, ...args: any[]): void;
-  debug(message: string, ...args: any[]): void;
-  info(message: string, ...args: any[]): void;
-  warn(message: string, ...args: any[]): void;
-  error(message: string, ...args: any[]): void;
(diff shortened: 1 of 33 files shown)


kmdrGroch:
@nicojs I have been thinking about making some utils for types such as:

```ts
// Key To Object
type KtoO<T extends number | string | symbol> = Record<T, unknown>;

// extend type
// e.x. 
// myObject.errorMessage cannot be used, but we know will exist
// instead of (myObject as any).errorMessage, you can use
// extend(myObject)<{ errorMessage: unkown }>().errorMessage;
// extend(myObject)<KtoO<'errorMessage'>>().errorMessage;
// or extend type in place: `const obj = extend(myObject)<KtoO<'errorMessage'>>()`;
// obj.errorMessage
function extend<T>(v: T) {
  return function <E>(): E & T {
    return v as unknown as E & T;
  };
}

// cast type to another type
// replaces `myObject as unknown as T`
// by `cast<T>(myObject)`
function cast<T>(v: unknown): T {
  return v as T;
}
```

nicojs:
@kmdrGroch  I'm not sure what the status is here. Could you update this branch?

kmdrGroch:
need to check how much has change since it was introduced. I shall close this PR, and open new one when I investigate it

Issue #3012, opened 2021-07-21
Killed mutants move to covered when adding unrelated tests


Edgpaez:
First of all **thank you** to all maintainers and contributors of this package. I hope this is actually a bug and I'm not wasting anyone's time.
Also, this seems to be related to #2989

**Summary**
Killed mutants move to covered when I add more tests for another class.

Initially I have a simple class that assigns some instance fields based on inputs on the constructor:
```javascript
class Frame {
  constructor(type) {
    this.state = Frame.STATES.IN_PLAY;
    this.type = type || Frame.TYPES.REGULAR;
    this.score = null;
    this.firstThrow = null;
    this.secondThrow = null;
  }
}
```
Then I have a simple test for it:
```javascript
  it("starts as a regular frame by default. could be started as a Fill Ball", () => {
    const regularFrame = new Frame();
    expect(regularFrame.type).toEqual(Frame.TYPES.REGULAR);

    const fillFrame = new Frame(Frame.TYPES.FILL_BALL_FOR_SPARE);
    expect(fillFrame.type).toEqual(Frame.TYPES.FILL_BALL_FOR_SPARE);
  });
```
Running stryker shows me I cover all mutants related to the constructor: 
```
🎯 Killed: 9 FILL_BALL_FOR_SPARE: "", (src/frame.js:13:26) (and 4 more) ☂️ Covered 4 mutants
🎯 9 FILL_BALL_FOR_SPARE: "", (src/frame.js:13:26)
🎯 12 this.type = false; (src/frame.js:18:17)
🎯 11 this.type = true; (src/frame.js:18:17)
🎯 10 constructor(type) {} (src/frame.js:16:21)
🎯 13 this.type = type && Frame.TYPES.REGULAR; (src/frame.js:18:17)
```
So far so good.

But then, as I implemented more parts of the problem, I added another class and another test for that class. 
Now, running stryker shows me this:
```
☂️ Covered 4 mutants (yet didn't kill any of them)
☂️ 61 constructor(type) {} (src/frame.js:16:21)
☂️ 62 this.type = true; (src/frame.js:18:17)
☂️ 63 this.type = false; (src/frame.js:18:17)
☂️ 64 this.type = type && Frame.TYPES.REGULAR; (src/frame.js:18:17)
```
Naturally I didn't change anything in the original code from above. Just a new `bowling.spec.js` and a `bowling.js` file

[...]

Issue #2996, opened 2021-07-08
Add configuration option to turn off bail


JSmits-utwente:
**Is your feature request related to a problem? Please describe.**
For my graduation assignment I would like to use StrykerJS's JSON reporter to export mutation information that I need. In particular I need to know for each mutant which test cases have _coverage_ over that mutant, and which test cases _kill_ that mutant.
Currently StrykerJS uses a feature called _bail_, where mutant execution stops once one test case has managed to kill that mutant. This increases performance, but also means the JSON report only ever lists a maximum of one test case as the killer of a mutant, even if there are more.

**Describe the solution you'd like**
Add an option to StrykerJS's configuration to allow toggling bail. The default for that option can then still be to turn it on. Something like:
`enableBail` [`boolean`]
Default: `true`
Command line: `-- enableBail true`
Config file: `"enableBail": true`

**Describe alternatives you've considered**
N/A

Edgpaez:
This seems to be supported already (although I haven't tested myself):
1. jest-runner seems to have an [enableBail option](https://github.com/stryker-mutator/stryker-js/blob/543fa2df8eeb88b753ac1d712fa857e0afb5f5d5/packages/jest-runner/schema/jest-runner-options.json#L30)
2. You could pass options to jest using  [`testRunnerNodeArgs`](https://stryker-mutator.io/docs/stryker-js/configuration#testrunnernodeargs-string) or using [config](https://stryker-mutator.io/docs/stryker-js/jest-runner#jestconfig-object) directly

nicojs:
Thanks for pitching in @Edgpaez ❤

You are correct, but the problem is that all test runner plugins a.t.m. only report back the first killed test. As you can see in the `MutantRunResult` interface:

https://github.com/stryker-mutator/stryker-js/blob/5466cfb4fce3de6c7b7817c2c37812215848000a/packages/api/src/test-runner/mutant-run-result.ts#L14-L28

In order to implement this feature, we should:
1. Change the `MutantRunResult` interface:
  ```diff
  export interface KilledMutantRunResult {
    status: MutantRunStatus.Killed;
  - killedBy: string;
  + killedBy: string[];
    failureMessage: string;
    nrOfTests: number;
  }
  ```
2. Update all `TestRunner` plugins to support this new interface.
3. Update the core to support this new interface
    * We should keep it backward compatible (still support plain strings), otherwise it would be a breaking change
[...]

PR #3061, opened 2021-08-08
WIP - Create the disableBail option.

Edgpaez:
As discussed in #2996 and #3012 this PR enables the user to set a `disableBail` option to collect all tests killing a mutant.

Notes on the implementation:
1. Initially only implemented for jest. The rest of the test runners should continue to work without changes
1. I deleted the `enableBail` option from `jest-runner-options.json`. I figured since this new `disableBail` option at the striker level it made no sense to keep it. It might be necessary to do the same for all test helpers
1. We only track the first error message of the first failing test. 

To be done:
- [ ] In the integration tests, jest doesn't bail and always report two tests killing a mutant, although I am sure ([and the tests confirm](https://github.com/stryker-mutator/stryker-js/pull/3061/files#diff-45feb2b206aba6659dc47f3a6f20a3449a6f5720f231929f464bf63d33cd5cefR97)) I'm telling Jest to run with `bail = true`. I am not sure what to do about this or if this is a problem with my implementation.
- [ ] enableBail should be deprecated, not removed.
- [ ] merge jasmine2-node-no-mocks into jasmine2-node
- [ ] add disableBail to the RunOptions (for both DryRunOptions and MutantRunOptions)

### DIFF
--- a/packages/api/src/test-runner/run-result-helpers.ts
+++ b/packages/api/src/test-runner/run-result-helpers.ts
@@ -11,7 +11,7 @@ export function determineHitLimitReached(hitCount: number | undefined, hitLimit:
   return;
 }
 
-export function toMutantRunResult(dryRunResult: DryRunResult): MutantRunResult {
+export function toMutantRunResult(dryRunResult: DryRunResult, reportAllKillers = false): MutantRunResult {
   switch (dryRunResult.status) {
     case DryRunStatus.Complete: {
       const failedTests = dryRunResult.tests.filter<FailedTestResult>((test): test is FailedTestResult => test.status === TestStatus.Failed);
@@ -21,7 +21,7 @@ export function toMutantRunResult(dryRunResult: DryRunResult): MutantRunResult {
         return {
           status: MutantRunStatus.Killed,
           failureMessage: failedTests[0].failureMessage,
-          killedBy: failedTests.map<string>((test) => test.id),
+          killedBy: reportAllKillers ? failedTests.map<string>((test) => test.id) : failedTests[0].id,
           nrOfTests,
         };
       } else {
--- a/packages/api/src/test-runner/run-result-helpers.ts
+++ b/packages/api/src/test-runner/run-result-helpers.ts
@@ -14,13 +14,14 @@ export function determineHitLimitReached(hitCount: number | undefined, hitLimit:
 export function toMutantRunResult(dryRunResult: DryRunResult): MutantRunResult {
   switch (dryRunResult.status) {
     case DryRunStatus.Complete: {
-      const killedBy = dryRunResult.tests.find<FailedTestResult>((test): test is FailedTestResult => test.status === TestStatus.Failed);
+      const failedTests = dryRunResult.tests.filter<FailedTestResult>((test): test is FailedTestResult => test.status === TestStatus.Failed);
       const nrOfTests = dryRunResult.tests.filter((test) => test.status !== TestStatus.Skipped).length;
-      if (killedBy) {
+
+      if (failedTests.length > 0) {
         return {
           status: MutantRunStatus.Killed,
-          failureMessage: killedBy.failureMessage,
-          killedBy: killedBy.id,
+          failureMessages: failedTests.map<string>((test) => test.failureMessage),
+          killedBy: failedTests.map<string>((test) => test.id),
           nrOfTests,
         };
       } else {
--- a/packages/api/src/test-runner/run-result-helpers.ts
+++ b/packages/api/src/test-runner/run-result-helpers.ts
@@ -20,7 +20,7 @@ export function toMutantRunResult(dryRunResult: DryRunResult): MutantRunResult {
       if (failedTests.length > 0) {
         return {
           status: MutantRunStatus.Killed,
-          failureMessages: failedTests.map<string>((test) => test.failureMessage),
+          failureMessage: failedTests[0].failureMessage,
           killedBy: failedTests.map<string>((test) => test.id),
           nrOfTests,
         };
--- a/packages/jest-runner/schema/jest-runner-options.json
+++ b/packages/jest-runner/schema/jest-runner-options.json
@@ -26,11 +26,6 @@
           "description": "Whether to run jest with the `--findRelatedTests` flag. When `true`, Jest will only run tests related to the mutated file per test. (See [_--findRelatedTests_](https://jestjs.io/docs/en/cli.html#findrelatedtests-spaceseparatedlistofsourcefiles)",
           "type": "boolean",
           "default": true
-        },
-        "enableBail": {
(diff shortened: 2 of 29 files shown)


packages/jest-runner/test/integration/jest-test-runner.it.spec.ts
```diff
@@ -44,8 +47,10 @@ describe(`${JestTestRunner.name} integration test`, () => {
       const runResult = await jestTestRunner.dryRun({ coverageAnalysis: 'off' });
 
       assertions.expectCompleted(runResult);
-      expect(runResult.tests[0].name).to.equal('Add should be able to add two numbers');
-      expect(runResult.tests[0].timeSpentMs).to.be.above(-1);
+      const result = runResult.tests.find((test) => test.id === 'Add should be able to add two numbers');
+      expect(result).to.not.be.null;
+      expect(result!.name).to.equal('Add should be able to add two numbers');
+      expect(result!.timeSpentMs).to.be.above(-1);
```
Edgpaez, line 53: I had to make this changes because I kept getting random errors when running the tests.

I figured that it was because by testing the first test in the array we were making assumptions about the order in which jest runs tests. According to [this](https://github.com/facebook/jest/issues/4032#issuecomment-315210901) jest cannot guarantee the order of the tests

packages/api/src/test-runner/run-result-helpers.ts
```diff
@@ -11,16 +11,17 @@ export function determineHitLimitReached(hitCount: number | undefined, hitLimit:
   return;
 }
 
-export function toMutantRunResult(dryRunResult: DryRunResult): MutantRunResult {
+export function toMutantRunResult(dryRunResult: DryRunResult, reportAllKillers = false): MutantRunResult {
```
Edgpaez, line 14: I'm not a huge fan of this boolean parameter (I have repeated for years that every time someone creates a boolean parameter a puppy dies 🐶 ) but I didn't think creating a type of options would be necessary. 

open for suggestions

nicojs: Haha 😀

Well, ... I don't think the boolean is needed here. We can simply always filter and fill the array that way. Performance is not an issue, because if bail was active, the final test would always be the failing one 🤷‍♀️

🐶

packages/jest-runner/schema/jest-runner-options.json
```diff
@@ -26,11 +26,6 @@
           "description": "Whether to run jest with the `--findRelatedTests` flag. When `true`, Jest will only run tests related to the mutated file per test. (See [_--findRelatedTests_](https://jestjs.io/docs/en/cli.html#findrelatedtests-spaceseparatedlistofsourcefiles)",
           "type": "boolean",
           "default": true
-        },
-        "enableBail": {
```
nicojs, line 30: This is technically a breaking change. I think we should migrate the option and log a warning here:

https://github.com/stryker-mutator/stryker-js/blob/518ebe6b946fc35138b636a015b569fe9a272ed0/packages/core/src/config/options-validator.ts#L35

packages/jest-runner/test/integration/jest-test-runner.it.spec.ts
nicojs: That makes sense. Good find!

packages/jest-runner/test/integration/jest-test-runner.it.spec.ts
```diff
@@ -157,5 +162,26 @@ describe(`${JestTestRunner.name} integration test`, () => {
       assertions.expectKilled(firstResult);
       assertions.expectSurvived(secondResult);
     });
+
+    it('should be able to collect all tests that kill a mutant when disableBail = true', async () => {
```
nicojs, line 166: Great to see you've added this test 🤩

tasks/instrument-test-resources.js
```diff
@@ -27,24 +27,28 @@ async function main() {
     './packages/jest-runner/testResources/jasmine2-node/src/Add.js': './packages/jest-runner/testResources/jasmine2-node-instrumented/src/Add.js',
     './packages/jest-runner/testResources/jasmine2-node/src/Circle.js': './packages/jest-runner/testResources/jasmine2-node-instrumented/src/Circle.js',
   },  '__stryker2__')
+  await instrument({
```
nicojs, line 30: We might be able to change an existing scenario, right? Instead of adding one 🤷‍♂️ Or is there a specific reason for another one?

nicojs (changes requested):
Great PR so far! I'm impressed with how much you've figured out! Always good to see integration tests being added. I see you've even found my little "tasks/instrument-test-resources.js" helper file 😁. Please take a look at my remarks.

Some additional remarks:

* Probably it makes sense to also add `disableBail` to the `RunOptions` (for both `DryRunOptions` and `MutantRunOptions`), that way a TestRunner doesn't have to use the `StrykerOptions` to determine it. It's more explicit that way.
* We probably want to add an e2e test. I would like to see it done in the same vein as the "coverage-analysis" e2e test. So 1 e2e test called "disable-bail", where we add a test per test runner implementation Stryker officially supports. The test would entail running Stryker with `disableBail` and see if the `mutation-report.json` has the correct killed-by arrays.

Would it be OK to work together on this? I could add the e2e tests for example and start implementing this in the other test runners.

packages/api/src/test-runner/run-result-helpers.ts
Edgpaez: I added the boolean because if `toMutantRunResult ` starts returning an array it'd break some of the test from other test runners. For example [this test](https://github.com/stryker-mutator/stryker-js/blob/master/packages/mocha-runner/test/integration/sample-project-instrumented.it.spec.ts#L114)

So I think it is necessary while we implement this for the other tests runners.

What do you think?

packages/jest-runner/schema/jest-runner-options.json
Edgpaez: You're right. I'll do that

tasks/instrument-test-resources.js
Edgpaez: I initially added it to `jasmine2-node` but then started to break other tests and then I found the bailing problem and wanted to isolate the test.

I'll give it a go. Should be simple enough

Edgpaez:
Thank you for the review @nicojs!
1. I'll do the `RunOptions` you mention 👍  It does make sense.
2. I would be happy to work on this together. I'll try to implement your suggestions in the following days. You should be able to implement the e2e and other test runners independently without much problem, right?

Anyway, ping me if you want to coordinate on something.

tasks/instrument-test-resources.js
nicojs: It might be better to add it to the jest-circus integration tests since jest-circus is now the default for jest since v27

nicojs:
This got merged by accident 🤦‍♀️. I've reverted that on master, but this PR is now in `merged` state. 

@Edgpaez can you create a new PR. When you do so, please don't call your branch "master". This was the reason I ended up pushing it to `master`

PR #3065, opened 2021-08-09
docs(troubleshooting): add decorator troubleshooting guide

joielee09:
Add the example and solution for the decorator-legacy issue.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -172,33 +172,30 @@ FATAL ERROR: Committing semi space failed. Allocation failed - JavaScript heap o
 
 To solve that, you need to limit the number of workers. In the example about, 11 workers were created in total (6 checker processes and 5 test runner processes). In order to solve this issue, you may try to set `--concurrency` setting in your stryker config. In case of the sample issue, adding `"concurrency": 4` to `stryker.conf.json` solved the problem.
 
+### Mutating fails: Cannot use the decorators and decorators-legacy plugin together
 
-### Babel decorator Error: `Cannot use the decorators and decorators-legacy plugin together`
-**Example**:
-```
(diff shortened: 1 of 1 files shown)


nicojs:
Thanks! 🎉 

I've changed the wording a bit and added an explanation of what is wrong.

joielee09:
@nicojs Thank you for your feedback. I definitely see the difference. I only describe 'How'. On the other hand, you explained 'Why'. Why it happens and why it is fixed with an alternative workaround. I feel a little embarrassing for how little did I know. When I meet issues I will find the answer in the way you do it. : )

docs/troubleshooting.md
```diff
@@ -172,37 +172,38 @@ FATAL ERROR: Committing semi space failed. Allocation failed - JavaScript heap o
...
-In order to solve this, you need to make sure your project have babel and the decorator-legacy plugin setting. For example,
-```
-    plugins: [
-      ["@babel/plugin-proposal-decorators", { legacy: true }],
-      ["@babel/plugin-proposal-class-properties", { loose: true }],
-    ],
-```
-See [here](https://babeljs.io/docs/en/babel-plugin-proposal-decorators) for babel decorators.
-You also have to let Stryker notice the plugins. Put this code in your stryker.conf.json.
+Stryker JS uses [babel](https://babeljs.io/docs/en/) to parse and mutate your files. When parsing your your `.js` files, it reads in the `.babelrc` and `babel.config.js` files you have, but applies its own default plugins first. In rare circumstances, those default plugins are incompatible with your babel configuration.
```
joielee09, line 195: There's 'your' written twice in ' When parsing your your `.js` files,' : )

nicojs:
Haha no problem. I'm one of the maintainers, so it makes sense that I can explain why. I'm very grateful for the work you did 😊

PR #3072, opened 2021-08-12
feat(ignore): support disable directives in source code

### LINKED ISSUE #1472: Ignore specific mutations

Garethp:
Add support `// Stryker disable` and `// Stryker restore` directives in source code.

Some examples:

```js
// Stryker disable all
const foo = 2 + 40; // mutants here are ignored
// Stryker restore all

// Stryker disable BooleanLiteral
const foo = true || false; // => BooleanLiteral mutant is ignored, but other mutants are tested.
// Stryker restore all

// Stryker disable next-line all
const foo = true || false; // => ignored mutants
const bar = true || false; // => mutants 

// Stryker disable next-line all: I don't care about this piece of code
const foo = bar || baz; // ignored mutants with reason "I don't care about this piece of code"
```

Ignored mutants _do end up in your mutation report_, but are not placed or tested and they don't count towards your mutation score.

image

Closes #1472


packages/instrumenter/src/transformers/babel-transformer.ts
```diff
@@ -139,6 +197,7 @@ export const transformBabel: AstTransformer<ScriptFormat> = (
    */
   function collectMutants(path: NodePath) {
     return [...mutate(path)]
+      .filter((mutant) => !isMutantInDisabledMap(path.node.loc, mutant))
```
nicojs, line 200: I don't want the mutants to be _filtered out_, instead I want them to be collected, but with an `ignoreReason` set. That way, they will still end up in the report, but won't be placed, tested or influence the mutation score.

packages/instrumenter/src/transformers/babel-transformer.ts
```diff
@@ -84,6 +92,40 @@ export const transformBabel: AstTransformer<ScriptFormat> = (
...
+   */
+  function addCommentsToMutantDisableMap(location: types.SourceLocation | null, comments: readonly types.Comment[] | null): void {
+    if (!comments || !location) {
+      return;
+    }
+
+    comments
+      .filter((comment) => comment.value.trim().startsWith('Stryker disable-next-line'))
+      .forEach((comment) => {
+        const [_, _disableType, mutations] = comment.value.trim().split(' ');
```
nicojs, line 107: We should also keep track of the `ignoreReason`. For example:

```js
// Stryker disable-next-line update The update mutator creates an infinite loop here.
```

The ignore reason we can then report is: `"Disabled by user comment. The update mutator creates an infinite loop here."`

For this to work, and keep parsing of the comment positional, we would need to force users to use `// Stryker disable-next-line all` if they want to ignore all, otherwise we cannot tell where the ignore reason starts.

What do you think about this?

nicojs (changes requested):
Thanks @Garethp . I have a few remarks, mainly that I want ignore mutants to end up the HTML report with status "ignored", for details see remarks.

Note: we shouldn't forget to add an e2e test for this. Probably it would make sense to add tests to the `e2e/test/ignore-project`

packages/instrumenter/src/transformers/babel-transformer.ts
Garethp: It's not my preference, having to type `all` seems coutner to how I'd expect it to work (Personally I'd expect it to work the way eslint does since we're in a JS context and that's how I kind of expect these things to work in JS). I'd **prefer** to have have us take the positional argument there and check it against the list of enabled mutations to determine if it's a specific mutant or the start of a comment. I think we'd have to do this anwyay if we want to force the positional argument in order to check if the user accidently left it off.

nicojs: Hmm good point. But still don't want to leave it up to "chance". 

Maybe we can force the mutator names to be in square brackets? Some examples:

```js
// Stryker disable-next-line [update] The update mutator creates an infinite loop here.
// Stryker disable-next-line We don't care about log statements
// Stryker disable-next-line [update] 
// Stryker disable-next-line
```

(the last 2 are without ignoreReasons)

If there are no ignore reasons given, we should still ignore them with "Disabled by user comment", or something like that.

Garethp:
@nicojs I've made it so that those mutants are marked with an ignore reason when disabled and added some tests to the `ignore-project`. I've also added the `// Stryker disable [mutators] [reason]` and `// Stryker restore [mutators] [reason]` commands. The HTML reporter doesn't show the user-defined reasons as far as I can see, but they do make it into the report and in the HTML viewer you see the comment right above the ignored mutation anyway

nicojs (approved):

nicojs:
@Garethp I actually found some more bugs and decided to change the implementation because of it. Much better now! See https://github.com/stryker-mutator/stryker-js/pull/3072/commits/fd5e978b7c0c1ce45c7e89fedb3283ab9d4aee51

PR #3074, opened 2021-08-14
feat(test runner): Support for disable bail

Edgpaez:
As discussed in #2996 and #3012 this PR enables the user to set a `disableBail` option to collect all tests killing a mutant. Continuation of #3061

Notes on the implementation:
1. Initially only implemented for jest. The rest of the test runners should continue to work without changes
1. We only track the first error message of the first failing test. 

To be done:
- [X] enableBail should be deprecated.
- [x] In the integration tests, jest doesn't bail and always report two tests killing a mutant, although I am sure ([and the tests confirm](https://github.com/stryker-mutator/stryker-js/pull/3061/files#diff-45feb2b206aba6659dc47f3a6f20a3449a6f5720f231929f464bf63d33cd5cefR97)) I'm telling Jest to run with `bail = true`. I am not sure what to do about this or if this is a problem with my implementation.  
    [@nicojs]: This is a problem of our jest-runner implementation. Running with `bail` apparently never worked 🤷‍♀️. I've removed the implementation and mimicked the behavior of reporting the first failing test when bail is enabled.
- [x] merge jasmine2-node-no-mocks into jest-circus
- [x] add disableBail to the RunOptions (for both DryRunOptions and MutantRunOptions)
- [x] Implement `disableBail` in the cucumber runner
- [x] Implement `disableBail` in the jasmine runner
- [x] Implement `disableBail` in the karma runner
- [x] Implement `disableBail` in the mocha runner


Edgpaez:
Ok. So I progressed a little:

1. I added the deprecation warning for `enableBail`  option in the jest config 
2. I created a new `MutantRunOptions` with the `disableBail` field so this is not a breaking change. Once everything is implemented we can move back to a single `MutantRunOptions`. Now it'd be easier for other test runners to add support to `disableBail`. I did not added it to `RunOptions` because we were bailing on the DryRun anyway. I didn't see much value in configuring the dryRun not to bail. You think that makes sense? or should I move it to `RunOptions`?

Edgpaez:
I've been trying to make jest bail in the integration tests based on config but I haven't succeeded.
 
Calling jest's `runCLI` with `runInBand: true` makes no difference. Neither does passing it in the `config` option. `maxConcurrency=1 ` and `maxWorkers=1` don't seem to make a difference either.

[I added a test ](https://github.com/stryker-mutator/stryker-js/pull/3074/files#diff-57914dd5cc748ee2fe21c87bb13d6f8852ea394d76a2ef27f3ee680631e7cf37R164-R180) that should pass if jest bailed on the first fail. It's skipped for now because jest does not bail

packages/api/src/test-runner/run-result-helpers.ts
```diff
@@ -11,16 +11,17 @@ export function determineHitLimitReached(hitCount: number | undefined, hitLimit:
   return;
 }
 
-export function toMutantRunResult(dryRunResult: DryRunResult): MutantRunResult {
+export function toMutantRunResult(dryRunResult: DryRunResult, reportAllKillers = false): MutantRunResult {
```
nicojs, line 14: >I added the boolean because if `toMutantRunResult` starts returning an array it'd break some of the test from other test runners. For example this test
> 
> So I think it is necessary while we implement this for the other tests runners.
> 
> What do you think?

Ah I understand. Great thinking. I think it's OK for `@stryker-mutator/api@5.4` to not be compatible with `@stryker-mutator/core@5.3`. This won't be a problem at all for Stryker, because we're pinning the dependency on `@stryker-mutator/api` with each release. If a test runner in the wild is using `@stryker-mutator/api`, than it should also pin the version on minor version. 

However, you are correct that it is strictly a breaking change. Feel free to keep it like this.

packages/core/src/config/options-validator.ts
```diff
@@ -87,6 +87,14 @@ export class OptionsValidator {
       const existingIgnorePatterns = Array.isArray(rawOptions[ignorePatternsName]) ? (rawOptions[ignorePatternsName] as unknown[]) : [];
       rawOptions[ignorePatternsName] = [...newIgnorePatterns, ...existingIgnorePatterns];
     }
+    // @ts-expect-error jest.enableBail
+    if (rawOptions.jest?.enableBail !== undefined) {
+      this.log.warn(
+        'DEPRECATED. Use of "enableBail" inside jest options is deprecated, please use "disableBail" in the stryker configuration. See https://stryker-mutator.io/docs/stryker-js/configuration#disablebail-boolean'
```
nicojs, line 93:
```suggestion
        'DEPRECATED. Use of "jest.enableBail" inside deprecated, please use "disableBail" instead. See https://stryker-mutator.io/docs/stryker-js/configuration#disablebail-boolean'
```

packages/jest-runner/src/jest-test-runner.ts
```diff
@@ -106,7 +106,7 @@ export class JestTestRunner implements TestRunner {
     }
     try {
       const { dryRunResult, jestResult } = await this.run({
-        jestConfig: withCoverageAnalysis(this.jestConfig, coverageAnalysis),
+        jestConfig: withCoverageAnalysis({ ...this.jestConfig, bail: true }, coverageAnalysis),
```
nicojs, line 109: It might make sense for `disableBail` to also be added to dry run options and used here. That is what I would expect if I were using `disableBail`. That way you can see all the error messages when the dry run fails.

nicojs (changes requested):
Thanks for your continued work on this. I've started implementing the e2e tests for jest (and prepared it for adding mocha, jasmine, karma and cucumber as well). 

Please take a look at my remarks with regards to jest.

packages/jest-runner/src/jest-test-runner.ts
Edgpaez: Makes sense. It's Done

packages/api/src/test-runner/run-result-helpers.ts
Edgpaez: I'll keep it like this if you don't mind. When all is done in the rest of the test runners it'll be easy to remove.

packages/api/src/test-runner/run-options.ts
```diff
@@ -19,4 +19,5 @@ export interface MutantRunOptions extends RunOptions {
   hitLimit?: number;
   activeMutant: Mutant;
   sandboxFileName: string;
+  disableBail?: boolean;
```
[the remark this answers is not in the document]
Edgpaez: Good. It's done now

nicojs (approved):

PR #3157, opened 2021-09-24
Fix typo in disable-mutants docs

edi9999:
The line above says "restore the EqualityOperator for 1 line:", so I guess the "next-line" keyword should be used.


nicojs:
Thx!

PR #3179, opened 2021-09-30
feat(checkers): allow custom checker node args

nicojs:
Allow specifing of custom checker node arguments.

For example: `--checkerNodeArgs "--inspect-brk" --concurrency 1` will allow a developer to debug their checker plugin.


nicojs:
@ThierryRietveld would this feature be useful for Danny and you?

PR #3183, opened 2021-10-04
killedBy is now always an array

Edgpaez:
This PRs remove some code needed to support plugins with and without support for disableBail.
Since most of the plugins support it now it is no longer necessary.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2667,41 +2667,41 @@
       }
     },
     "@typescript-eslint/parser": {
-      "version": "4.32.0",
-      "resolved": "https://registry.npmjs.org/@typescript-eslint/parser/-/parser-4.32.0.tgz",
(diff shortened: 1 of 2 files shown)


packages/jest-runner/test/integration/jest-test-runner.it.spec.ts
```diff
@@ -161,24 +161,6 @@ describe(`${JestTestRunner.name} integration test`, () => {
       assertions.expectSurvived(secondResult);
     });
 
-    it('should only report the first failing test in `killedBy` when disableBail = false', async () => {
```
nicojs, line 164: Does this mean that mimicking the bail behavior (`disableBail = false`) for jest no longer works?

Edgpaez: You're right. I kept the behaviour for this to keep working

PR #3187, opened 2021-10-06
killedby is now always an array

Edgpaez:
This PRs remove some code needed to support plugins with and without support for disableBail.
Since most of the plugins support it now it is no longer necessary.

We keep behavior in jest where we only report the first failing test

```diff
const runResult: KilledMutantRunResult = {
  status: MutantRunStatus.Killed,
-  killedBy: 'spec-1',
+  killedBy: ['spec-1'],
  failureMessage: 'foo should be bar,
  nrOfTests: 3
}
```


nicojs:
@Edgpaez thanks for the work! We're going to merge this in the `epic/esm` branch (which is the Stryker v6 branch)

PR #3188, opened 2021-10-06
specify behavior of jest when disableBail

Edgpaez:
Just a minor update to be extra clear to users:

We never bail in Jest but we mimic based on disableBail


nicojs (approved):

nicojs:
Thanks! Great addition 👍

PR #3194, opened 2021-10-11
Fix cache behavior on case-insensitive file systems

fredericbonnet:
This PR fixes an edge case on Windows where some cache entries are not recorded properly due to the case insensitivity of the filesystem. The problem was detected during a debugging session where some module filenames had a drive letter in a different casing than the one returned by `process.cwd()` (e.g. `c:\path\to\project` vs. `C:\path\to\project\file.js`).

The fix introduces a new module `platform` in the `util` package that exports the function `caseSensitiveFs()`. This option is more portable and easier to test than explicitly relying on `process.platform` from within `DirectoryRequireCache`.

### DIFF
--- a/packages/util/src/directory-require-cache.ts
+++ b/packages/util/src/directory-require-cache.ts
@@ -23,9 +23,9 @@ export class DirectoryRequireCache {
     if (!this.cache) {
       const cache = (this.cache = new Set());
       const cwd = process.cwd();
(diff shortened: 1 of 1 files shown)


nicojs:
Great find! I understand what you did here, just a very small remark. If you don't have the time for it, I can do it as well.

Issue #3174, opened 2021-09-30
 Implement hit counter in the jest-runner


nicojs:
Implement the hit counter for Jest 🃏. This is to prevent infinite loops from resulting in a timeout, see #3023.

nicojs:
Closed with #3439

PR #3236, opened 2021-10-30
Implement hit counter in the jest-runner

erie0210:
This PR fixes issue [#3174](https://github.com/stryker-mutator/stryker-js/issues/3174)

✔ Add jest instrumentation, testResources
✔ Configure hitLimit, hitCout on messaging and StrykerJestEnvironment btw nodeJS and VM
✔ Unit tests pass on Jest 

✔ Integration tests pass on Jest
✔ Test for setup in jest
✔ Test and E2E suites pass on parent project 

Now I am working on Integration tests. It seems I have to make some changes to make state reset btw runs.
I need more time for this issue to be done, and would want to work more if it is okay with stryker :) Thanks!

-----

updated


packages/jest-runner/src/jest-plugins/mixin-jest-environment.ts
```diff
@@ -39,14 +39,25 @@ export function mixinJestEnvironment<T extends typeof JestEnvironment>(JestEnvir
         if (state.coverageAnalysis === 'perTest' && event.name === 'test_start') {
           const ns = (this.global[this.global.__strykerGlobalNamespace__] = this.global[this.global.__strykerGlobalNamespace__] ?? {});
           ns.currentTestId = fullName(event.test);
+          ns.hitCount = state.hitCount;
```
nicojs, line 42: I don't think we should reset the `hitCount` between tests. When Stryker's core calculates the `hitLimit`, it takes all tests into account.

packages/jest-runner/src/jest-plugins/mixin-jest-environment.ts
```diff
@@ -39,14 +39,25 @@ export function mixinJestEnvironment<T extends typeof JestEnvironment>(JestEnvir
...
+          ns.hitLimit = state.hitLimit;
         }
       };
 
       public async teardown() {
         const mutantCoverage = this.global[this.global.__strykerGlobalNamespace__]?.mutantCoverage;
+        state.hitCount = this.global[this.global.__strykerGlobalNamespace__]?.hitCount;
```
nicojs, line 49: It's probably more clear here if we initialize the `state.hitCount` with `0`

packages/jest-runner/test/integration/jest-test-runner.it.spec.ts
```diff
@@ -124,7 +124,7 @@ describe(`${JestTestRunner.name} integration test`, () => {
 
       assertions.expectKilled(runResult);
       expect(runResult.killedBy).eq('Add should be able to add two numbers');
-      expect(runResult.failureMessage).contains('Expected: 7').contains('Received: -3');
+      // expect(runResult.failureMessage).contains('Expected: 7').contains('Received: -3');
```
nicojs, line 127: Why is this comment here?

nicojs (changes requested):
Sorry it took me a while to take a look 👀

Looks great so far 🤩! Just a couple of small remarks.

packages/jest-runner/test/integration/jest-test-runner.it.spec.ts
erie0210: This is because this test throws error:
It expects 'Expected: 7' and 'Received: -3' but runResult contains 'Expected: \x1B[32m7\x1B[39m\n' + 'Received: \x1B[31m-3\x1B[39m\n'. I was trying to figure it out.

packages/jest-runner/src/jest-plugins/mixin-jest-environment.ts
nicojs: Ow sorry, my mistake. THis is the `teardown` 😅

packages/jest-runner/src/jest-plugins/mixin-jest-environment.ts
```diff
@@ -39,14 +39,25 @@ export function mixinJestEnvironment<T extends typeof JestEnvironment>(JestEnvir
...
+        state.hitLimit = this.global[this.global.__strykerGlobalNamespace__]?.hitLimit;
         state.handleMutantCoverage(this.fileName, mutantCoverage);
         await super.teardown();
       }
+
+      public async setup() {
+        await super.setup();
+        const ns = (this.global[this.global.__strykerGlobalNamespace__] = this.global[this.global.__strykerGlobalNamespace__] ?? {});
+        ns.hitCount = state.hitCount;
```
nicojs, line 58: It's probably more clear here if we initialize the `state.hitCount` with `0`

nicojs, line 58: It's probably more clear here if we initialize the `state.hitCount` with `0`

erie0210:
Hi @nicojs, I made some changes,
I agreed with you on most of your comments and I fixed it.
I also changed `overrideEnvironment` method to be used only when there is `hitLimit`. If not so, it changes every config file.

For the `testcase where I put "//"` (to make it as a comment), it turns out to be the problem of ANSI code ( to make it printed with color) I found it works fine when I run it in debug mode. So I uncommented it but I think I have to examine it a bit more.

Also, I found my commit message was not how it should be( e.g. `Implement Jest hit limit` -> without `feat(hit limit)` in the front), I tried to rebase it with reword.. which made me add 14 commits more which is not what I intended to do... So I want to know how you would do in my situation.

Thanks for your advice and patience... 😥

nicojs:
Don't worry about the commit message, we will use squash merge and when doing that think of the correct commit message (you will still be marked as the author in git and github).

nicojs (approved):

erie0210:
@nicojs  Hi nico, I was taken away for a while. I am still working on it and it seems e2e is not working. I'll try to figure it out within a week. Thanks for you patience and have a good day 🌱

erie0210:
Yes. I merged your PR ( I'm really grateful for your PR for taking care of me.) but couldn't see any commit. Maybe, as you said, I recreated it tried to solve the error(the one you solved) and got stucked again.. I'm really sad to say this but I think you should take this issue.. @nicojs

nicojs:
It's merged. Let me know if you need more help. No problem at all

nicojs:
@erie0210 how do you want to proceed? Are you going to update this PR?

nicojs:
@erie0210 thanks for all the help! I'm finishing it up today in #3439 ❤

erie0210:
Thank you. @nicojs

PR #3256, opened 2021-11-15
docs: --mutate as an alternative to --files

pmcelhaney:
If you want to target a specific file (or files) for mutation, such as those that are staged in Git, the `--mutate` option works better than `--ignorePatterns`.


nicojs:
Ho @pmcelhaney , thanks for opening this issue. `--files` really is simply replaced by `--ignorePatterns`, so the way you've rephrased, might also cause some confusion. Could I suggest this?

```diff
- **DEPRECATED**. Please use [`ignorePatterns`](#ignorepatterns-string) instead.
+ **DEPRECATED**. Please use [`ignorePatterns`](#ignorepatterns-string) instead, or use [mutate](#mutate-string) to select which files to mutate. 
```

pmcelhaney:
Yep, that makes sense!

PR #3294, opened 2021-12-02
fix(config): `declarationMap` should disabled as well

kamontat:
Even user enabled is config in their `tsconfig.json` file, typescript compiler will throw exception below anyway. So it better to just disabled in adjustment config.

```
TS5069: Option 'declarationMap' cannot be specified without specifying option 'declaration' or option 'composite'.
```


nicojs:
Thx! Released in v5.5.1

PR #3312, opened 2021-12-10
Update guide

suzanicris:
The @stryker-mutator/html-reporter package is a stub package. The html reporter is now included with @stryker-mutator/core. There is no need to install the @stryker-mutator/html-reporter package. Mutator plugins are no longer needed in Stryker@^4.0, so remove "@stryker-mutator/javascript-mutator" from devDependencies.


nicojs:
Thanks @suzanicris !!

Wow, that guide was old! I went ahead and did some more changes.

PR #3316, opened 2021-12-14
feat(instrumenter/mutators): Implement missing MethodExpression

kaesardev:
This pull request contains the missing MethodExpressions implemented for:

enum MethodExpressions {
  'forEach' = 'map',
  'map' = 'forEach',
  'filter' = 'reduce',
  'reduce' = 'filter',
  'some' = 'every',
  'every' = 'some',
  'from' = 'assign',
  'assign' = 'from',
  'slice' = 'split',
  'split' = 'slice',
}

### DIFF
--- a/docs/guides/react.md
+++ b/docs/guides/react.md
@@ -7,7 +7,7 @@ Stryker supports React projects using Jest with both JSX and TSX code.
 
 ## JSX project
 
-Recommended stryker packages: `npm i -D @stryker-mutator/core @stryker-mutator/jest-runner @stryker-mutator/javascript-mutator @stryker-mutator/html-reporter`
+Recommended stryker packages: `npm i -D @stryker-mutator/core @stryker-mutator/jest-runner`
 
 Recommended other packages:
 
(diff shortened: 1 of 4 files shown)


kmdrGroch:
Hey first of all, thanks for contributing!

Unfortunately I don't see the cases for this type of mutant. I feel like most of these functions you provided are not interchangeable. What I mean is that doing for example split() and slice() will provide completely different types, which, yes, might detect missing tests, but most likely will fail by itself because of types regulations (in case of typescript). As for map/foreach and some/every - they make more sense, although I would love to see some test cases for them, to prove their usefulness and correctness. Also consider if there is chance and how high of a chance for equivalent mutants being generated :)

(Also for some reason I can't see why the build is failing, but you should check it ;) )

nicojs:
Thanks, @kaesardev for another PR ❤

I agree somewhat with @kmdrGroch . We did some thinking work on this mutator a while back. You can read the design here:

https://github.com/stryker-mutator/stryker-js/issues/1907#issuecomment-564466199

What do you think of that list there?

packages/instrumenter/test/unit/mutators/method-expression-mutator.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+import { methodExpressionsMutator as sut } from '../../../src/mutators/method-expression-mutator';
+import { expectJSMutation } from '../../helpers/expect-mutation';
+
+describe(sut.name, () => {
+  it('should have name "MethodExpression"', () => {
+    expect(sut.name).eq('MethodExpression');
+  });
+
+  it('should mutate forEach and map', () => {
+    expectJSMutation(sut, 'a.forEach()', 'a.map()');
```
edi9999, line 12: Mutating a "forEach" into a "map" would in my opinion almost always survive the test, or do you see a case where it would not ?

nicojs:
We've since implemented #3508, which implements these replacements: https://github.com/stryker-mutator/stryker-js/issues/1907#issuecomment-564466199

If you feel like method substitutions are missing, feel free to open an issue.

Issue #3340, opened 2022-01-08
Support Jasmine 4


nicojs:
**Is your feature request related to a problem? Please describe.**
Jasmine 4 released about a week ago and **is not supported**. 

The biggest issue is that jasmine 4 will use `import` by default. Previously I didn't think this would be an issue, because using `import` on a cjs module is supported in nodejs. However, apparently, this _breaks our ability to bust the file from require cache_. It _is_ in the cache and can be removed, but importing it again will result in the same file in either case. I will investigate this further when I have some time.

This makes #2922 even more important, we should give it the highest priority.

Another smaller change is `failFast` got renamed to `stopOnSpecFailure`.

See [release notes](https://github.com/jasmine/jasmine-npm/blob/main/release_notes/4.0.0.md) for a full list of changes. 

This was also apparent in our CI pipeline. The CI for node 16 was broken. Node 16 comes with npm 7 which _installs peerDependencies by default_. This made it so we got jasmine 4 as a dependency of jasmine-runner, which broke the ci. See https://github.com/stryker-mutator/stryker-js/runs/4738780880?check_suite_focus=true

```
@stryker-mutator/jasmine-runner:   JasmineRunner integration with code instrumentation
@stryker-mutator/jasmine-runner:     dryRun
@stryker-mutator/jasmine-runner:       ✔ should not report coverage when coverageAnalysis is "off" (42ms)
@stryker-mutator/jasmine-runner:       1) should report static coverage when coverageAnalysis is "all"
@stryker-mutator/jasmine-runner:       2) should report static and perTest coverage when coverageAnalysis is "perTest"
@stryker-mutator/jasmine-runner:     mutantRun
@stryker-mutator/jasmine-runner:       3) should be able to kill a mutant
@stryker-mutator/jasmine-runner:       ✔ should be able report "survive" when a mutant is invincible
@stryker-mutator/jasmine-runner:       4) should be able to kill again after a mutant survived
@stryker-mutator/jasmine-runner:       5) should report all killed mutants when disableBail is true
@stryker-mutator/jasmine-runner:   JasmineRunner integration
@stryker-mutator/jasmine-runner:     using the jasmine-init project
@stryker-mutator/jasmine-runner:       ✔ should run the specs
@stryker-mutator/jasmine-runner:       6) should be able to run twice in short succession
@stryker-mutator/jasmine-runner:       7) should be able to filter tests
[...]

PR #3341, opened 2022-01-08
fix(jasmine): correct peer dependency for jasmine

nicojs:
Clearly document that jasmine 4 is not yet supported in the `peerDependencies`


nicojs:
Related to #3340

PR #3346, opened 2022-01-08
feat(mutators) Avoid creating some unnecessary mutators for (a && b) …

### LINKED ISSUE #3339: Avoid using "true" condition replacement in "||" statements, and avoid using "false" replacement in "&&" statements

I have some code like this : 

```
if (part.type === "content" || part.type === "tag") {
    return part;
}
```

And there are 3 applied mutants : 

```
if (true) {
    return part;
}

if (true || part.type === "tag") {
    return part;
}
if (part.type === "content" || true) {
    return part;
}
```

I think it would make sense to just run one of the mutants in this case.

edi9999:
Fix #3339

Will be adding a few comments inline

Before:

```diff
- if (a && b) {}
+ if (false) {}
+ if (false && b) {}
// And others

- if (a || b) {}
+ if (true) {}
+ if (true && b) {}
// And others
```

After:

```diff
- if (a && b) {}
+ if (false) {}
// And others

- if (a || b) {}
+ if (true) {}
// And others
```

### DIFF
--- a/packages/instrumenter/src/mutators/conditional-expression-mutator.ts
+++ b/packages/instrumenter/src/mutators/conditional-expression-mutator.ts
@@ -14,6 +14,22 @@ export const conditionalExpressionMutator: NodeMutator = {
       yield types.booleanLiteral(true);
       yield types.booleanLiteral(false);
     } else if (isBooleanExpression(path)) {
+      if (path.parent?.type === 'LogicalExpression') {
+        // For (x || y), do not generate the (true || y) mutation as it
+        // has the same behavior as the (true) mutator, handled in the
+        // isTestOfCondition branch above
+        if (path.parent.operator === '||') {
+          yield types.booleanLiteral(false);
+          return;
+        }
+        // For (x && y), do not generate the (false && y) mutation as it
+        // has the same behavior as the (false) mutator, handled in the
+        // isTestOfCondition branch above
+        if (path.parent.operator === '&&') {
+          yield types.booleanLiteral(true);
+          return;
+        }
+      }
       yield types.booleanLiteral(true);
       yield types.booleanLiteral(false);
     } else if (path.isForStatement() && !path.node.test) {
--- a/packages/instrumenter/test/helpers/expect-mutation.ts
+++ b/packages/instrumenter/test/helpers/expect-mutation.ts
@@ -52,6 +52,10 @@ export function expectJSMutation(sut: NodeMutator, originalCode: string, ...expe
       }
     },
   });
-  expect(mutants).lengthOf(expectedReplacements.length);
-  expectedReplacements.forEach((expected) => expect(mutants, `was: ${mutants.join(',')}`).to.include(expected));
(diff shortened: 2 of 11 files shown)


packages/instrumenter/test/unit/mutators/conditional-expression-mutator.spec.ts
```diff
@@ -70,6 +70,61 @@ describe(sut.name, () => {
...
+      sut,
+      'if ((c1 || c2) && (c3 || c4)) { a++ }',
+      'if (true) { a++ }',
+      'if (false) { a++ }',
+      'if ((true) && (c3 || c4)) { a++ }',
+      'if ((c1 || c2) && (true)) { a++ }'
+    );
+  });
+
+  it('should mutate (c1 || c2 || c3)', () => {
```
nicojs, line 117: This test ist failing in CI

packages/instrumenter/test/helpers/expect-mutation.ts
```diff
@@ -52,6 +52,10 @@ export function expectJSMutation(sut: NodeMutator, originalCode: string, ...expe
       }
     },
   });
-  expect(mutants).lengthOf(expectedReplacements.length);
-  expectedReplacements.forEach((expected) => expect(mutants, `was: ${mutants.join(',')}`).to.include(expected));
+  /* eslint-disable @typescript-eslint/require-array-sort-compare */
+  /* because we know mutants and expectedReplacements are strings */
+  mutants.sort();
```
nicojs, line 57: Good idea!

packages/instrumenter/src/mutators/conditional-expression-mutator.ts
```diff
@@ -14,6 +14,22 @@ export const conditionalExpressionMutator: NodeMutator = {
       yield types.booleanLiteral(true);
       yield types.booleanLiteral(false);
     } else if (isBooleanExpression(path)) {
+      if (path.parent && path.parent.type === 'LogicalExpression') {
```
nicojs, line 17: Can we use optional chaining here? `path.parent?.type === 'LogicalExpression'`

nicojs (changes requested):
Looks incredibly simple if you put it like this 🙂

Got some small remarks. Also CI is failing

nicojs:
> Maybe this test should simply be removed from this pull request, and a new issue created for this issue.

Yes, that's probably best. Probably a good idea to discuss it in a separate issue as well, as I think the test cases are already covered largely by the `logicalOperatorMutator`

edi9999:
@nicojs  is there a way to update the e2e tests quickly ?

I'm running : 

```
npm run e2e
```

Since it stops after the first fail, I have to manually relaunch the tests and copy paste the output.

nicojs:
You can run them manually by running `npm test` in an e2e/test/xxx directory. But every time you change something in Stryker, you need to reinstall with `npm install` from the e2e root dir

edi9999:
Tests are now passing, can you please review @nicojs ?

nicojs (approved):
LGTM 👍

PR #3353, opened 2022-01-14
feat(hot-reload): implement hot-reload in jasmine and mocha

### LINKED ISSUE #2413: Implement hot reload for mocha test runner
As discussed with  @Lakitna  in #1514

> **Hot reload**. Test runners won't have to clear their test suite between tests (there is a caveat here regarding static code, see hurdles).

We want to implement this feature in the mocha test runner, which supports hot reload since ~7.2 (see https://github.com/mochajs/mocha/pull/4234). 

I want to make it optional, since it might break some test suites that don't cleanup nicely between test runs.

For example:

```js
describe('foo', () => {
  const myStub = sinon.stub(); // Bad practice, will be used across runs
  it('should work', () => {
    myStub();
    sinon.assert.calledOnce(spy)
  });
});
```

This test will succeed the first time, but fail the second time. This would mark all mutants as killed.

Suggestions:

1. `"hotReloadEnabled": true` 
    `false` is default.`
2.  `"hotReload": true` 
    `false` is default`
3.  `"reloadMode": "hot"`
    `"reload"` by default.  Leaves room for adding other modes in the future (for example "restart", which restarts the test runner process between each run)

nicojs:
TODO:

- [x] Add `TestRunnerCapabilities` interface and `capabilities()` method to `TestRunner` interface
- [x] Add `reloadEnvironment` boolean to the `MutantRunOptions`
- [x] Implement hot-reload in mocha
- [ ] Implement hot-reload in jasmine
- [x] Retrieve and store the test runner capabilities in Stryker Core.
- [x] Support `reloadEnvironment: true` in Stryker's core: correctly set the `reloadEnvironment` boolean
- [x] Support `reloadEnvironment: false` in Stryker's core: restart the test runner.
- [ ] Document
- [ ] Determine migration path.
- [ ] Add e2e test for an esm project with mocha

Fixes #2413


nicojs:
I've since migrated this to other PRs and have been merged.

PR #3358, opened 2022-01-17
Fix declarationDir is inside compilerOptions not on root config

kamontat:


packages/typescript-checker/test/unit/typescript-helpers.spec.ts
```diff
@@ -101,13 +101,7 @@ describe('typescript-helpers', () => {
...
           )
         ).compilerOptions
-      ).deep.include({
-        noEmit: true,
-        incremental: false,
-        composite: false,
-        declaration: false,
-        declarationMap: false,
-      });
+      ).not.hasOwnProperty('declarationDir');
```
nicojs, line 104: Great find! So we had a test for it, but it never worked correctly? 😢

nicojs:
Thanks for the PR. You can use `npm run lint -- --fix` to fix linting If I'm not mistaken.

packages/typescript-checker/test/unit/typescript-helpers.spec.ts
kamontat: Yes, I just spotted this as well, because previous test only look for what is there, but not check what should missing.

nicojs: 🤦‍♂️

nicojs: Serves me right for not adding an integration test when I implemented this fix.

PR #3366, opened 2022-01-19
feat(group-mutants): checkers check groups of mutants

danny12321:
The checkers (also known as the TypeScript-checker) now receive an array of mutants. The group of mutants can be checked at once which will increase performance in larger projects.


packages/core/src/concurrent/pool.ts
```diff
@@ -12,7 +12,7 @@ const MAX_CONCURRENT_INIT = 2;
 /**
  * Represents a Checker that is also a Resource (with an init and dispose)
  */
-export type CheckerResource = Checker & Resource;
+export type CheckerResource = ActiveChecker & Checker & Resource;
```
nicojs, line 15: See the previous remark. I don't think a `CheckerResource` should be ` Checker` anymore.

packages/core/src/concurrent/pool.ts
```diff
@@ -89,6 +89,14 @@ export class Pool<TResource extends Resource> implements Disposable {
     );
   }
 
+  /**
+   * Runs a task on all individual resources
+   * @param task The task to execute on each resource
+   */
+  public async runOnAllResources(task: (resource: TResource) => Promise<void>): Promise<void> {
```
nicojs, line 96: This method isn't needed anymore when you move to the new `CheckerResource` proposal.

packages/core/src/process/4-mutation-test-executor.ts
```diff
@@ -91,33 +91,76 @@ export class MutationTestExecutor {
...
+  private async executeChecker(
+    checkerType: string,
+    previousPassedMutants$: Observable<MutantTestCoverage>,
+    checkResult$: Subject<MutantResult>,
+    passedMutant$: Subject<MutantTestCoverage>
+  ) {
+    const mutants = await lastValueFrom(merge(previousPassedMutants$).pipe(toArray()));
+
+    // Set the active checker on all checkerWorkers
+    await this.checkerPool.runOnAllResources(async (checkerResource) => {
```
nicojs, line 138: Can be removed when moved to the new `CheckerResource` proposal.

packages/core/src/process/4-mutation-test-executor.ts
```diff
@@ -91,33 +91,76 @@ export class MutationTestExecutor {
...
+        if (result.checkResult.status === CheckStatus.Passed) {
+          passedMutant$.next(result.mutant);
+        } else {
+          checkResult$.next(this.mutationTestReportHelper.reportCheckFailed(result.mutant, result.checkResult));
+        }
+      });
+    });
+
+    await lastValueFrom(run$);
+    passedMutant$.complete();
```
nicojs, line 163: I think we might be able to work without custom Subjects, instead work with pure RxJS pipes. That way we also don't need to manually `complete`.

packages/core/src/process/4-mutation-test-executor.ts
```diff
@@ -91,33 +91,76 @@ export class MutationTestExecutor {
...
+    };
+  }
+
+  private async executeChecker(
+    checkerType: string,
+    previousPassedMutants$: Observable<MutantTestCoverage>,
+    checkResult$: Subject<MutantResult>,
+    passedMutant$: Subject<MutantTestCoverage>
+  ) {
+    const mutants = await lastValueFrom(merge(previousPassedMutants$).pipe(toArray()));
```
nicojs, line 135: This is fine I think for now. We might need to group them using an operator like [`bufferTime`](https://rxjs.dev/api/operators/bufferTime) when we have support more checker implementations.

packages/typescript-checker/package.json
```diff
@@ -25,7 +25,9 @@
   "author": "Nico Jansen <jansennico@gmail.com>",
   "contributors": [
     "Simon de Lang <simondelang@gmail.com>",
-    "Nico Jansen <jansennico@gmail.com>"
+    "Nico Jansen <jansennico@gmail.com>",
+    "Danny Berkelaar <dannyberkelaar@gmail.com>",
+    "Thierry Rietveld <thierry.rietveld0505@gmail.com>"
```
nicojs, line 30: 😎

packages/typescript-checker/src/compilers/compiler-with-watch.ts
```diff
@@ -0,0 +1,177 @@
...
+      },
+      (status) => {
+        this.log.debug(status.messageText.toString());
+      },
+      () => {
+        this.currentTask.resolve();
+      }
+    );
+
+    const dependencyFiles: SourceFiles = {};
```
nicojs, line 80: Probably better for performance to use a `new Map<string, SourceFile>()` here.

packages/typescript-checker/src/compilers/compiler-with-watch.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    for (const file in dependencyFiles) {
+      const imports = dependencyFiles[file].imports;
+
+      for (const importFile of imports) {
+        dependencyFiles[importFile]?.importedBy.add(file);
+      }
+    }
+
+    function filterDependency(file: ts.SourceFile) {
```
nicojs, line 117: Can we move this to a `helper` or something? Also might need a new name. What are we filtering?

packages/typescript-checker/src/compilers/compiler-with-watch.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  public async check(): Promise<ts.Diagnostic[]> {
+    await this.currentTask.promise;
+    const errors = [...this.currentErrors];
+    this.currentTask = new Task();
+    this.currentErrors = [];
+    return errors;
+  }
+
+  private resolveFilename(fileName: string): string[] {
```
nicojs, line 134: Rename suggestion: `findMappedSourceFiles` or something like that. It's not simply `resolving` a given file.

packages/typescript-checker/src/compilers/compiler.ts
```diff
@@ -0,0 +1,15 @@
...
+export type SourceFiles = Record<
+  string,
+  {
+    fileName: string;
+    imports: Set<string>;
+    importedBy: Set<string>;
+  }
+>;
+
+export interface TypescriptCompiler {
```
nicojs, line 12: How does interface differ from `I<CompilerWithWatch>`?

packages/typescript-checker/src/fs/hybrid-filesystem.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  public writeFile(fileName: string, content: string): ScriptFile {
+    fileName = toPosixFileName(fileName);
+    const existingFile = this.files.get(fileName);
+    if (existingFile) {
+      existingFile.write(content);
+      return existingFile;
+    } else {
+      // this.log.trace('Writing to file "%s"', fileName);
```
nicojs, line 38: This `trace` can go?

packages/typescript-checker/src/typescript-checker.ts
```diff
@@ -1,201 +1,144 @@
...
-   * @param mutant The mutant to check
-   */
-  public async check(mutant: Mutant): Promise<CheckResult> {
-    if (this.fs.existsInMemory(mutant.fileName)) {
-      this.clearCheckState();
-      this.fs.mutate(mutant);
-      return this.currentTask.promise;
-    } else {
-      // We allow people to mutate files that are not included in this ts project
+  public async check(mutants: MutantTestCoverage[]): Promise<Array<{ mutant: MutantTestCoverage; checkResult: CheckResult }>> {
```
nicojs, line 55: Probably we can split this into an `optimisticCheck` (check the entire group) and `pessimisticCheck` (check mutants one by one) method?

packages/typescript-checker/src/typescript-checker.ts
```diff
@@ -1,201 +1,144 @@
...
-        status: CheckStatus.Passed,
+        mutant,
+        errors: [],
       };
-    }
-  }
+    });
+
+    // We allow people to mutate files that are not included in this ts project
+    mutants = mutants.filter((mutant) => this.fs.existsInMemory(mutant.fileName));
```
nicojs, line 64: Only the mutants that exist in memory will remain. If this mutant lives inside a file that isn't included in the TS project, it will not be reported whatsoever, also not in a `passed` result, right? Maybe we should also add an integration test for it.

packages/typescript-checker/src/group.ts
```diff
@@ -0,0 +1,64 @@
+import { Mutant, MutantTestCoverage } from '@stryker-mutator/api/core';
+
+import { SourceFiles } from './compilers/compiler';
+
+import { toPosixFileName } from './fs/tsconfig-helpers';
+
+export function createGroups(sourceFiles: SourceFiles, mutants: MutantTestCoverage[]): MutantTestCoverage[][] {
```
nicojs, line 7: This probably deserves a JS doc explaining what we're doing and why.

packages/typescript-checker/src/group.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const group: Array<{ fileName: string; mutant: MutantTestCoverage }> = [{ fileName: firstMutantFilename, mutant: firstMutant }];
+
+    // We allow people to mutate files that are not included in this ts project
+    if (!firstSourceFile) {
+      addGroupToList();
+      continue;
+    }
+
+    let ignoreList = [firstSourceFile.fileName, ...firstSourceFile.importedBy];
```
nicojs, line 24: Can be a `const`. Instead of using `ignoreList = [...ignoreList, val, val2]` you can use `ignoreList.push(val, val2)`. Also: you probably want to use a `Set`, because it is much more performant when checking for existence.

packages/typescript-checker/src/group.ts
```diff
@@ -0,0 +1,64 @@
...
+import { SourceFiles } from './compilers/compiler';
+
+import { toPosixFileName } from './fs/tsconfig-helpers';
+
+export function createGroups(sourceFiles: SourceFiles, mutants: MutantTestCoverage[]): MutantTestCoverage[][] {
+  let mutantsWithoutGroup = [...mutants];
+  let groups: MutantTestCoverage[][] = [];
+
+  while (mutantsWithoutGroup.length) {
+    const firstMutant = mutantsWithoutGroup[0];
```
nicojs, line 12: Suggestion: use `shift` or `pop` to get a mutant from the list. That way, you also removed it (and don't need to start from `1` later).

packages/typescript-checker/src/group.ts
```diff
@@ -0,0 +1,64 @@
...
+        group.push({ fileName: activeSourceFile.fileName, mutant: activeMutant });
+        ignoreList = [...ignoreList, activeSourceFile.fileName, ...activeSourceFile.importedBy];
+      }
+    }
+
+    addGroupToList();
+
+    function addGroupToList() {
+      // Can be improved, we know the index in the for loop but can't remove it because we loop over the same array we want te remove from
+      // A solution can be to store a array with indexes to remove
```
nicojs, line 45: Probably better yes.

packages/typescript-checker/src/typescript-checker.ts
```diff
@@ -1,201 +1,144 @@
...
-        this.allTSConfigFiles.add(referencedProject);
+    const errors = await this.typeCheckMutants(mutants);
+    const mutantsToTestIndividual = new Set<Mutant>();
+
+    errors.forEach((error) => {
+      const possibleMutants = this.matchMutantsFromError(mutants, error);
+      if (possibleMutants.length === 1) {
+        mutantResults[mutantResults.findIndex((mutantResult) => mutantResult.mutant.id === possibleMutants[0].id)].errors.push(error);
+      } else if (possibleMutants.length > 1) {
+        possibleMutants.forEach((mutant) => mutantsToTestIndividual.add(mutant));
```
nicojs, line 74: You probably want to log on debug level here. Useful information for debugging performance issues later.

nicojs (changes requested):
Wow, that seems to be about a half years work condensed into 1 PR 😅. Well done! 

This is my first round of review comments. I didn't open in VSCode yet, so it was a colorful experience. I will have a closer look in the second round (after fixes are implemented). No time pressure, let's make sure we get it right.

Also: you probably want to merge `epic/esm` into this branch. Some of the changes will go away as well as the merge conflict.

e2e/test/babel-transpiling/verify/verify.ts
```diff
@@ -3,10 +3,10 @@ import { expectMetricsJson } from '../../../helpers';
 describe('Verify stryker has ran correctly', () => {
   it('should report expected score', async () => {
     // File       | % score | # killed | # timeout | # survived | # no cov | # error |
-    // All files  |   57.45 |       27 |         0 |         20 |        0 |       1 |
+    // All files  |   55.56 |       25 |         0 |         20 |        0 |       1 |
     await expectMetricsJson({
-      killed: 27,
-      mutationScore: 57.45,
+      killed: 25,
```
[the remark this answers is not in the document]
danny12321: This change came from an upstream fetch.

package.json
```diff
@@ -16,17 +16,17 @@
     "@types/sinon": "^10.0.0",
     "@types/sinon-chai": "~3.2.4",
     "@types/source-map-support": "~0.5.3",
-    "@typescript-eslint/eslint-plugin": "~5.9.1",
-    "@typescript-eslint/parser": "~5.9.1",
+    "@typescript-eslint/eslint-plugin": "~5.10.0",
+    "@typescript-eslint/parser": "~5.10.0",
```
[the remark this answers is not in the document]
danny12321: This change also came from an upstream fetch.

nicojs:
I'm thinking of the best way to merge this in. Since this is a very big change and it already diverged from the base branch a lot. I mean, we've migrated to ESM in the meantime, literally changing _very code file in our repository_ 😇.

We might be better off abandoning this branch, and instead, create fresh ones from `epic/esm`. To not make it big I think we should split it in 2 PR's:

1. Changes in API and core, but without grouping in the TyperScript. So simply don't implement the `group` method, and do minimal changes to make the old implementation work with the new API.
2. Add grouping algorithm and type checking to the TypeScript checker.

@danny12321 @ThierryRietveld do you agree with this approach? We've planned some more Stryker days next week Friday and the following Monday, maybe we can discuss it then?

I've also sent you guys invites to become contributors on StrykerJS. That way you can work on the `stryker-mutator` git remote directly. That will help with collaborating.

ThierryRietveld:
Hi Nico, thanks for the invitation to join the Stryker team! @danny12321 and I agree that splitting the PR into two separate PR's is the best option to make it smaller. Next week Friday and Monday we got 'Open Leercentrum', so I don't know if we can get some time to participate in the Stryker days. This week I got some time to start the separation of this PR.

nicojs:
Ok, great. We can at least have a talk via teams. I'll plan a meeting. See you then, have fun in the mean time! 🤗

nicojs:
@ThierryRietveld @danny12321 is this PR still relevant? Or will you create a new one?

Issue #2922, opened 2021-05-27
Supporting native ES imports for jasmine and mocha


nicojs:
**Is your feature request related to a problem? Please describe.**

Native ECMAScript modules or ESM for short, are a new way to import files using the `import` and `export` keywords. NodeJS 12+ supports this out of the box, see https://nodejs.org/api/esm.html for more info. Contrary to CommonJS modules, or CJS for short, which is the way nodejs has always worked (`module.exports` and `require`). Both [Mocha](https://mochajs.org/#nodejs-native-esm-support) and [Jasmine](https://jasmine.github.io/setup/nodejs.html#using-es-modules) support ESM. 

However, although Mocha and Jasmine support them, our `@stryker-mutator/mocha-runner` and `@stryker-mutator/jasmine-runner` plugins still need to clear files from esm cache between runs. For CJS, this is done using the [DirectoryRequireCache](https://github.com/stryker-mutator/stryker-js/blob/master/packages/util/src/directory-require-cache.ts) helper class. However, clearing files from cache is a feature that is [not supported with EMS](https://github.com/nodejs/modules/issues/307). Since reloading the files is important for both Mocha and Jasmine, to support ESM, you will currently have to run with [`--maxTestRunnerReuse`](https://stryker-mutator.io/docs/stryker-js/configuration#maxtestrunnerreuse-number) set to `1`, so a fresh process is created for each test run.

There are 2 reasons why all files currently always need to reload between runs:
1. These test runner plugins currently don't support what we call ["hot-reload"](https://github.com/stryker-mutator/stryker-js/issues/2413), this means that mocha and jasmine can only run once after all files are loaded.
2. Some mutants are executed as soon as your file _is loaded_, instead of during test execution. We call these mutants _static mutants_, and all your tests are executed as a result. For example:
    ```diff
    // hello.js
    -export const hello ='👋'
    +export const hello = ''
    ```
    Mutation switching will place this mutant correctly, but since the file is only loaded once, it will never execute after being activated. Stryker _does_ know exactly which mutant is static, as long as [`--coverageAnalysis`](https://stryker-mutator.io/docs/stryker-js/configuration/#coverageanalysis-string) is _not_ `"off"`

**Describe the solution you'd like**
I can think of a couple of features to improve ESM support:
1. Implement hot reload for both jasmine and mocha (this is a minimal requirement).
[...]

PR #3369, opened 2022-01-20
feat(reload test environment): implement test environment reload 

nicojs:
Add the ability for Stryker to restart a test runner in order to reload its test environment. This is used to test static mutants where the test runner is incapable of reloading the test environment by itself. This is a requirement to support es modules (ESM) since it is impossible to reload an ESM using something like `delete require.cache['...']`. Instead, Stryker will restart the entire test runner process for those mutants.

The way for the test runner to communicate to Stryker whether it is capable of reloading the environment is by implementing the new `capabilities()` method.

```ts
capabilities() {
  return { reloadEnvironment: true }
}
```

If you do this, you can expect the new `reloadEnvironment` boolean in your `MutantRunOptions` to be true when you need to reload the test environment after activating the mutant.

See #2413 and #2922

I also moved some code around (from the `MutationTestProcess` to a new class `MutantTestPlanner`).

BREAKING CHANGE: Test runner plugins must provide `TestRunnerCapabilities` by implementing the `capabilities` method.


nicojs:
@simondel do you think it is OK to make `capabilities()` required?

nicojs:
This change also contains a breaking change for the reporter API:

```diff
- export type MutantTestCoverage = Mutant &
-  Pick<schema.MutantResult, 'coveredBy' | 'static'> & {
-    estimatedNetTime: number;
-    hitCount?: number;
-    testFilter: string[] | undefined;
-  };
+ export type MutantTestCoverage = Mutant & Pick<schema.MutantResult, 'coveredBy' | 'static'>;
```

I've removed `estimatedNetTime`, `hitCount`, and `testFilter`. They weren't used by our reporters and they were only there to pass along state internally in `core`, between the `find-mutant-test-coverage` and `mutation-test-executor`. I found this very obfuscated and it frustrated me. 

That's why I now choose to do all the work in the new `MutantTestPlanner` class. It is responsible for planning which mutant should be run with which options and which should be reported as an early result. 

We can in a later PR replace the `onAllMutantsMatchedWithTests` with a new `onMutantPlanDetermined`, which can include the run options per test. That way we can improve the `progess` reporter. Instead of counting the same progress per mutant, it can count mutants that require a test env reload 3 times for example.

packages/core/src/mutants/mutant-test-planner.ts
```diff
@@ -0,0 +1,211 @@
...
+      };
+    } else if (this.dryRunResult.mutantCoverage) {
+      // If there was coverage information (coverageAnalysis not "off")
+      const tests = this.testsByMutantId.get(mutant.id) ?? [];
+      const coveredBy = toTestIds(tests);
+      if (!isStatic || (this.options.ignoreStatic && coveredBy.length)) {
+        // If not static, or it was "hybrid" (both static and perTest coverage) and ignoreStatic is on.
+        // Only run covered tests
+        const estimatedNetTime = calculateTotalTime(tests);
+        return {
```
simondel, line 82: It might be nice to have a factory/create method for these `MutantTestPlan` objects because current function is quite long

packages/core/src/mutants/mutant-test-planner.ts
```diff
@@ -0,0 +1,211 @@
...
+            testsByMutantId.set(mutantId, tests);
+          }
+          tests.add(foundTest);
+        }
+      });
+    });
+  return testsByMutantId;
+}
+
+function calculateTotalTime(testResults: Iterable<TestResult>): number {
```
simondel, line 178: Isn't this just a `reduce` function?

packages/core/src/mutants/mutant-test-planner.ts
```diff
@@ -0,0 +1,211 @@
...
+  let total = 0;
+  for (const test of testResults) {
+    total += test.timeSpentMs;
+  }
+  return total;
+}
+
+function toTestIds(testResults: Iterable<TestResult>): string[] {
```
simondel, line 186: Isn't this just a `map` function?

simondel (approved):

packages/core/src/mutants/mutant-test-planner.ts
```diff
@@ -0,0 +1,211 @@
...
+    logger: Logger
+  ) {
+    this.testsByMutantId = findTestsByMutant(this.dryRunResult.mutantCoverage?.perTest, dryRunResult.tests, logger);
+    this.hitsByMutantId = findHitsByMutantId(this.dryRunResult.mutantCoverage);
+    this.timeSpentAllTests = calculateTotalTime(this.dryRunResult.tests);
+  }
+
+  public makePlan(mutants: readonly Mutant[]): readonly MutantTestPlan[] {
+    const mutantTestPlans = mutants.map((mutant) => this.planMutant(mutant));
+    this.reporter.onAllMutantsMatchedWithTests(mutantTestPlans.map(({ mutant }) => mutant));
```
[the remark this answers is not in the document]
nicojs: Indeed, a nitpick 😉

packages/core/src/mutants/mutant-test-planner.ts
nicojs: Sure thing

packages/core/src/mutants/mutant-test-planner.ts
nicojs: Yes, but it is called with both a `Set` as an `Array`. I didn't want to create a bogus array just to be able to call `reduce`. This is better.

packages/core/src/mutants/mutant-test-planner.ts
nicojs: Same as above. Note that `reduce` and `map` might be coming to Iterables in an ES specification near you: https://github.com/tc39/proposal-collection-methods#proposal

packages/core/src/mutants/mutant-test-planner.ts
nicojs: I've created factory methods for early result as well as for mutant plan

PR #3371, opened 2022-01-21
Add comment

JorisNienkemper:


packages/core/src/checker/checker-child-process-proxy.ts
```diff
@@ -8,6 +8,7 @@ import { Resource } from '../concurrent/pool';
 
 import { CheckerWorker } from './checker-worker';
 
+/* not documented should have a better documentation */
 export class CheckerChildProcessProxy implements Checker, Disposable, Resource {
   private readonly childProcess: ChildProcessProxy<CheckerWorker>;
```
JorisNienkemper, line 13: Child process has etc.

nicojs:
Thanks, but: no thanks.

PR #3396, opened 2022-01-28
feat(esm): support native es modules in the jasmine runner.

### LINKED ISSUE #2922: Supporting native ES imports for jasmine and mocha

### LINKED ISSUE #3340: Support Jasmine 4

nicojs:
* Uses `autoCleanClosures: false` in order to reuse jasmine instances between runs. This comes with a big performance boost (4x)
* Implements hot reload in jasmine.
* Sets `reloadEnvironment` capabilities to `false`. Meaning that static mutants will always force a reload of the environment. This comes with a performance penalty. If you don't want that, please run with `--ignoreStatic`.
* Add support for jasmine 4 (Closes #3340)

Together with #3393  closes #2922

BREAKING CHANGE: The `@stryker-mutator/jamsine-runner` now requires jasmine@3.10 or higher.


e2e/package.json
```diff
@@ -21,8 +21,8 @@
     "cross-env": "~7.0.3",
     "enzyme-to-json": "^3.6.2",
     "grunt": "~1.4.1",
-    "jasmine": "~3.10.0",
-    "jasmine-core": "~3.10.1",
+    "jasmine": "^4.0.2",
```
simondel, line 24: Do we not want to pin a minor version?

simondel: Same goes for other package.json's

simondel (approved):

PR #3432, opened 2022-02-15
feat(esm config): support config file as pure esm

### LINKED ISSUE #3408: Support stryker.conf[.cjs|.mjs] by default
Since we're planning to support native ESM from Stryker v6 onward I think it makes sense to also support `stryker.conf.mjs` by default (when stryker.conf.json and stryker.conf.js is missing). This will allow projects to define their config as esm. While we're at it, we might as well support `stryker.conf.cjs` .

The config loader should handle these new extensions.

Not supporting these extensions, but that is not a good user experience IMO.

**Additional context**
This should be merged in the `epic/esm` branch.

nicojs:
Support your `stryker.conf.js` file to be a pure esm when `{ "type": "module" }` in your package.json file.

Also add support for 2 new default configuration file names: `stryker.conf.mjs` and `stryker.conf.cjs`.

- [x] Implement new algorithm in the config reader
- [x] Add unit tests and update integration tests
- [x] Update the docs to include `export default { }`
- [x] Update `stryker init` to include auto-detect package types and choose the correct format.

BREAKING CHANGE: `module.exports = function(config) {}` is no longer supported in your stryker.conf.js file. This was already deprecated, but now will give an error.

Closes #3408


nicojs:
@simondel I've decided that `stryker init` will output a `stryker.conf.mjs` file from Stryker 6 onward. That way we don't need to support 2 js output formats in the `stryker init` command. 

Loading `stryker.conf.mjs` files is supported from nodejs 12 onward, so it should be no problem for Stryker@6

PR #3456, opened 2022-03-04
fix: set correct background color for html report

hugo-vrijswijk:
I noticed the background color of the index.html is different than the actual report element. This PR changes that to use whatever background-color the html reports it is currently using


hugo-vrijswijk:
Lots of compile errors, but I think it's unrelated to this change

nicojs:
Yeah... fixed this in the `epic/esm` branch. I hate maintaining multiple branches 🙄

PR #3470, opened 2022-03-28
test(checker-facade): Added tests for checker-facade.ts

### LINKED ISSUE #3459: Test the `CheckerFacade`
We've merged #3450 early so we could release the beta. Please add unit tests for `src/checker/checker-facade.ts` as well as at least 1 integration test.

@danny12321 @ThierryRietveld do you think one of you can pick this up?

danny12321:
Added tests for ```checker-facade.ts``` added with the 6.0 beta version. Got 100% mutation score and fixed a minor bug 😎.

Closes #3459

### DIFF
--- a/packages/core/test/unit/checker/checker-facade.spec.ts
+++ b/packages/core/test/unit/checker/checker-facade.spec.ts
@@ -1,0 +1,108 @@
+import { CheckStatus } from '@stryker-mutator/api/check';
+import { factory } from '@stryker-mutator/test-helpers';
+
+import { expect } from 'chai';
+import sinon from 'sinon';
+
+import { CheckerFacade } from '../../../src/checker/checker-facade.js';
+import { CheckerResource } from '../../../src/checker/index.js';
+
+import { createMutantRunPlan } from '../../helpers/producers.js';
+
+describe.only(CheckerFacade.name, () => {
+  let innerChecker: sinon.SinonStubbedInstance<CheckerResource>;
+  let sut: CheckerFacade;
+
+  beforeEach(() => {
+    innerChecker = {
+      init: sinon.stub(),
+      group: sinon.stub(),
+      check: sinon.stub(),
+      dispose: sinon.stub(),
+    };
+    const checkers = [innerChecker];
+    sut = new CheckerFacade(() => checkers.shift()!);
+  });
+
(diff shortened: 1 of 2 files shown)


packages/core/test/unit/checker/checker-facade.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  beforeEach(() => {
+    innerChecker = {
+      init: sinon.stub(),
+      group: sinon.stub(),
+      check: sinon.stub(),
+      dispose: sinon.stub(),
+    };
+    const checkers = [innerChecker];
+    sut = new CheckerFacade(() => checkers.shift()!);
```
nicojs, line 24: Why the use of an array here?

packages/core/test/unit/checker/checker-facade.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+import { expect } from 'chai';
+import sinon from 'sinon';
+
+import { CheckerFacade } from '../../../src/checker/checker-facade.js';
+import { CheckerResource } from '../../../src/checker/index.js';
+
+import { createMutantRunPlan } from '../../helpers/producers.js';
+
+describe(CheckerFacade.name, () => {
```
nicojs, line 12: I see you've got tests for `check()` and `group()` in here, could you separate them out into their own inner `describe`s?

nicojs (changes requested):
Awesome to see these tests! Just have 2 small remarks, would you please take a look?

nicojs:
Closes #3459

packages/core/test/unit/checker/checker-facade.spec.ts
danny12321: Because I thought the lambda paramater would need to return `null` after all checkers where loaded, but it seems that isn't the case. I would change this in next commit.

nicojs (approved):

packages/core/test/unit/checker/checker-facade.spec.ts
nicojs: No, indeed. The factory method is called whenever the decorator or facade wants to create a new inner resource. Thanks for updating!

PR #3472, opened 2022-03-29
fix(core): report failed check results only once

### LINKED ISSUE #3461: Progress reporter shows more mutants are done then there are available
**Summary**
When running Stryker beta 6.0.0 the progress reporter shows that there are more mutants done then there are available (see image below). Sometimes it will still complete the run with no other issues but other times it will get stuck. 

image

**Stryker environment**

```
+-- @stryker-mutator/core@6.0.0-beta.0
```

```
+-- mocha@6.0.0-beta.0
```

**Test runner environment**

```shell
# cd packages/core/ && node bin/stryker.js run -m src/checker/checker-facade.ts
```

**Your Environment**

| ---------------- | -------
| node             | 16.14.0
| npm              | 8.3.1
| Operating System | Windows 10

**Add stryker.log**

[stryker.log](https://github.com/stryker-mutator/stryker-js/files/8213080/stryker.log)

nicojs:
Share the replay of failed check results accross all subscribers using the `shareReplay()` rxjs operator.

Reporting of a failed checks is done using a side effect (with the `tap()` operator). When later we subscribe to it twice, which is a result of the `partition()` operator, the reporting was done twice.

Fixes #3461

### DIFF
--- a/packages/core/src/process/4-mutation-test-executor.ts
+++ b/packages/core/src/process/4-mutation-test-executor.ts
@@ -117,7 +117,10 @@ export class MutationTestExecutor {
     let passedMutant$ = input$;
     for (const checkerName of this.options.checkers) {
       // Use this checker
-      const [checkFailedResult$, checkPassedResult$] = partition(this.executeSingleChecker(checkerName, passedMutant$), isEarlyResult);
+      const [checkFailedResult$, checkPassedResult$] = partition(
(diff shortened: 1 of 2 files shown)


danny12321 (approved):
This seems the have fixed the problem of #3461.

nicojs:
Thanks @danny12321 !

PR #3474, opened 2022-03-29
Feature/eslint checker plugin

TimLehner:
Inspired by the existing Typescript Checker Plugin. This checker is a more bare-bones version of the typesript plugin, and doesn't have much in the way of customisability or testing yet. 

I have this working for my personal project which uses next-js+typescript, but I haven't tried this using other projects like create-react-app or bundlers like webpack.

I am also unsure what the best way to add the types for `StrykerOptions`, I wish to add a `lintConfigFile?: string`, but I only saw a big interface defined in src_generated which I thought I shouldn't touch.

Finally apologies I didn't open an issue before, I only just started on this earlier today and it all came together quite quickly


TimLehner:
Progress:

- [x] Implementation
- [x] unit tests
- [x] integration tests
- [x] e2e tests:
  - [x] mocha-javascript-eslint
  - [ ] jest-js-eslint
  - [ ] jest-ts-eslint
  - [ ] jest-old-version-eslint
  - [ ] other e2e tests?

nicojs:
I don't think it is needed to have an eslint e2e test for each test runner. My reasoning is: checkers are pretty isolated from test runners. The added value is minimal.

Rather I'm now thinking the following e2e tests:

- A JS project with eslint recommended rule set
- A JS project with [eslint type-aware recommended](https://typescript-eslint.io/docs/linting/type-linting/) rules active 
- A TS project with the typescript-checker AND [eslint type-aware recommended](https://typescript-eslint.io/docs/linting/type-linting/) rules active.

@TimLehner what do you think?

TimLehner:
Yes, that sounds much better 👍

TimLehner:
@nicojs I have implemented the requested e2e tests. I think this is now ready for a proper review

nicojs:
@TimLehner I've got time to look at this tomorrow and next Monday. The build is failing because of some dependency stuff, which I've fixed yesterday on master. Would you mind merging those changes in your branch?

nicojs:
Thanks for the push rights! Will be looking at this today 👀

packages/eslint-checker/README.md
```diff
@@ -0,0 +1,51 @@
...
+## `lintConfigFile` [`string`]
+
+Default: `.eslintrc.*`
+
+The path to your [eslint configuration file](https://eslint.org/docs/user-guide/configuring/configuration-files#using-configuration-files).
+
+_Note: the following eslint options are always overridden by @stryker-mutator/eslint-checker to avoid false positives, based on a similar reasoning to that in [@stryker-mutator/typescript-checker](https://stryker-mutator.io/docs/stryker-js/typescript-checker/)_
+
+```json
+{
```
simondel, line 36: I think it would also be smart to disable https://eslint.org/docs/rules/max-len due to the fact that mutation switching will result in (much) longer lines of code

packages/eslint-checker/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "stryker",
+    "checker",
+    "plugin"
+  ],
+  "author": "Tim Lehner <timothy.lehner.2011@live.rhul.ac.uk>",
+  "homepage": "https://github.com/TimLehner/stryker-js#readme",
+  "license": "ISC",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/TimLehner/stryker-js.git"
```
simondel, line 18: This url (and other urls) will have to be changed prior to merging

packages/eslint-checker/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/TimLehner/stryker-js/issues"
+  },
+  "dependencies": {
+    "@stryker-mutator/api": "^6.0.0-beta.0",
+    "@stryker-mutator/util": "6.0.0-beta.0",
+    "cosmiconfig": "^7.0.1",
+    "fast-glob": "^3.2.11",
+    "typescript": "^4.6.3"
```
simondel, line 36: I can't find typescript as a direct dependency in your source files. Am I overlooking something or is can this be part of the devDependencies?

packages/eslint-checker/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "test:integration": "mocha 'dist/test/integration/**/*.js'",
+    "stryker": "node ../core/bin/stryker.js run"
+  },
+  "bugs": {
+    "url": "https://github.com/TimLehner/stryker-js/issues"
+  },
+  "dependencies": {
+    "@stryker-mutator/api": "^6.0.0-beta.0",
+    "@stryker-mutator/util": "6.0.0-beta.0",
+    "cosmiconfig": "^7.0.1",
```
simondel, line 34: @nicojs (how) do we want to set these dependency version? As any part of this major version? Or fixed to a minor/patch version?

packages/eslint-checker/src/esconfig-helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+    ...parsedConfig.config?.rules,
+    ...LINT_RULES_OVERRIDES,
+  };
+
+  const config = {
+    ...parsedConfig.config,
+    rules,
+  };
+
+  if (config.extends) {
```
simondel, line 28: I think this if check can be removed:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#description

>If the property which you are trying to delete does not exist, delete will not have any effect and will return true.

packages/eslint-checker/src/result-helpers.ts
```diff
@@ -0,0 +1,22 @@
+import { CheckResult, CheckStatus, FailedCheckResult } from '@stryker-mutator/api/check';
+import { ESLint } from 'eslint';
+
+export function isFailedResult(result: CheckResult): result is FailedCheckResult {
+  return result.status === CheckStatus.CompileError;
+}
+
+export async function makeResultFromLintReport(reports: ESLint.LintResult[], formatter: ESLint.Formatter): Promise<CheckResult> {
+  const failures = reports.filter((_result) => _result.errorCount > 0);
```
simondel, line 9: I think the `_` can be removed in the param name `_result`

nicojs (changes requested):
This looks clean and promising. Still some work to be done, though. I agree with @simondel 's remarks unless I say so in the comments

packages/eslint-checker/package.json
nicojs: I think `typescript` is added here as a transient dependency. You need it, if you use the `HybridFileSystem`. Not the best solution IMHO, see my remarks there.

packages/eslint-checker/schema/eslint-checker-options.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "LintCheckerOptions",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "lintConfigFile": {
```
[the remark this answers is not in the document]
TimLehner: I'm not entirely sure I understand what you mean, so I've not addressed this one yet

packages/util/src/fs/hybrid-file-system.ts
```diff
@@ -3,10 +3,17 @@ import { Mutant } from '@stryker-mutator/api/core';
 import { Logger } from '@stryker-mutator/api/logging';
```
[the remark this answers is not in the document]
TimLehner: I think I lean towards having slightly duplicated code, and keeping this `fs` `async`? In which case I'll revert the extraction to utils and use `fs.readFile`

packages/eslint-checker/src/eslint-checker.ts
```diff
@@ -0,0 +1,72 @@
...
+    const formatter = await this.linter.loadFormatter();
+    return await makeResultFromLintReport(results, formatter);
+  }
+
+  public async init(): Promise<void> {
+    this.logger.debug('Running initial lint check');
+    const overrideConfig = await getConfig(configLoader, this.options.lintConfigFile as string | undefined);
+    this.linter = new ESLint({ overrideConfig });
+
+    const fileNames = await fg(this.options.mutate);
```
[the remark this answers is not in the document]
TimLehner: This sounds useful, I'll hold of on making any changes here before seeing that PR

packages/eslint-checker/README.md
TimLehner: Yes, I was thinking of getting a minimal set of rules in initially, and refining this later. Happy to add any rules you think are sensible before merging this PR

TimLehner:
I think I've addressed the majority of your comments, let me know if I've missed anything

nicojs:
I've was _finally_ able to merge # which allows us to progress here. I've updated the code, file descriptions are now being passed into the constructor.

Also did some more work:
- rename e2e tests to start with the name "eslint", that makes more sense IMO (i was looking for 'eslint')
- Ive added a e2e test for the validation schema
- I went ahead and changed the config setting to be grouped under `eslint`. That allows us to expand on eslint settings in the future. This is also what we did for mocha, karma, etc. The `tsconfigFile` is an exception, because Stryker's core also uses that setting.
- Ive added the vscode config to easily run the tests from vscode

I was looking at the report in the e2e tests, and was not happy with the result TBH:

eslint-false-negatives

As you can see, there are 3 linting errors here. However, these are all false negatives in my opinion. Maybe its better to _whitelist_ certain rules, instead of the blacklisting approach.

@TimLehner what do you think about that?

packages/util/src/fs/hybrid-file-system.ts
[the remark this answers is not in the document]
nicojs: I agree! Feel free to revert 🙏

TimLehner:
Hmm, I think you make a good point on the false negative there, whitelisting would be useful, then if there is a specific surviving mutant you dismiss due to your lint rules, you can specifically enable the relevant rule and use Stryker as normal. I do feel the current default settings are too aggressive and kill too many mutants that are valid cases of lacking tests

nicojs:
Yeah, what rules would you say are really useful to enable? What was the reason for you to embark on this journey?

TimLehner:
Apologies for the delay in my response, I've been distracted.

It's tricky, since my initial motivation were a number of rather niche react-specific ESLint rules, namely: [react-hooks/exhaustive-deps](https://www.npmjs.com/package/eslint-plugin-react-hooks)

Although I have also found it too be potentially too aggressive there. I'm not sure what your thoughts are on this use-case?

PR #3499, opened 2022-05-05
chore(dependabot): scan packages sub directories

nicojs:


nicojs:
This is not working, did some changes on the master branch instead

PR #3511, opened 2022-05-25
test(mutators): verify original node immutability

### LINKED ISSUE #3510: Add an internal lint rule to require mutators yield only cloneDeep-ed nodes
**Is your maintenance request related to a problem? Please describe.**
I was amazed to learn that all mutators must `cloneDeep` any Babel nodes they yield, to avoid post-mutation ASTs including parts of the pre-mutation AST. Very cool edge case, and also very easy to miss in code!

Suggestion: how about writing an internal lint rule that makes sure any `yield`ed node is certainly a value returned by `cloneDeep(..., true)`? https://typescript-eslint.io/docs/development/custom-rules has docs on how to write type-checker-informed lint rules.

Note: this would not be a straightforward rule to implement perfectly. It would have to keep track of where each yielded value comes from. That will get tricky for edge cases and may require using [@typescript-eslint/scope-manager](https://www.npmjs.com/package/@typescript-eslint/scope-manager). I'd suggest making an MVP of this that catches just easy-to-detect cases first.

I can't think of any.

**Additional context**
Discussed [on a Twitch stream](https://www.twitch.tv/videos/1492775616) with @nicojs ✨

nicojs:
Verifies that mutators don't accidentally reinsert original parts of the AST inside mutated parts. This has been a problem in the past, so probably best to avoid it using an assertion helper.

Fixes #3510

### DIFF
--- a/packages/instrumenter/test/helpers/expect-mutation.ts
+++ b/packages/instrumenter/test/helpers/expect-mutation.ts
@@ -44,14 +44,26 @@ export function expectJSMutation(sut: NodeMutator, originalCode: string, ...expe
     sourceType: 'module',
   });
   const mutants: string[] = [];
+  const originalNodeSet = nodeSet(ast);
+
   babel.traverse(ast, {
     enter(path) {
       for (const replacement of sut.mutate(path)) {
         const mutatedCode = generate(replacement).code;
         const beforeMutatedCode = originalCode.substring(0, path.node.start ?? 0);
         const afterMutatedCode = originalCode.substring(path.node.end ?? 0);
+        const mutant = `${beforeMutatedCode}${mutatedCode}${afterMutatedCode}`;
+        mutants.push(mutant);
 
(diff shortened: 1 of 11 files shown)


nicojs:
@JoshuaKGoldberg 53 unit tests are now failing... so yeah, we forgot a few `cloneNode(..., true)` 😅

packages/instrumenter/test/helpers/expect-mutation.ts
```diff
@@ -44,14 +44,26 @@ export function expectJSMutation(sut: NodeMutator, originalCode: string, ...expe
     sourceType: 'module',
   });
   const mutants: string[] = [];
+  const originalNodeSet = nodeSet(ast);
```
JoshuaKGoldberg, line 47: Very clean! Nice!!

nicojs:
Thanks for the review @JoshuaKGoldberg 🙏

PR #3765, opened 2022-10-03
feat(disableTypeChecks): add option 'true' to disable all type checks

### LINKED ISSUE #3741: Support `true` for `disableTypeChecks`
We should allow `true` setting for `disabledTypeChecks`. 

We're currently allowing `false` and `string`. Users expect there to be a `true` as well. With true, every supported file would be disabled (same as `'**/*.{js,ts,jsx,tsx,cjs,cts,mjs,mts,html,vue}'`)

We should make `true` the default in the future. (breaking change)

simondel:
Fixes #3741 

Adds option 'true' to the `disableTypeChecks` config option to disable all type checks


docs/configuration.md
```diff
@@ -137,13 +137,13 @@ As you can see, when you disable bail, a lot more tests get the "Killing" status
 
 _Note: Disable bail needs to be supported by the test runner plugin in order to work. All official test runner plugins (`@stryker-mutator/xxx-runner`) support this feature except for Jest. Jest always runs without --bail (see [#11766](https://github.com/facebook/jest/issues/11766)) inside Stryker, however it will report only the first failing test when disableBail=false and all failing tests when disableBail=true_
 
-### `disableTypeChecks` [`false | string`]
+### `disableTypeChecks` [`false | true | string`]
```
nicojs, line 140:
```suggestion
### `disableTypeChecks` [`boolean` | `string`]
```

packages/core/src/config/file-matcher.ts
```diff
@@ -7,17 +7,21 @@ import { normalizeFileName } from '@stryker-mutator/util';
  * A helper class for matching files using the `disableTypeChecks` setting.
  */
 export class FileMatcher {
-  private readonly pattern: string | false;
+  private readonly pattern: string | false | true;
```
nicojs, line 10:
```suggestion
  private readonly pattern: string | boolean;
```

packages/core/src/config/file-matcher.ts
```diff
@@ -7,17 +7,21 @@ import { normalizeFileName } from '@stryker-mutator/util';
...
 
-  constructor(pattern: string | false) {
-    if (pattern !== false) {
+  constructor(pattern: string | false | true) {
```
nicojs, line 12:
```suggestion
  constructor(pattern: string | boolean) {
```

nicojs (changes requested):
This looks pretty elegant. But it is not working. 

If you change this:

```diff
diff --git a/e2e/test/typescript-transpiling/stryker.conf.json b/e2e/test/typescript-transpiling/stryker.conf.json
index 77b352b0..b7fdea85 100644
--- a/e2e/test/typescript-transpiling/stryker.conf.json
+++ b/e2e/test/typescript-transpiling/stryker.conf.json
@@ -8,6 +8,7 @@
   "checkers": ["typescript"],
   "tsconfigFile": "tsconfig.json",
   "buildCommand": "npm run build",
+  "disableTypeChecks": true,
   "plugins": [
     "@stryker-mutator/mocha-runner",
     "@stryker-mutator/typescript-checker"
```

And run `npm test` there you get a bunch of these warnings: 

```
07:52:14 (9921) WARN DisableTypeChecksPreprocessor Unable to disable type checking for file "/home/nicojs/github/stryker-js/e2e/test/typescript-transpiling/package.json". Shouldn't type checking be disabled for this file? Consider configuring a more restrictive "disableTypeChecks" settings (or turn it completely off with `false`) Error: Unable to parse /home/nicojs/github/stryker-js/e2e/test/typescript-transpiling/package.json. No parser registered for .json!
```

I think we should silently ignore files that cannot be parsed from `src/disable-type-checks.ts`, that makes the most sense 🤷‍♀️

nicojs:
@simondel are you planning to pick up the last remark as well? 😇

nicojs (approved):

nicojs:
I've implemented it during hacktoberfest 🐱‍👤

PR #3776, opened 2022-10-06
Avoid indirection

karlhorky:
Hi, thanks for Stryker Mutator, really cool project! 🙌

The readme has a link to the Configuration docs page, so to avoid the indirection (and 2 clicks instead of 1), the link could go directly to the Configuration docs page (as is done in this PR)


nicojs:
Thanks for the docs improvement and the kind words 🙏

karlhorky:
Glad to help :) thanks for the merge!

Issue #3780, opened 2022-10-08
 ERROR DryRunExecutor :  TypeError: Cannot set properties of undefined (setting 'currentTestId') 


bodote:
**Summary**

wenn stryker-js is run with config listed below on  this project:
```bash
git clone git@github.com:bodote/book-monkey.git
git checkout v0.2
```
I get this error during DryRunExecution on all tests:
```
...
08:41:57 (37435) INFO DryRunExecutor Starting initial test run (karma test runner with "perTest" coverage analysis). This may take a while.
08:41:57 (37435) DEBUG DryRunExecutor Using timeout of 300000 ms.
08:41:57 (37435) DEBUG TimeoutDecorator Starting timeout timer (300000 ms) for a test run
08:41:58 (37435) TRACE ChildProcessProxy Chrome Headless 105.0.5195.125 (Mac OS 10.15.7): Executed 0 of 134 SUCCESS (0 secs / 0 secs)

08:41:58 (37435) TRACE ChildProcessProxy Chrome Headless 105.0.5195.125 (Mac OS 10.15.7) FormMessagesComponent should create an array of errormessage that fit to a "title" control  FAILED
	TypeError: Cannot set properties of undefined (setting 'currentTestId')
	    at Object.specStarted (node_modules/@stryker-mutator/karma-runner/dist/src/karma-plugins/test-hooks-middleware-21f23d35-a4c9-4b01-aeff-da9c99c3ffc0.js:5:44)
	    at <Jasmine>
	    at ZoneQueueRunner.jasmine.QueueRunner.ZoneQueueRunner.execute (node_modules/zone.js/fesm2015/zone-testing.js:692:42)
	    at <Jasmine>
Chrome Headless 105.0.5195.125 (Mac OS 10.15.7): Executed 1 of 134 (1 FAILED) (skipped 25) (0 secs / 0.018 secs)

...
``` 

Whereas if I replace `"src/app/app.component.ts",` in the stryker.conf.json by  ` "src/app/shared/*.ts"`
It works without errors.
And althought of cause the `ng test` for src/app/app.component.ts or for the hole project runs also without any hickups

**Stryker config**

```json
{
  "$schema": "./node_modules/@stryker-mutator/core/schema/stryker-schema.json",
  "_comment": "This config was generated using 'stryker init'. Please see the guide for more information: https://stryker-mutator.io/docs/stryker-js/guides/angular",
  "mutate": [
    "src/app/app.component.ts",
[...]

PR #3787, opened 2022-10-11
fix(karma-runner):  bug #3780

bodote:
- fix bug, #3780 
- add unit test for that same bug, 
- correct description  in `CONTRIBUTING.md` on how to run stryker-js locally


package-lock.json
```diff
@@ -12592,7 +12592,8 @@
       "version": "1.0.4",
```
nicojs, line 12592: Please don't include this file in the PR 😇

packages/karma-runner/src/karma-plugins/test-hooks-middleware.ts
```diff
@@ -88,6 +88,7 @@ export class TestHooksMiddleware {
       case 'jasmine':
         this.currentTestHooks = `
       window.${SHOULD_REPORT_COVERAGE_FLAG} = true;
+      window.__stryker__ = window.__stryker__ || {};
```
nicojs, line 91:
```suggestion
      window.${NAMESPACE} = window.${NAMESPACE} || {};
```

nicojs: Also, the same should be done for mocha below 👇.

Ideally, we should declare this part above and reuse it for both `mocha` and `jasmine`.

packages/karma-runner/test/unit/karma-plugins/test-hooks-middleware.spec.ts
```diff
@@ -38,6 +38,7 @@ describe(TestHooksMiddleware.name, () => {
         expect(sut.currentTestHooks)
           .contains('window.__strykerShouldReportCoverage__ = true')
           .contains('window.__stryker__.currentTestId = spec.id')
+          .contains('window.__stryker__ = window.__stryker__ || {};')
```
nicojs, line 41: 😍

nicojs (changes requested):
Thanks for picking this up. I've got some small remarks.

package-lock.json
bodote: ok , reverted to the previous version

packages/karma-runner/src/karma-plugins/test-hooks-middleware.ts
bodote: changed `__stryker__` to `${NAMESPACE }`, moved the line , together with 
`window.${SHOULD_REPORT_COVERAGE_FLAG} = true;` before the `switch` so that the change should work for mocha, too

packages/karma-runner/test/unit/karma-plugins/test-hooks-middleware.spec.ts
bodote: added the same for respective mocha unit test

nicojs:
Awesome, thanks!

Issue #3088, opened 2021-08-24
Support only run the initial test run (dry run)


karfau:
**Question**

Since our stryker run takes around 2 hours to complete we currently only run it on the default branch.
It would be nice if we could still execute just the initial test run in branches to avoid that it fails on master and we need to figure out why it broke/revert...

**Stryker environment**

```
+-- @stryker-mutator/core@^5.2.2
```

```
+-- jest@^27.0.6
```

**Additional context**

nicojs:
My hopes are that running unit tests normally is enough. You generally want to share the test runner configuration with stryker using `jest.config.js`, `.mocharc.json`, etc. That way, Stryker should have no issues running your tests.

That being said, there are still slight differences for Stryker vs your test runner. I think this will work: `npx stryker run -m not-exists.js`.

We might want to add a `--dry-run` in the future, to only run the dry run.

karfau:
Ok, i just checked what `-m` means and it refers to https://stryker-mutator.io/docs/stryker-js/configuration#mutate-string
So it would also work yo have an additional config file with `"mutate": []`, right?
I wonder if we can pass an "empty glob/list" from the cli instead of a non existing file?
Most likely will experiment with this very soon.

As a side note: maybe it's worth documenting this behavior of the option? Something like
> Passing a glob that matches no files will cause a "dryrun": The configuration will be read and the tests will be run once, but no files will be mutated.

(I'm very curios about the expected exit code when using this approach...)

nicojs:
Great, that's what I was afraid of. Please open an issue for the instrumentation error, we'll use this issue to track progress on the dry-run feature

nicojs:
We probably still want this feature.

PR #3814, opened 2022-10-28
feat: add argument --dryRunOnly to only run initial tests

florisg-infosupport:
This is to provide a better option to check if the Stryker configuration is correct. 
Prevent rollbacks, as a dry-run takes less time than a full test run.
closes: #3088

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -145,6 +145,14 @@ Config file: `"disableTypeChecks": false`
 
 Configure a pattern that matches the files of which type checking has to be disabled. This is needed because Stryker will create (typescript) type errors when inserting the mutants in your code. Stryker disables type checking by inserting `// @ts-nocheck` atop those files and removing other `// @ts-xxx` directives (so they won't interfere with `@ts-nocheck`). The default setting allows these directives to be stripped from all JavaScript and friend files in `lib`, `src` and `test` directories. You can specify a different glob expression or set it to `false` to completely disable this behavior.
 
+### `dryRun` [`boolean`]
+
+Default: `false`<br />
+Command line: `--dryRun`<br />
+Config file: `"dryRun": 5`
+
+Execute the initial test run only without doing actual mutation testing. Doing a dry run can be used to test that StrykerJS can run your test setup, for example, in CI pipelines.
+
 ### `dryRunTimeoutMinutes` [`number`]
 
 Default: `5`<br />
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -145,11 +145,11 @@ Config file: `"disableTypeChecks": false`
 
 Configure a pattern that matches the files of which type checking has to be disabled. This is needed because Stryker will create (typescript) type errors when inserting the mutants in your code. Stryker disables type checking by inserting `// @ts-nocheck` atop those files and removing other `// @ts-xxx` directives (so they won't interfere with `@ts-nocheck`). The default setting allows these directives to be stripped from all JavaScript and friend files in `lib`, `src` and `test` directories. You can specify a different glob expression or set it to `false` to completely disable this behavior.
 
-### `dryRun` [`boolean`]
+### `dryRunOnly` [`boolean`]
 
 Default: `false`<br />
-Command line: `--dryRun`<br />
-Config file: `"dryRun": 5`
+Command line: `--dryRunOnly`<br />
+Config file: `"dryRunOnly": false`
 
 Execute the initial test run only without doing actual mutation testing. Doing a dry run can be used to test that StrykerJS can run your test setup, for example, in CI pipelines.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -151,7 +151,7 @@ Default: `false`<br />
 Command line: `--dryRunOnly`<br />
 Config file: `"dryRunOnly": false`
 
-Execute the initial test run only without doing actual mutation testing. Doing a dry run can be used to test that StrykerJS can run your test setup, for example, in CI pipelines.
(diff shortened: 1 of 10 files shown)


nicojs (changes requested):
Looks good, just some small remarks.

docs/configuration.md
```diff
@@ -145,6 +145,14 @@ Config file: `"disableTypeChecks": false`
 
 Configure a pattern that matches the files of which type checking has to be disabled. This is needed because Stryker will create (typescript) type errors when inserting the mutants in your code. Stryker disables type checking by inserting `// @ts-nocheck` atop those files and removing other `// @ts-xxx` directives (so they won't interfere with `@ts-nocheck`). The default setting allows these directives to be stripped from all JavaScript and friend files in `lib`, `src` and `test` directories. You can specify a different glob expression or set it to `false` to completely disable this behavior.
 
+### `dryRun` [`boolean`]
+
+Default: `false`<br />
+Command line: `--dryRun`<br />
+Config file: `"dryRun": 5`
```
nicojs, line 152:
```suggestion
Config file: `"dryRun": true`
```

florisg-infosupport: 🙈sloppy paste, will fix. Along with the change to dryRunOnly

packages/core/src/process/3-dry-run-executor.ts
```diff
@@ -110,6 +110,10 @@ export class DryRunExecutor {
   }
 
   private async executeDryRun(testRunner: TestRunner): Promise<DryRunCompletedEvent> {
+    if (this.options.dryRunOnly) {
+      this.log.info('Note: the dry-run has been started, no mutations will be run.');
```
nicojs, line 114:
```suggestion
      this.log.info('Note: running the dry-run only. No mutations will be tested.');
```

packages/core/src/stryker-cli.ts
```diff
@@ -92,6 +92,10 @@ export class StrykerCli {
         'Configure a build command to run after mutating the code, but before mutants are tested. This is generally used to transpile your code before testing.' +
           " Only configure this if your test runner doesn't take care of this already and you're not using just-in-time transpiler like `babel/register` or `ts-node`."
       )
+      .option(
+        '--dryRunOnly',
+        'Execute the initial test run only, without doing actual mutation testing. Doing a dry run can be used to test that StrykerJS can run your test setup, for example, in CI pipelines.'
```
nicojs, line 97:
```suggestion
        'Execute the initial test run only, without doing actual mutation testing. Doing a dry run only can be used to test that StrykerJS can run your test setup, for example, in CI pipelines.'
```

packages/core/test/unit/process/4-mutation-test-executor.spec.ts
```diff
@@ -383,4 +383,18 @@ describe(MutationTestExecutor.name, () => {
     // Assert
     expect(testInjector.logger.info).calledWithExactly('Done in %s.', '2 seconds, tops!');
   });
+
+  it('should short circuit when dryRunOnly is enabled', async () => {
+    // Arrange
+    testInjector.options.dryRunOnly = true;
```
nicojs, line 389: We probably also want to use `arrangeScenario()` here, to make sure it isn't short-circuited because there weren't any mutants.

docs/configuration.md
```diff
@@ -145,6 +145,14 @@ Config file: `"disableTypeChecks": false`
 
 Configure a pattern that matches the files of which type checking has to be disabled. This is needed because Stryker will create (typescript) type errors when inserting the mutants in your code. Stryker disables type checking by inserting `// @ts-nocheck` atop those files and removing other `// @ts-xxx` directives (so they won't interfere with `@ts-nocheck`). The default setting allows these directives to be stripped from all JavaScript and friend files in `lib`, `src` and `test` directories. You can specify a different glob expression or set it to `false` to completely disable this behavior.
 
+### `dryRunOnly` [`boolean`]
+
+Default: `false`<br />
+Command line: `--dryRunOnly`<br />
+Config file: `"dryRunOnly": false`
+
+Execute the initial test run only without doing actual mutation testing. Doing a dry run can be used to test that StrykerJS can run your test setup, for example, in CI pipelines.
```
nicojs, line 154:
```suggestion
Execute the initial test run only without doing actual mutation testing. Dry run only will still mutate your code before doing the dry run without those mutants being active, thus can be used to test that StrykerJS can run your test setup. This can be useful, for example, in CI pipelines.
```

packages/api/schema/stryker-core.json
```diff
@@ -290,6 +290,11 @@
       "$ref": "#/definitions/dashboardOptions",
       "default": {}
     },
+    "dryRunOnly": {
+      "description": "Runs only the initial test run, without running the mutators. This can be used to test for a correct setup of Stryker.",
```
nicojs, line 294:
```suggestion
      "description": "Execute the initial test run only without doing actual mutation testing. Dry run only will still mutate your code before doing the dry run without those mutants being active, thus can be used to test that StrykerJS can run your test setup. This can be useful, for example, in CI pipelines.",
```

nicojs (changes requested):

packages/core/test/unit/process/4-mutation-test-executor.spec.ts
florisg-infosupport: I think that does not really matter for this test, as we try to short circuit even before that check.
Or at least that is what is happening now. If we do want to include the earlyResults with the mutantPlan, that that needs to change

nicojs (approved):

nicojs:
Released in v6.3 🎉
https://github.com/stryker-mutator/stryker-js/releases/tag/v6.3.0

Issue #1212, opened 2018-10-26
Add documentation to generated stryker.conf.js (stryker init)


nicojs:
Right now, the `stryker init` command results in a stripped down config file, without any explanation. For example, if you choose "mocha" as your test runner, you'll get an example mocha configuration in your stryker.conf.js, but without any further info. When the `stryker run` command later fails because of miss configuration, you're effectively stuck. 

It would be awesome if the command adds the links to the plugins web page. So people know where to find more options.

* ☕ https://stryker-mutator.io/docs/stryker-js/mocha-runner/
* 🃏 https://stryker-mutator.io/docs/stryker-js/jest-runner/
* 🥒 https://stryker-mutator.io/docs/stryker-js/cucumber-runner/
* 🇰 https://stryker-mutator.io/docs/stryker-js/karma-runner/
* 🟣https://stryker-mutator.io/docs/stryker-js/jasmine-runner/

The `stryker init` commands underwater retrieves the package.json files of the installed plugins in order to get at the `initStrykerConfig` package.json value. We can reuse that to get at the `homepage` value. Example url for package.json is:
https://registry.npmjs.org/stryker-jasmine-runner/latest

kmdrGroch:
I think it can be closed now

nicojs:
Closed with #3817.

Thanks @Giovds ♥

PR #3817, opened 2022-10-28
Add documentation to generated stryker.conf.js

Giovds:
**Add feature #1212**
Initial setup providing fetching the URL from the response and displaying it in the comment property of the `stryker.conf.json`.

image


nicojs:
Released in v6.3 🎉
https://github.com/stryker-mutator/stryker-js/releases/tag/v6.3.0

Issue #1463, opened 2019-03-21
File permissions are not copied correctly to the sandbox


connorjclark:
Couldn't immediately find the name of the package for the babel transpiler from docs, so I googled it and got https://www.npmjs.com/package/stryker-babel-transpiler . But that doesn't work. Eventually found it is `@stryker-mutator/babel-transpiler`.

I guess all the packages moved to an org scope recently. Maybe deprecate the old ones + put a big disclaimer in the readme pointing to the correct package?

Hoten:
Actually, I am have more issues, the above package didn't work.

Trying to run against `terser`.

```
10:54:39 (66028) ERROR StrykerCli an error occurred { Error: Could not inject "InitialTestExecutor". Inner error: Could not inject "TranspilerFacade". Inner error: Cannot load Transpiler plugin "babel". In fact, no Transpiler plugins were loaded. Did you forget to install it?
    at ClassProvider.injectClass (/Users/cjamcl/src/terser/node_modules/typed-inject/src/InjectorImpl.js:35:19)
    at ClassProvider.injectClass (/Users/cjamcl/src/terser/node_modules/typed-inject/src/InjectorImpl.js:95:22)
    at Stryker.<anonymous> (/Users/cjamcl/src/terser/node_modules/@stryker-mutator/core/src/Stryker.js:74:30)
    at step (/Users/cjamcl/src/terser/node_modules/tslib/tslib.js:133:27)
    at Object.next (/Users/cjamcl/src/terser/node_modules/tslib/tslib.js:114:57)
    at fulfilled (/Users/cjamcl/src/terser/node_modules/tslib/tslib.js:104:62)
    at <anonymous>
```

```
module.exports = function(config) {
  config.set({
    mutator: "javascript",
    packageManager: "npm",
    reporters: ["html", "clear-text", "progress"],
    testRunner: "command",
    transpilers: ["babel"],
    coverageAnalysis: "all"
  });
};
```

```
diff --git a/package.json b/package.json
index d05e452e..92f99546 100644
--- a/package.json
+++ b/package.json
@@ -35,6 +35,10 @@
     "source-map-support": "~0.5.9"
   },
   "devDependencies": {
+    "@stryker-mutator/babel-transpiler": "^1.1.1",
+    "@stryker-mutator/core": "^1.1.1",
+    "@stryker-mutator/html-reporter": "^1.1.1",
+    "@stryker-mutator/javascript-mutator": "^1.1.1",
[...]

PR #3824, opened 2022-10-30
Bug/1463 filesystem permissions are not copied in certain cases

florisg-infosupport:
This change is to more explicitly set the file rights from the original file. In the case of fs.copyFile this is in theory not needed, but for fs.writeFile it is. Otherwise the default right would be 644, which is not always sufficient.

This change also introduces integration tests to cover these scenarios. For windows they are not run, as it is does not fully support changing file rights via chmod.

Fixes: #1463


florisg-infosupport:
if for the tests, the 777 is too scary, then this could also be changed

packages/core/src/fs/project-file.ts
```diff
@@ -77,13 +83,26 @@ export class ProjectFile implements FileDescription {
...
     }
+    await this.setChmod(backupFileName);
     return backupFileName;
   }
 
   public get hasChanges(): boolean {
     return this.#currentContent !== this.#originalContent;
   }
+
+  private async getChmod(file: string): Promise<number | undefined> {
```
nicojs, line 98: This class uses ES privates, not typescript `private`. So instead of `private getChmod` we should use `#getChmod`. ES Privates are a recent addition to JavaScript, so they're natively supported.

packages/core/test/integration/fs/cleanup-file-system.ts
```diff
@@ -0,0 +1,55 @@
+import fs from 'fs';
```
nicojs, line 1: Why is this class needed? Removing a directory and it's children can simply be done using:

```ts
import fs from 'fs';

await fs.promises.rm('path/to/dir', { recursive: true });
```

See https://nodejs.org/api/fs.html#fspromisesrmpath-options

The reason to keep track of the concurrent IO tasks in the production code is that Stryker does so many parallel IO actions you can run out of OS-level file handles. We don't worry about that during integration testing.

packages/core/test/integration/fs/project-file.it.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+import { testInjector } from '@stryker-mutator/test-helpers';
+import { expect } from 'chai';
+
+import { FileSystem, ProjectFile } from '../../../src/fs/index.js';
+import { resolveFromRoot } from '../../helpers/test-utils.js';
+
+import { CleanupFileSystem } from './cleanup-file-system.js';
+
+const resolveTestResource = resolveFromRoot.bind(undefined, 'testResources', 'project-file');
+const TARGET_DIRECTORY = 'temp';
```
nicojs, line 12: could we use the OS temp directory here?

```ts
import fs from 'fs';

const targetDirectory = await fs.promises.mkdtemp('stryker-project-file-test');
```

https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options

packages/core/test/integration/fs/project-file.it.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  const originalFile = resolveTestResource('full-rights-file.txt');
+  await fileSystem.chmod(originalFile, '0777');
+
+  const originFileStats = await fileSystem.stat(originalFile);
+  expect(statToChmod(originFileStats)).eq('0777');
+  return originalFile;
+}
+
+function statToChmod(stats: Stats): string {
+  const unixFilePermissions = '0' + (stats.mode & parseInt('777', 8)).toString(8);
```
nicojs, line 129: I'll trust this is correct 😅

packages/core/test/unit/fs/project-file.spec.ts
```diff
@@ -240,6 +242,99 @@ describe(ProjectFile.name, () => {
     });
   });
 
+  describe('chmod', () => {
```
nicojs, line 245: I see you've separated the mode tests into a `describe` here. This is not what I would prefer. I would assume there to be a `describe` per public method. 

In this case, I would expect 1~ish additional test per public method to test the `chmod` stuff specifically. 

What do you think?

nicojs (changes requested):
Thanks a lot for picking this up. Looks good so far, some small remarks.

nicojs:
Confused emoji? Isn't this the test you've added in a separate PR?

florisg-infosupport:
I think you are confused with a different pr, the test that fails is indeed from this pr. But I just created a new pr for a different problem

PR #3832, opened 2022-10-31
docs(parallel-workers): new documentation page for `parallel workers`

### LINKED ISSUE #3828: Document the way StrykerJS uses parallel workers
I was looking for a page to document the new `STRYKER_MUTATOR_WORKER` environment variable, added in 6.3.0 with #3821, but I couldn't find one. It might make sense to add a page for it, probably also where we explain how Stryker creates workers depending on the `concurrency` configuration option. There we can also explain the `STRYKER_MUTATOR_WORKER` environment variable.

We can draw inspiration from node-tap: https://node-tap.org/docs/api/parallel-tests/#parallel-tests

ariba-05:
Closes #3828


docs/parallel-workers.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Parallel Workers
+custom_edit_url: https://github.com/stryker-mutator/stryker-js/edit/master/docs/parallel-workers.md
+---
+
+Stryker will always run checkers and test runners in parallel by creating worker processes (note, not `worker_threads`). The number of such processes forked is determined by the configuration option [`--concurrency`](./configuration.md#concurrency-number). 
+
+However, imagine running these parallel processes on a test suite which uses resources like a database connection, web server or file system. This means these processes can conflict if they write to the same database, file or utilize the same port. To solve this problem, Stryker sets a unique number (starting from 0 and incremented by 1) to each worker's environment variable `STRYKER_MUTATOR_WORKER`. This variable can be utilized to distribute the resources without conflicts - for instance, by setting the server port to `4444 + (+process.env.STRYKER_MUTATOR_WORKER)` instead of `4444`.
```
nicojs, line 8: I think it would be awesome to document a small but complete example here using `jest`. Including:

- A tiny app that creates 2 simple endpoints. (i.e. one that echoes whatever is sent and a health check that responds with 'OK'). 
- A test file with `jest` that starts a server in `beforeAll`, closes it in `afterAll`, and has 3 `it`s. 2 its for the echo and 1 for the health check endpoint, without using the environment variable first. Maybe make sure that 1 mutant should survive somehow.

Run this example with concurrent runners in StrykerJS and show precisely the problem with a screenshot of the HTML report. Namely, the mutation score is 100%, and the error message complains about a port being unavailable.

Then show the changes to make with a git diff sample:

```diff
let app;
beforeAll(() => {
-  app = new MyApp(4444);
+  app = new MyApp(4444 + (+process.env.STRYKER_MUTATOR_WORKER || 0);
});
```

And then show what the report looks like after.

Do you think you're up for this challenge? If your example works, we can also make it an actual `e2e` test that runs inside our CI pipeline.

nicojs (changes requested):
Wow, you're working fast! 🏃‍♀️

Thanks for picking this up! By default, markdown files won't appear on the stryker-mutator.io website unless referenced in the [sidebar.json file](https://github.com/stryker-mutator/stryker-js/blob/master/docs/sidebar.json). Please add this one there, just after 'incremental'. You can also run the website locally to see what it looks like by following the instructions here: https://github.com/stryker-mutator/stryker-mutator.github.io/blob/develop/CONTRIBUTING.md#run-locally

I would love a more detailed example; see my remark. If you don't have time, don't worry; I can still merge this and document it afterward. Your choice.

docs/parallel-workers.md
ariba-05: Hm, you're right - this would be a nice addition. I think I can attempt to add this.

nicojs (approved):
Wow that looks great!

ariba-05:
> You can also run the website locally to see what it looks like by following the instructions here: https://github.com/stryker-mutator/stryker-mutator.github.io/blob/develop/CONTRIBUTING.md#run-locally

This doesn't mean opening a new PR in the `stryker-mutator.github.io` repo, does it? I ran the website locally by just copying over the files from `stryker-js/docs` and ran `npm i && npm start` - which looks fine! So, the docusaurus workflow should automatically do that?  I don't quite understand it. :/

Also, we might want to update the screenshots which expose my system's file paths... `/Users/aribasiddiqui/...blabla`. What is the best way in your opinion? Blur it out? 

 > just in time for Hacktoberfest

haha, not really aiming for it anymore. 😄 This will only be my 2/4 PRs.

nicojs:
It's live! https://stryker-mutator.io/docs/stryker-js/parallel-workers/
😎

> This doesn't mean opening a new PR in the `stryker-mutator.github.io` repo, does it?

Indeed, it doesn't. Just to see what it will look like.

PR #3842, opened 2022-11-05
Warn when Mutator in directive does not exist

### LINKED ISSUE #3812: Stryker Equality ignore  once not working
image
image

Stryker Version: 
```
"@stryker-mutator/core": "^6.2.3",
"@stryker-mutator/karma-runner": "^6.2.3",
"@stryker-mutator/typescript-checker": "^6.1.2",
```

Giovds:
Closes #3812 

When a directive references an unknown Mutator, log a warning that this directive is not used.

**Mutated code**
```js
const compare = (a, b) => {
  // Stryker disable HelloThere
  // Stryker disable next-line Equality: Wrong Mutator name
  return a - b >= 0 ? 1 : -1;
}

module.exports = { compare };

module.exports = { compare };
```

**Cli**
image

### DIFF
--- a/packages/instrumenter/src/transformers/directive-bookkeeper.ts
+++ b/packages/instrumenter/src/transformers/directive-bookkeeper.ts
@@ -59,16 +59,17 @@ export class DirectiveBookkeeper {
 
   public processStrykerDirectives({ loc, leadingComments }: types.Node): void {
     leadingComments
-      ?.map(
-        (comment) =>
-          this.strykerCommentDirectiveRegex.exec(comment.value) as
-            | [fullMatch: string, directiveType: string, scope: string | undefined, mutators: string, reason: string | undefined]
-            | null
-      )
-      .filter(notEmpty)
-      .forEach(([, directiveType, scope, mutators, optionalReason]) => {
+      ?.map((comment) => ({
+        comment,
+        matchResult: this.strykerCommentDirectiveRegex.exec(comment.value) as
+          | [fullMatch: string, directiveType: string, scope: string | undefined, mutators: string, reason: string | undefined]
+          | null,
+      }))
+      .filter((commentAndRegexMatches) => notEmpty(commentAndRegexMatches.matchResult))
+      .forEach((commentAndRegexMatches) => {
+        const [, directiveType, scope, mutators, optionalReason] = commentAndRegexMatches.matchResult!;
         let mutatorNames = mutators.split(',').map((mutator) => mutator.trim());
-        this.warnAboutUnusedDirective(mutatorNames, directiveType, scope!, leadingComments);
+        this.warnAboutUnusedDirective(mutatorNames, directiveType, scope, commentAndRegexMatches.comment);
         mutatorNames = mutatorNames.map((mutator) => mutator.toLowerCase());
         const reason = (optionalReason ?? DEFAULT_REASON).trim();
         switch (directiveType) {
@@ -101,17 +102,17 @@ export class DirectiveBookkeeper {
     return this.currentIgnoreRule.findIgnoreReason(mutatorName, line);
   }
 
-  private warnAboutUnusedDirective(mutators: string[], directiveType: string, scope: string, leadingComments: types.Comment[]) {
-    for (const directive of mutators) {
-      if (directive === WILDCARD) continue;
-      if (!this.allMutatorNames.includes(directive.toLowerCase())) {
-        const directiveComment = leadingComments.find((comment) => comment.value.includes(directive));
+  private warnAboutUnusedDirective(mutators: string[], directiveType: string, scope: string | undefined, comment: types.Comment) {
+    for (const mutator of mutators) {
+      if (mutator === WILDCARD) continue;
+      if (!comment.value.includes(mutator)) return;
+      if (!this.allMutatorNames.includes(mutator.toLowerCase())) {
         this.logger.warn(
           // Scope can be global and therefore undefined
           `Unused 'Stryker ${
             scope ? directiveType + ' ' + scope : directiveType
-          }' directive. Mutator with name '${directive}' not found. Directive found at: ${this.originFileName}:${directiveComment!.loc!.start.line}:${
-            directiveComment!.loc!.start.column
+          }' directive. Mutator with name '${mutator}' not found. Directive found at: ${this.originFileName}:${comment.loc!.start.line}:${
+            comment.loc!.start.column
           }.`
         );
       }
--- a/packages/instrumenter/src/transformers/directive-bookkeeper.ts
+++ b/packages/instrumenter/src/transformers/directive-bookkeeper.ts
@@ -76,7 +76,7 @@ export class DirectiveBookkeeper {
               ignoreReason: `Unused 'Stryker ${directiveType}' directive`,
               replacement: node,
             });
-            collector.collect(originFileName, node, mutant);
+            collector.collect(originFileName, node, mutant, { line: -1, position: 0 }); // Assumption that the directive is always -1 above it...
           }
         }
 
--- a/packages/instrumenter/src/transformers/directive-bookkeeper.ts
+++ b/packages/instrumenter/src/transformers/directive-bookkeeper.ts
@@ -1,5 +1,10 @@
 import type { types } from '@babel/core';
-import { notEmpty } from '@stryker-mutator/util';
+import { notEmpty, I } from '@stryker-mutator/util';
+
+import { Mutant } from '../mutant.js';
+import { NodeMutator } from '../mutators/node-mutator.js';
+
+import { MutantCollector } from './mutant-collector.js';
 
 const WILDCARD = 'all';
 const DEFAULT_REASON = 'Ignored using a comment';
@@ -48,10 +53,8 @@ export class DirectiveBookkeeper {
 
   private currentIgnoreRule = rootRule;
 
-  public readonly uniqueDirectives = new Set<string>();
-
-  public processStrykerDirectives({ loc, leadingComments }: types.Node): void {
-    leadingComments
+  public processStrykerDirectives(node: types.Node, allMutators: NodeMutator[], collector: I<MutantCollector>, originFileName: string): void {
+    node.leadingComments
       ?.map(
         (comment) =>
           this.strykerCommentDirectiveRegex.exec(comment.value) as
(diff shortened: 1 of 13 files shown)


Giovds:
A couple of questions/issues I still have:
1. I've made the assumption here: https://github.com/stryker-mutator/stryker-js/pull/3842/files#diff-87b4b61b8bd5767c34cf6b141cc3721b722a730e884627e7adc127c8449a07b4R79. Is this correct?
2. It probably is confusing that ignored mutant in the HTML view shows a diff on the comment but has no 'addition line'.
3. It seems that the line number is off by 1. There is a `toBableLine()` method which adds 1 to the line number. Should this be used to re-convert the lines to their original numbers?

nicojs (changes requested):
It is pretty clever to see how you report mutants for unused stryker directives, but I don't think it's a good idea :/. Mutants are mutations in the code, not unused directives. 

I think logging them as a warning is good enough for now. For example, users might want Stryker to fail fast instead of showing them in the report 🤷‍♀️.

packages/instrumenter/src/transformers/directive-bookkeeper.ts
```diff
@@ -59,12 +64,28 @@ export class DirectiveBookkeeper {
...
+          .map((mutator) => mutator.trim())
+          .filter((mutator) => mutator !== WILDCARD);
+        for (const directive of directives) {
+          if (!allMutators.map((x) => x.name.toLowerCase()).includes(directive.toLowerCase())) {
+            const mutant = new Mutant(directive, originFileName, node, {
+              mutatorName: directive,
+              ignoreReason: `Unused 'Stryker ${directiveType}' directive`,
+              replacement: node,
+            });
+            collector.collect(originFileName, node, mutant, { line: -1, position: 0 }); // Assumption that the directive is always -1 above it...
```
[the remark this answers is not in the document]
Giovds: Fair point, as I previously mentioned I was misusing them to get them in the report, so that they could get highlighted 🙈 . 

Would you rather inject a logger here, or a function and call it through all the layers? Option 1 probably will log `WARN DirectiveBookkeeper` rather than `WARN Instrumenter`?

packages/instrumenter/src/transformers/directive-bookkeeper.ts
```diff
@@ -48,8 +53,8 @@ export class DirectiveBookkeeper {
 
   private currentIgnoreRule = rootRule;
 
-  public processStrykerDirectives({ loc, leadingComments }: types.Node): void {
-    leadingComments
+  public processStrykerDirectives(node: types.Node, allMutators: NodeMutator[], collector: I<MutantCollector>, originFileName: string): void {
```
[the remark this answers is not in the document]
Giovds: 👍

packages/instrumenter/src/transformers/directive-bookkeeper.ts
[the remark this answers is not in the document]
nicojs: We don't _really_ use the logger categories ATM. I'd be happy with it just being `WARN Instrumenter`. That seems the easiest and straight forward way for now.

packages/instrumenter/src/transformers/directive-bookkeeper.ts
```diff
@@ -58,6 +64,7 @@ export class DirectiveBookkeeper {
       )
       .filter(notEmpty)
       .forEach(([, directiveType, scope, mutators, optionalReason]) => {
+        this.warnAboutUnusedDirective(mutators, directiveType, scope!);
         const mutatorNames = mutators.split(',').map((mutator) => mutator.trim().toLowerCase());
```
nicojs, line 68: Please pass this `mutatorNames` to `warnAboutUnusedDirecttive` instead of `mutators`. That way we only split on `,` once.

packages/instrumenter/src/transformers/directive-bookkeeper.ts
```diff
@@ -89,4 +96,21 @@ export class DirectiveBookkeeper {
...
     return this.currentIgnoreRule.findIgnoreReason(mutatorName, line);
   }
+
+  private warnAboutUnusedDirective(mutators: string, directiveType: string, scope: string) {
+    const directives = mutators
+      .split(',')
+      .map((mutator) => mutator.trim())
+      .filter((mutator) => mutator !== WILDCARD);
+    for (const directive of directives) {
+      if (!this.allMutators.map((x) => x.name.toLowerCase()).includes(directive.toLowerCase())) {
```
nicojs, line 106: Can we do this `map` in the constructor of this class? Now we're doing it every time we find a directive. 

Its a small performance hit, but mostly this `warn` function shouldn't be doing too much logic.

packages/instrumenter/src/transformers/directive-bookkeeper.ts
```diff
@@ -89,4 +96,21 @@ export class DirectiveBookkeeper {
...
+        this.logger.warn(
+          // Scope can be global and therefore undefined
+          `Unused 'Stryker ${
+            scope ? directiveType + ' ' + scope : directiveType
+          }' directive. Mutator with name '${directive}' not found. Directive found at: ${this.originFileName}.`
```
nicojs, line 111: We can use `comment.loc!.start.line` and `comment.loc!.start.column` to also print the location. That would be awesome!

packages/instrumenter/test/unit/transformers/babel-transformer.spec.ts
```diff
@@ -60,7 +65,8 @@ describe('babel-transformer', () => {
       path.replaceWith(types.sequenceExpression([...[...appliedMutants.values()].flatMap((val) => val.expressions), ...path.node.expressions])),
   };
 
-  beforeEach(() => {
+  beforeEach(function () {
+    if (this.currentTest?.title === 'should warn users when a mutator name does not match any of the enabled mutators.') return; // override beforeEach for issue 3812
```
nicojs, line 69: Please don't ever do this 🙈.

Also: I don't understand why this exclusion for the test is needed. I think the test can work with this before each. There are 2 mutators active here, right?

packages/instrumenter/test/unit/transformers/babel-transformer.spec.ts
```diff
@@ -443,6 +449,35 @@ describe('babel-transformer', () => {
         expect(notIgnoredMutants()).lengthOf(1);
       });
 
+      // issue https://github.com/stryker-mutator/stryker-js/issues/3812
+      it('should warn users when a mutator name does not match any of the enabled mutators.', () => {
+        // Explicitly override the before each because we are interested in the default Mutators
```
nicojs, line 454: Please add a test per scenario.

packages/instrumenter/test/unit/transformers/babel-transformer.spec.ts
```diff
@@ -443,6 +449,35 @@ describe('babel-transformer', () => {
...
+                return a - b >= 0 ? 1 : -1;
+              }
+            `,
+        });
+        act(ast);
+
+        expect(logger.warn).calledWithMatch(
+          sinon.match("Unused 'Stryker disable' directive. Mutator with name 'SomeOtherThing' not found. Directive found at: example.ts")
+        );
+        expect(logger.warn).calledWithMatch(
```
nicojs, line 476: Don't forget to add a test where you `expect(logger.warn).not.called`.

nicojs (changes requested):
Looking good! 🤩 , still some stuff left.

packages/instrumenter/test/unit/transformers/babel-transformer.spec.ts
Giovds: I think I was testing stuff for my original idea were I was exploring some other stuff as well. But you are right I should've removed it afterwards 🙃. I can probably just do with the given scenario.

packages/instrumenter/src/transformers/directive-bookkeeper.ts
Giovds: I probably should have placed a comment for this, but I'll quote the PR description

> I've not included the line number and column since these did not correctly seem to tell where the directive was found, and in case of a next-line directive it had to subtract 1. I think the filename should be sufficient though to find the unused directive. WDYT?

Giovds: I'll retest and implement the feedback somewhere tomorrow or so.  Thanks!

nicojs: Seems to work fine :) 

image

Giovds: Ah there are comment nodes as well, I was looking at the nodes where the mutation is applied. Thanks!

packages/instrumenter/src/transformers/directive-bookkeeper.ts
Giovds: Is it OK to show lowercase mutator names, in that case I don't need to split op the assignment to the `mutatorNames`?

packages/instrumenter/src/transformers/directive-bookkeeper.ts
```diff
@@ -89,4 +100,21 @@ export class DirectiveBookkeeper {
     mutatorName = mutatorName.toLowerCase();
     return this.currentIgnoreRule.findIgnoreReason(mutatorName, line);
   }
+
+  private warnAboutUnusedDirective(mutators: string[], directiveType: string, scope: string, leadingComments: types.Comment[]) {
+    for (const directive of mutators) {
```
nicojs, line 105: This should be `mutator` instead of `directive`, right?

packages/instrumenter/src/transformers/directive-bookkeeper.ts
```diff
@@ -89,4 +100,21 @@ export class DirectiveBookkeeper {
...
+      if (directive === WILDCARD) continue;
+      if (!this.allMutatorNames.includes(directive.toLowerCase())) {
+        const directiveComment = leadingComments.find((comment) => comment.value.includes(directive));
```
nicojs, line 108: I think it is better to flow the comment through from the calling side instead of trying to search for it again here.

nicojs (changes requested):
This looks awesome; just a couple of remarks left. Let me know if you need more help.

nicojs (approved):

nicojs:
Thanks! 🎉

PR #3877, opened 2022-11-26
docs(config): clarify --mutate, --ignorePatterns options

bodote:
With this pull request I try to clear my great confusion reading the docs about the usage of the deprecated `--file` option  and the `--ignorePattern` and `--mutate` options , because I suspect other new users of stryker-js are and will be confused, too, if we don't change the docs.

(  See also the interesting conversation in #3815  but
somehow the hole fork of mine got mixed up with tons of files I really didn't want to change at all, I also couldn't revert those unwanted changes, so I closed that old pull request, and try again with this new one.)


bodote:
Now I understand what went wrong with #3815 : I did my changes on the master branch of my fork , instead of creating a branch on my fork and commit my changes to my branch. Unfortunately I did the same mistake again with this pull request  
Do I need to close this pull request as well and create a new pull request again the right way ? Sorry for the confusion.

PR #3900, opened 2022-12-09
Added Typescript group checking 🚀

danny12321:
Add grouping mutants to the `@stryker-mutator/typescript-checker` to improve performance.

The typescript checker is slow because mutants are checked for compile errors one by one. Each time a mutant is checked, TypeScript will do full type-checking. This is sped up by running multiple checkers in parallel, using typescript's `--watch` mode, and only compiling in memory. Yet still, this is very slow, as a rule of thumb: 10x slower than not enabling this checker.

This PR improves this by grouping mutants unrelated to each other in the same typescript compilation run. Any compilation errors that result from one of the mutants in a group can always be associated with one mutant. 

One downside to this approach is that a TypeScript compilation sometimes forgets to report some of the errors (although it is a design goal of TypeScript to report them all). For example https://github.com/microsoft/TypeScript/issues/46272

This is why you disable this grouping behavior by adding:

```json
{
  "typeScriptChecker": {
    "prioritizePerformanceOverAccuracy": false
  }
}
```

This is just a draft PR. Todo:
- [x] fix tests
- [x] add options to choose testing strategy (https://github.com/stryker-mutator/stryker-js/issues/3899)


packages/typescript-checker/src/grouping/create-groups.ts
```diff
@@ -0,0 +1,116 @@
...
+ * In this example we can type check B and D at the same time.
+ * This is because these files can't throw errors in each other.
+ * If we type check them and let's say B throws an error.
+ * We know for sure that the mutant in B was the one creating the type error.
+ * If we type check B and D at the same time it is possible that an error shows up in A.
+ * When this happens we go down de dependency graph and individual test the mutants who were in that group.
+ *
+ * In this function we create the groups of mutants who can be tested at the same time.
+ */
+export function createGroups(mutants: Mutant[], nodes: Map<string, Node>): string[][] {
```
nicojs, line 41: I've proposed a refactoring in a separate PR

packages/typescript-checker/src/typescript-compiler.ts
```diff
@@ -0,0 +1,251 @@
...
+    if (!sourceMappingURL) return [fileName];
+
+    const sourceMapFileName = path.resolve(fileName, '..', sourceMappingURL);
+    const sourceMap = this.fs.getFile(sourceMapFileName);
+    if (!sourceMap) throw new Error(`Could not find ${sourceMapFileName}`);
+
+    const content = JSON.parse(sourceMap.content);
+
+    return content.sources.map((sourcePath: string) => toPosixFileName(path.resolve(sourceMapFileName, '..', sourcePath)));
+  }
```
nicojs, line 220: What is this method?

packages/typescript-checker/src/grouping/node.ts
```diff
@@ -0,0 +1,38 @@
+import { Mutant } from '@stryker-mutator/api/src/core';
+
+export class Node {
```
nicojs, line 3: I'm thinking this might not need to be a class. Maybe it can be a simple interface:

```ts
interface Node {
  parents: Set<Node>;
  children: Set<Node>;
}
```

What do you think?

nicojs (changes requested):
I've taken an hour to take a look. Got some remarks.

packages/typescript-checker/src/grouping/create-groups.ts
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * In this example we can type check B and D at the same time.
+ * This is because these files can't throw errors in each other.
+ * If we type check them and let's say B throws an error.
+ * We know for sure that the mutant in B was the one creating the type error.
+ * If we type check B and D at the same time it is possible that an error shows up in A.
+ * When this happens we go down de dependency graph and individual test the mutants who were in that group.
+ *
+ * In this function we create the groups of mutants who can be tested at the same time.
+ */
```
nicojs, line 40:
```suggestion
/**
 * To speed up the type-checking we want to check multiple mutants at once.
 * When multiple mutants in different files don't have overlap in affected files (or have small overlap), we can type-check them simultaneously.
 * These mutants who can be tested at the same time are called a group.
 * Therefore, the return type is an array of arrays, in other words: an array of groups.
 *
 * @param mutants All the mutants of the test project.
 * @param nodes A graph representation of the test project.
 *
 * @example
 * Let's assume we got the following dependencies in files of a project, and in every file is one mutant.
 *
 *          ========
 *          = A.ts =
 *          ========
 *         /        \
 * ========          ========
 * = B.ts =          = C.ts =
 * ========          ========
 *                           \
 *                            ========
 *                            = D.ts =
 *                            ========
 *
 * A imports B and C
 * C imports D
 *
 * In this example, we can type-check B and D simultaneously.
 * This is because these files can't throw errors in each other.
 * If we type check them, let's say B reports an error.
 * We know that the mutant in B created the type error.
 * If we type check B and D at the same time, it is possible that an error shows up in A.
 * When this happens, we go down the dependency graph and individually test the mutants in that group.
 * 
 * In this function, we create groups of mutants who can be tested at the same time.
 */
```

This is a great explanation, I used Grammarly to pimp it a bit.

packages/typescript-checker/src/typescript-compiler.ts
danny12321: This method resolves the original filename(s) based on a .d.ts file.

packages/typescript-checker/src/grouping/node.ts
odinvanderlinden: In @danny12321 and my opinion this should be a class. The node knows who its parents and children are so i think its easier to do this object oriented instead of a functional way of programming. For instance if you would do it functional the function getAllChildReferencesIncludingSelf would look like
```ts
getAllChildReferencesIncludingSelf(allChildReferences: Set<Node> = new Set<Node>(), nodeToSearchIn: Node){
    allChildReferences.add(nodeToSearchIn);
    nodeToSearchIn.children.forEach((child) => {
      if (!allChildReferences.has(child)) {
        child.getAllChildReferencesIncludingSelf(allChildReferences);
      }
    });
    return allChildReferences;
  }
```
I think using the this keyword is more elegant than nodeToSearchIn

packages/typescript-checker/src/typescript-checker.ts
```diff
@@ -33,176 +29,149 @@ export function create(injector: Injector<PluginContext>): TypescriptChecker {
...
+    const mutantsThatCouldNotBeTestedInGroups = new Set<Mutant>();
+
+    //If there is only a single mutant the error has to originate from the single mutant
+    if (errors.length && mutants.length === 1) {
+      errorsMap[mutants[0].id] = errors;
+      return errorsMap;
+    }
+
+    for (const error of errors) {
+      const nodeErrorWasThrownIn = nodes.get(error.file?.fileName ?? '');
```
nicojs, line 126: When is `file` or `file.fileName` `undefined`, and why is `''` a valid alternative?

packages/typescript-checker/src/typescript-checker.ts
```diff
@@ -33,176 +29,149 @@ export function create(injector: Injector<PluginContext>): TypescriptChecker {
...
+      if (!nodeErrorWasThrownIn) {
+        throw new Error('Error not found in any node');
```
nicojs, line 128: I would like some more details in this error message. Which file name couldn't be found? How can a user solve this error? Or should they open an issue?

packages/typescript-checker/src/typescript-checker.ts
```diff
@@ -33,176 +29,149 @@ export function create(injector: Injector<PluginContext>): TypescriptChecker {
...
+          errorsMap[mutantsRelatedToError[0].id] = [error];
+        }
+      } else if (mutantsRelatedToError.length === 0) {
+        for (const mutant of mutants) {
+          mutantsThatCouldNotBeTestedInGroups.add(mutant);
+        }
+      } else {
+        for (const mutant of mutantsRelatedToError) {
+          mutantsThatCouldNotBeTestedInGroups.add(mutant);
+        }
```
nicojs, line 145: Could you add a rule of comment in each of the 3 scenarios here? What does each scenario mean?

packages/typescript-checker/src/typescript-compiler.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    const compiler = ts.createSolutionBuilderWithWatch(host, [this.tsconfigFile], {});
+    compiler.build();
+    return await this.check([]);
+  }
+
+  public async check(mutants: Mutant[]): Promise<ts.Diagnostic[]> {
+    this.lastMutants.forEach((mutant) => {
+      const file = this.fs.getFile(mutant.fileName);
+      file?.resetMutant();
```
nicojs, line 149: Is it valid for files to not exist here?

packages/typescript-checker/src/typescript-compiler.ts
```diff
@@ -0,0 +1,243 @@
...
+    });
+    mutants.forEach((mutant) => {
+      const file = this.fs.getFile(mutant.fileName);
+      file?.mutate(mutant);
```
nicojs, line 153: Same here: is it valid for a `file` to not exist? What would happen? I think the `currentTask.promise` would simply never resolve. Better to crash in that case, feel free to use `!` instead of `?`.

packages/typescript-checker/src/typescript-compiler.ts
```diff
@@ -0,0 +1,243 @@
...
+    });
+    await this.currentTask.promise;
+    const errors = this.currentErrors;
+    this.currentTask = new Task();
+    this.currentErrors = [];
+    this.lastMutants = mutants;
+    return errors;
+  }
+
+  public getFileRelationsAsNodes(): Map<string, Node> {
```
nicojs, line 163:
```suggestion
  public get nodes(): Map<string, Node> {
```

I think this should be property.

packages/typescript-checker/src/typescript-compiler.ts
nicojs: I saw that :). Let me clarify:
1. Why do we need to resolve the original TS file name? 
2. Shouldn't we call it `resolveTSInputFile` instead?
3. Does this work when [inlineSourceMap](https://www.typescriptlang.org/tsconfig#inlineSourceMap), [inlineSources](https://www.typescriptlang.org/tsconfig#inlineSources), [mapRoute](https://www.typescriptlang.org/tsconfig#mapRoot) or [sourceRoute](https://www.typescriptlang.org/tsconfig#sourceRoot) are defined ?

packages/typescript-checker/src/grouping/node.ts
nicojs, line 3: Is `Node` a clear enough name? What about `TSFileNode` or maybe `TSFileGraph`? 

Can we add a description above it to explain why this class exists? I assume it is because TS only provides a `getAllDependencies(file)` API, and we need a 2-way dependency graph.

nicojs (changes requested):
Great work so far! It's looking better and better.

Did a second, more detailed review.

packages/typescript-checker/test/unit/grouping/create-groups.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+describe('create-group createGroups', () => {
+  it('single mutant should create single group', () => {
+    const mutants = [factory.mutant({ fileName: 'a.js', id: '1' })];
+    const mutantsClone = structuredClone(mutants);
+    const nodes = new Map<string, Node>([['a.js', new Node('a.js', [], [])]]);
+    const groups = createGroups(mutants, nodes);
+    expect(groups).to.have.lengthOf(1);
+    expect(groups[0][0]).to.be.equal(mutantsClone[0].id);
+  });
```
nicojs, line 17:
```suggestion
  it('single mutant should create single group', () => {
    const mutants = [factory.mutant({ fileName: 'a.js', id: 'mutant-1' })];
    const nodes = new Map<string, Node>([['a.js', new Node('a.js', [], [])]]);
    const groups = createGroups(mutants, nodes);
    expect(groups).to.have.lengthOf(1);
    expect(groups[0]).to.have.lengthOf(1);
    expect(groups[0][0]).to.be.equal('mutant-1');
  });
```

`structuredClone` [just landed in node 18](https://github.com/nodejs/node/pull/39759). Since we still support node 14, we cannot use it.

Since we're only using it to keep the order of mutant arrays and pick off the `id` fields which we've hard-coded in the first place, i think we should use those hard-coded ids directly.

packages/typescript-checker/src/typescript-compiler.ts
danny12321: The file should exist here. If there are files unrelated to the project we group them together in the first group. In the `check` method of the typescript-checker (not the compiler) we check if it is the group with unrelated mutants. If this is true we return this group with the status passed. This is done so the testrunners could start as fast as possible.

packages/typescript-checker/src/typescript-compiler.ts
danny12321: 1. We need the source filename to complete the dependency graph used by the grouping algorithm. If we don't use the source filename we can not identify the relation between projects/files.
2. I would only add the word `name` so it is clear that it is about file names. Do you agree with `resolveTSInputFileName`?
3. Nope. Should we try to implement it or could we just remove/change those options without sacrifices in the `tsconfig-helpers.ts` file?

packages/typescript-checker/src/typescript-checker.ts
```diff
@@ -33,176 +29,149 @@ export function create(injector: Injector<PluginContext>): TypescriptChecker {
...
+      getNewLine: () => EOL,
+    });
+  }
+
+  private loadSetup(options: StrykerOptions): TypeScriptCheckerOptions {
+    const defaultTypeScriptCheckerConfig: TypeScriptCheckerOptions = {
+      typeScriptChecker: { strategy: 'noGrouping' },
     };
-  };
+    return Object.assign(defaultTypeScriptCheckerConfig, options as TypeScriptCheckerOptionsWithStrykerOptions);
```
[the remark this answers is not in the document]
odinvanderlinden: For some reason i can't get it to work without manually assigning the defaults. How would this exactly work?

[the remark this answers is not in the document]
nicojs: It is done here:

https://github.com/stryker-mutator/stryker-js/blob/00706e0d6c6407ffc1747bc2807f9bccb307d633/packages/core/src/config/options-validator.ts#L183

Stryker uses [AJV](https://ajv.js.org/) as schema validator. AJV fills in defaults based on the defaults defined in the JSON schema. Stryker will merge your schema into the main schema when it finds an export from your plugin called `strykerValidationSchema`. I see that you have forgotten to do so. 

For example: https://github.com/danny12321/stryker-js-typescript-checker/blob/98d44ed91d11bf9628a78b36b5c1508a97b88b94/packages/jest-runner/src/index.ts#L13-L15

packages/typescript-checker/src/typescript-compiler.ts
nicojs: In that case, can we throw an error here instead of silently ignoring the mutants? Using an exclamation mark is fine: `!` (that will throw an error if the file is missing 😏)

packages/typescript-checker/src/grouping/ts-file-node.ts
```diff
@@ -0,0 +1,40 @@
...
+    allChildReferences.add(this);
+    this.children.forEach((child) => {
+      if (!allChildReferences.has(child)) {
+        child.getAllChildReferencesIncludingSelf(allChildReferences);
+      }
+    });
+    return allChildReferences;
+  }
+
+  // todo mogelijk mutants een map van maken [filename]: mutant[]
```
nicojs, line 28: TODO :)

danny12321:
> @danny12321 CI pipeline is failing. It seems that you're still using `structuredClone`. See my comment here: [#3900 (comment)](https://github.com/stryker-mutator/stryker-js/pull/3900#discussion_r1056939638)
> 
> Can we move this PR out of "draft"?

Oops, thought we had removed all of them. I have fixed it now.

nicojs (approved):
Great job! Can't wait to get this merged 😍

packages/typescript-checker/src/typescript-compiler.ts
```diff
@@ -0,0 +1,268 @@
...
+    if (!sourceMap) throw new Error(`Could not find ${sourceMapFileName}`);
+
+    const sources: string[] = JSON.parse(sourceMap.content).sources;
+
+    if (sources.length === 1) {
+      const sourcePath = sources[0];
+      return toPosixFileName(path.resolve(sourceMapFileName, '..', sourcePath));
+    }
+
+    return dependencyFileName;
```
nicojs, line 240: This part has no code coverage. We should add a TS solution with `.tsbuildinfo` files and project references to the `testResources` directory and use that to test some of these code paths.

nicojs (changes requested):
I've noticed that the coverage on some files is low. I've added unit tests to test the edge cases of the `typescript-checker.ts` file, but the `typescript-compiler.ts` tests are still missing some important code paths. See my remark there.

nicojs (approved):

odinvanderlinden:
Great work guys 🎉

nicojs:
:shipit:

PR #3929, opened 2023-01-11
Svelte support in StrykerJS

TeunLamers:
Working:
- Able to mutate Svelte files (including template expressions)
- Svelte integration tested with Jest test runner (e2e test)

Not done:
- Typescript support (Typescript Svelte files are unable to be parsed)
- Syntax highlighting in Svelte files HTML reporter


e2e/test/svelte-jest-project/jest.config.json
```diff
@@ -0,0 +1,14 @@
...
+    "transform": {
+        "^.+\\.js$": "babel-jest",
+        "^.+\\.svelte$": "svelte-jester",
+        ".+\\.(css|styl|less|sass|scss|png|jpg|ttf|woff|woff2)$": "jest-transform-stub"
+    },
+    "moduleFileExtensions": [
+        "svelte",
+        "js"
+    ],
+    "testEnvironment": "jsdom",    "setupFilesAfterEnv": [
```
nicojs, line 11: Nit: strange formatting

e2e/test/svelte-jest-project/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "@testing-library/svelte": "^3.2.2",
+    "@testing-library/user-event": "^14.4.3",
+    "babel-jest": "^29.2.2",
+    "babel-plugin-transform-vite-meta-env": "^1.0.3",
+    "jest": "^29.2.2",
+    "jest-environment-jsdom": "^29.2.2",
+    "jest-transform-stub": "^2.0.0",
+    "svelte": "^3.52.0",
+    "svelte-jester": "^2.3.2",
+    "vite": "^3.2.0"
```
nicojs, line 22: Can we add these dependencies to the `e2e/package.json` file instead? That makes for a faster CI pipeline.

e2e/test/svelte-jest-project/src/App.test.js
```diff
@@ -0,0 +1,35 @@
+// @ts-ignore
+const App = require('./App.svelte').default;
+// @ts-ignore
+const sum = require('./App.svelte').sum;
+// @ts-ignore
+const multiply = require('./App.svelte').sum;
+// @ts-ignore
+const isOldEnough = require('./App.svelte').isOldEnough;
+// @ts-ignore
+const loop = require('./App.svelte').loop;
```
nicojs, line 10: Why these `@ts-ignore` statements? Shouldn't this use `import` instead?

e2e/test/svelte-jest-project/stryker.conf.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "../../node_modules/@stryker-mutator/core/schema/stryker-schema.json",
+  "packageManager": "npm",
+  "testRunner": "jest",
+  "tempDirName": "stryker-tmp",
+  "concurrency": 1,
+  "logLevel": "debug",
+  "plugins": [
+    "../../../packages/jest-runner/dist/src/index.js"
```
nicojs, line 9: This should be removed. The `localDependencies` in the `e2e/package.json` file will install the `jest-runner`.

packages/instrumenter/package.json
```diff
@@ -41,13 +41,15 @@
     "@stryker-mutator/api": "6.3.1",
     "@stryker-mutator/util": "6.3.1",
     "angular-html-parser": "~3.0.0",
+    "svelte": "^3.52.0",
```
nicojs, line 44: Adding this here will increase the footprint of installing Stryker. Can we make it a `devDependency` instead? 

This would mean that end users should install `svelte` themselves if they want to mutate `.svelte` files, which is reasonable. However, these import statements will _fail_ as a result: `import ... from 'svelte'`. Instead of using import statements, we should use import _expressions_. I will place a remark at one of those files to make it clear.

packages/instrumenter/src/parsers/svelte-parser.ts
```diff
@@ -0,0 +1,100 @@
+import { BaseNode, Program } from 'estree';
```
nicojs, line 1: Nit: These are type imports only. Marking them with `import type` will make sure no value import remeans after TypeScript compilation.

```suggestion
import type { BaseNode, Program } from 'estree';
```

packages/instrumenter/src/parsers/svelte-parser.ts
```diff
@@ -0,0 +1,100 @@
...
+import { parse as svelteParse, walk } from 'svelte/compiler';
```
nicojs, line 2:
```suggestion
```

As mentioned before, this import statement should not be done here. Loading the file will fail when `svelte` is not installed.

packages/instrumenter/src/parsers/svelte-parser.ts
```diff
@@ -0,0 +1,100 @@
...
+import { Ast as InternalSvelteAst } from 'svelte/types/compiler/interfaces.js';
+
+import { AstFormat, SvelteAst, SvelteRootNode, SvelteNode } from '../syntax/index.js';
+
+import { ParserContext } from './parser-context.js';
+
+const header = '<script></script>\n\n';
+
+export async function parse(text: string, fileName: string, context: ParserContext): Promise<SvelteAst> {
+  let ast = svelteParse(text);
```
nicojs, line 12: This will make the import lazy; thus will only be done when an _actual_ svelte file exists.
```suggestion
  const { parse as svelteParse } = await import ('svelte/compiler');
  let ast = svelteParse(text);
```

packages/instrumenter/src/parsers/svelte-parser.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function getAdditionalScripts(svelteAst: InternalSvelteAst, text: string, fileName: string, context: ParserContext): Promise<SvelteNode[]> {
+  const additionalScriptsAsPromised: Array<Promise<SvelteNode>> = [];
+
+  if (svelteAst.instance?.content && svelteAst.module?.content) {
+    additionalScriptsAsPromised.push(sliceScript(text, fileName, svelteAst.module.content, context));
+  }
+
+  walk(svelteAst.html, {
```
nicojs, line 60:
```suggestion
  const { walk } = await import ('svelte/compiler');
  walk(svelteAst.html, {
```

packages/instrumenter/src/parsers/svelte-parser.ts
```diff
@@ -0,0 +1,100 @@
...
+
+function collectBindingExpression(node: BaseNode): BaseNode | null {
+  switch (node.type) {
+    case 'MustacheTag':
+    case 'IfBlock':
+    case 'ConstTag':
+    case 'EachBlock':
+    case 'AwaitBlock':
+    case 'KeyBlock':
+      return (node as any).expression;
```
nicojs, line 94: Question: shouldn't the switch on type `node.type` function as a type guard? Why is this `any` cast needed here?

packages/instrumenter/src/printers/svelte-printer.ts
```diff
@@ -0,0 +1,28 @@
+import { SvelteAst } from '../syntax';
```
nicojs, line 1: I love the simplicity here 😍

nicojs (changes requested):
Great job so far! I have some remarks, but you're definitely on the right track!

packages/instrumenter/src/mutant.ts
```diff
@@ -78,16 +78,17 @@ export class Mutant implements Mutable {
...
   };
+  return loc;
 }
 
-function toPosition(source: Position, lineOffset: number): Position {
+function toPosition(source: Position, offset: Offset): Position {
   return {
-    column: source.column,
-    line: source.line + lineOffset - 1, // Stryker works 0-based internally
+    column: source.column + (source.line === 1 ? offset.position : 0),
```
nicojs, line 91: We need to add unit tests to verify the column offset.

hugo-vrijswijk (approved):

TeunLamers:
> It would be nice to merge this and provide some example repo. @nosovk 
I've recently have come back from a break and will be working on this in the near term to get it merged. An example repo would be a good idea, thanks for the suggestion.

nicojs:
I've been working a bit on this. Almost finished. 

- [ ] Add svelte minimal version e2e
- [ ] Support [`<script lang="ts">`](https://svelte.dev/docs/typescript#script-lang-ts)
- [ ] Docs: add svelte guide

Thanks @TeunLamers, for updating it with master by the way, made the work a lot easier 😅

nicojs:
Still something wrong with the offset when reporting mutants:
image

nicojs (approved):
Let's document this in a separate PR

PR #4043, opened 2023-03-18
Feat/vitest

odinvanderlinden:
Todo:

- [x] implement test filter for mutantRun in [vitest-test-runner.ts](https://github.com/stryker-mutator/stryker-js/blob/feat/vitest/packages/vitest-runner/src/vitest-test-runner.ts) with regex filter
- [x] more unit tests if possible
- [x] more integration tests
- [x] add 1 e2e test
- [x] refactor [vitest-test-runner.ts](https://github.com/stryker-mutator/stryker-js/blob/feat/vitest/packages/vitest-runner/src/vitest-test-runner.ts)
- [x] Enable Stryker on integration tests
- [x] Unit test the `FileCommunicator`
- [x] Remove other changes.
- [x] Support [workspaces](https://vitest.dev/guide/#workspaces-support), see https://github.com/vitest-dev/vitest/issues/3365 and https://github.com/stryker-mutator/stryker-js/pull/4043#issuecomment-1547278332
- [ ] Clean state after each run, see https://github.com/vitest-dev/vitest/discussions/3017#discussioncomment-5901499

Optional:
- [x] Add option to change config file path


packages/vitest-runner/src/vitest-file-communication.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const setupFiles = {
+  coverageFile: resolveSetupFile('__stryker-coverage__.json'),
+  dryRun: resolveSetupFile('dry-run.js'),
+  globalNamespace: resolveSetupFile('global-namespace.js'),
+  hitCountFile: resolveSetupFile('hitCount.txt'),
+  activeMutant: resolveSetupFile('active-mutant.js'),
+  vitestSetup: resolveSetupFile('vitest-setup.js'),
+  hitLimit: resolveSetupFile('hit-limit.js'),
+};
```
nicojs, line 17: I realize that this won't work! The `activeMutant` setup file would be shared between all Stryker workers. We should use [`process.env.STRYKER_MUTATOR_WORKER` ](https://stryker-mutator.io/docs/stryker-js/parallel-workers/) to differentiate. Using this for the entire "file-communication protocol" is probably the safest.

Maybe we could change this file to be a class `"FileCommunicator"`. In the constructor, we can let the global environment file be injected. It has methods to read and write from the files. I think they should _not land inside the setup_, instead we should use a temp dir:

```ts
class FileCommunicator {
  private readonly communicationDir = path.resolve(os.tmpdir(), `vitest-runner-${process.env.STRYKER_MUTATOR_WORKER}`);
  constructor(private globalNamespace: string) {
  }

  public async init(){
    await fs.rm(this.communicationDir, { recursive: true, force: true });
  }
  
   //...
}
```

We can simplify further by writing only 1 setup file. This class would be responsible for it

```ts
class FileCommunicator {
 
  get setupFile() {
    return path.resolve(this.communicationDir, 'vitest.setup.js');
  }
  
  get coverageFile() {
     // ....
  }

  async setDryRun(options: DryRunOptions) {
    // ... write content to the setup file
    await fs.writeFile(this.setupFile, `
       // Write hit count, hit limit, isDryRun, global namespace, etc. Altogether in 1 file
    `);
  }
  
   async setMutantRun(options: MutantRunOptions) {
    // ... write content to the setup file
    await fs.writeFile(this.setupFile, `
       // Write active mutant, hit count, hit limit, isDryRun, global namespace, etc. Altogether in 1 file
    `);
  }
```

nicojs (changes requested):
We didn't think of a huge oversight, see the comment.

packages/vitest-runner/src/vitest-file-communication.ts
nicojs: Now that I think of it, using the temp dir probably won't work because you'll need to import from `vitest`.  To that end, only the Current Working Directory (cwd) is valid. We can use `path.resolve` for this purpose.

```ts
class FileCommunicator {
  private readonly communicationDir = path.resolve(`.vitest-runner-${process.env.STRYKER_MUTATOR_WORKER}`);
}
```

packages/vitest-runner/test/unit/vitest-file-communication.it.spec.ts
nicojs, line 1: What happened here? 😱

nicojs (changes requested):
Almost done :). I've requested some small changes.

packages/vitest-runner/schema/vitest-runner-options.json
```diff
@@ -0,0 +1,20 @@
...
+  "additionalProperties": false,
+  "properties": {
+    "vitest": {
+      "description": "Configuration for @stryker-mutator/vitest-runner",
+      "title": "VitestRunnerOptions",
+      "additionalProperties": false,
+      "type": "object",
+      "default": {},
+      "properties": {
+        "config": {
```
nicojs, line 13:
```suggestion
        "configFile": {
```

nicojs:
As of `vitest@0.30.0`, vitest supports workspaces. This has 2 consequences:

1. We cannot use the `vitest.config` anymore to set `setupFiles`. They now need to be set for each project in the workspace.
2. The `afterAll` hook we use to write stuff to the communication dir is now being called once for each workspace. We should I think collect data inside a dir, instead of a file.

nicojs:
I've discovered another issue with vitest. See https://github.com/vitest-dev/vitest/issues/3403. This is a show-stopper for us. 😢

nicojs (approved):

nicojs:
I think I managed to fix all issues.

This is a list of all the workarounds/changes I've applied since @odinvanderlinden's first draft

- Since v0.30.0 vitest supports workspaces. This had a breaking change in the programmable API. Instead of using `vitest.config` we now use `vitest.projects.forEach(project => project.config)`. See https://github.com/vitest-dev/vitest/issues/3365
- Our `vitest.setup.js` file is now inlined using `deps.inline`, because otherwise they are cached by nodejs. See https://github.com/vitest-dev/vitest/issues/3403
- When inlined, the `vitest.setup.js` file is cached between runs, because we're not using `--watch` mode. So we manually bust the cache for it using `vitest.projects[0].server.moduleGraph.invalidateModule`, see https://github.com/vitest-dev/vitest/issues/3409

@steschi, thanks for your reaction, but we should be able to get this working 😅

steschi:
@nicojs sorry, didn't wanna confuse anybody with my reaction but I can relate to some of the burden of making things work with vite from my workplace.

Generally really excited to see this landing eventually, finally getting to have proper mutation testing on a modern web application stack 😍. Did not think I'd be able to apply some of the principles thought at university within a very modern web ecosystem like vite and vitest.

Thanks for the work you and your team is doing on this and greetings from Austria! 👋🏻

PR #4061, opened 2023-03-25
:memo: Add precision about file conflicts within project

homersimpsons:
Fixes #4034

PR #4062, opened 2023-03-25
:memo: Add precision about file conflicts within project

homersimpsons:
(Revival of #4061, I messed up with GitHub codespaces)

Fixes https://github.com/stryker-mutator/stryker-js/issues/4034


nicojs:
Thanks! 

https://stryker-mutator.io/docs/stryker-js/parallel-workers/

PR #4075, opened 2023-03-31
Feat/tap runner

danny12321:
Added tap runner.


danny12321:
Waiting for #4105 to be merged because node version 14 is not supported.

packages/tap-runner/src/setup/hook.cts
```diff
@@ -0,0 +1,28 @@
...
+process.on('exit', finalCleanup);
+
+['SIGABRT', 'SIGINT', 'SIGHUP', 'SIGTERM'].forEach((signal) =>
+  process.on(signal, (_, signalNumber) => {
+    process.exit(128 + signalNumber);
+  })
+);
+
+function finalCleanup() {
+  fs.writeFileSync(`stryker-output-${process.pid}.json`, JSON.stringify(global[strykerGlobalNamespace]));
```
nicojs, line 27: Reporting the coverage is only needed in dry run. Maybe we can `delete global[strykerGlobalNamespace].mutationCoverage` when not in dry run?

packages/tap-runner/src/tap-test-runner.ts
```diff
@@ -0,0 +1,183 @@
...
+        if (hitLimitReached) {
+          reject(new HitLimitError(hitLimitReached.reason));
+          return;
+        }
+
+        resolve({ testResult: this.tapResultToTestResult(testFile, result, failedTests), coverage: file.mutantCoverage });
+      });
+
+      parser.on('bailout', () => {
+        if (os.platform() !== 'win32') {
```
nicojs, line 148: Maybe add a comment to explain why

packages/tap-runner/test/unit/tap-runner.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  let sut: TapTestRunner;
+
+  beforeEach(async () => {
+    sut = testInjector.injector.injectFunction(createTapTestRunnerFactory('__stryker2__'));
+  });
+
+  describe('capabilities', () => {
+    it('should communicate reloadEnvironment =false', () => {
+      const expectedCapabilities: TestRunnerCapabilities = {
+        reloadEnvironment: false,
```
nicojs, line 17: Should be true. The tap runner can reload the test environment because it runs every time in separate child processes.

nicojs (changes requested):

packages/tap-runner/package.json
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "devDependencies": {
+    "@stryker-mutator/test-helpers": "6.4.2",
+    "@types/glob": "8.0.0",
+    "@types/node": "18.15.11",
+    "tap": "^16.3.2"
+  },
+  "peerDependencies": {
+    "@stryker-mutator/core": "^6.4.0",
+    "mocha": ">= 7.2 < 11"
```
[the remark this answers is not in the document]
danny12321: Also the Stryker core dependency?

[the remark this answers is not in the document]
nicojs: That's fine. Should always be installed right next to stryker core. Copy a line from another plugin to be sure.

packages/tap-runner/schema/tap-runner-options.json
```diff
@@ -0,0 +1,22 @@
...
+      "description": "Configuration for @stryker-mutator/tap-runner",
+      "title": "TapRunnerOptions",
+      "additionalProperties": false,
+      "type": "object",
+      "default": {},
+      "properties": {
+        "testFiles": {
+          "description": "Specify glob expressions to test files to run. Defaults to {**/@(test|tests|__test__|__tests__)/**,**/*.@(test|tests|spec)}.@(cjs|mjs|js|jsx|ts|tsx)",
+          "type": "string",
+          "default": "{**/@(test|tests|__test__|__tests__)/**,**/*.@(test|tests|spec)}.@(cjs|mjs|js|jsx|ts|tsx)"
```
nicojs, line 16: Now that I think about it, it might be better to allow an array of strings and allow excluding using `!`. But fine to leave it like this for now.

packages/tap-runner/test/integration/tap-test-runner.it.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    sandbox = new TempTestDirectorySandbox('example');
+    await sandbox.init();
+    (testInjector.options as TapRunnerOptionsWithStrykerOptions).tap = {
+      testFiles: defaultTestFilesGlob,
+    };
+    sut = testInjector.injector.injectFunction(createTapTestRunnerFactory('__stryker2__'));
+    await sut.init();
+
+    const excludeFiles = ['tests/bail.spec.js', 'tests/error.spec.js'];
+    testFilter = (await findTestyLookingFiles(defaultTestFilesGlob)).filter((file) => !excludeFiles.includes(file));
```
nicojs, line 28: This is a common use case. I think we should allow users to configure exclusions.

nicojs (approved):
Looks good. Small remark. We can pick it up in a separate PR.

Issue #4122, opened 2023-04-18
Error in tap runner: `Error: EBUSY: resource busy or locked`


nicojs:
Our pipeline is flaky because of the recently added tap-runner package:
> Error: EBUSY: resource busy or locked, rmdir 'D:\a\stryker-js\stryker-js\packages\tap-runner\testResources\tmp\workDir-5612723'

This indicates that the actual child process that runs the tests isn't fully closed when the tap runner plugin resolves. This might lead to bad things, because the next mutant might be started, which might cause high CPU load --> process starvation --> false timeouts.. 

I've investigated this a bit, and I think the `complete` callback provided to `tap-runner` gets executed before the child process is completed:

https://github.com/tapjs/tap-parser/blob/a46c1a01e3bb308ff97a93f13b7cca3f782e0e14/src/index.ts#L477

See also for an example: https://github.com/stryker-mutator/stryker-js/actions/runs/4724456135/jobs/8381692413?pr=4121#step:5:3413

It might be best to convert both the [spawn](https://github.com/stryker-mutator/stryker-js/blob/master/packages/tap-runner/src/tap-test-runner.ts#L131) and [the parser complete callback](https://github.com/stryker-mutator/stryker-js/blob/master/packages/tap-runner/src/tap-test-runner.ts#L136) to promises that you can then wait on with `Promise.all`.

nicojs:
@danny12321 would you mind picking this one up?

danny12321:
Yes I will look into it.

danny12321:
@nicojs I have added a test that ensures that the process has exit before resolving the result, but I am not very happy about the way the `spawn`  method and `TapParser` class are mocked. My experience isn't that great with Sinon and ran into some weird problems so could we please sit together to make it a bit prettier?

nicojs:
Sure. For now feel free to merge it. We'll look at it later 👍

danny12321:
#4123 is merged. Closing this issue.

PR #4123, opened 2023-04-18
Fix(tap-runner): wait for spawned process to exit

danny12321:
A (possible) fix for the issue described in [#4122.](https://github.com/stryker-mutator/stryker-js/issues/4122). The tap-runner now waits for the spawned process to exit before returning the results.

I think this will fix the problem we encounter in de ci pipeline, but I could not recreate the problem locally.

### DIFF
--- a/packages/tap-runner/src/setup/env.cts
+++ b/packages/tap-runner/src/setup/env.cts
@@ -1,3 +1,6 @@
 export const strykerDryRun = 'STRYKER_DRY_RUN';
 export const strykerHitLimit = 'STRYKER_HIT_LIMIT';
 export const strykerNamespace = 'STRYKER_NAMESPACE';
+export function tempTapOutputFileName(pid: number | undefined): string {
+  return `stryker-output-${pid}.json`;
+}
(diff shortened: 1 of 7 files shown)


nicojs (changes requested):
I've refactored the `runFile` method. This had 2 reasons:

1. I wasn't happy with having multiple levels of abstraction: it was responsible for running the file as well as handling the `tap` details. Extracting the `tap` details to a `parseTap` function inside the `tap-helper.ts` file improves readability and maintainability.
2. It had a small bug. What happens when the async function rejects here:
   ```ts
   const parser = new tap.Parser(config, async (result) => {
       await stuff;
       resolve(...)
   });
   ```
   My guess is that the tap runner worker process would exit unexpectedly. By refactoring it so the root promise gets rejected, it will be handled by Stryker core's worker code.

@danny12321, what do you think about this new setup?

> but I could not recreate the problem locally.

Indeed. I think this is a good reason to add a unit test, with a line of comment pointing to the issue. Could you add that, please? 😇

nicojs:
BTW the fix is working. No false negatives have happened in this PR yet. I've run it a couple of times now.

PR #4134, opened 2023-04-22
Remove non-existent file data from mutation test results

mkizka:
fix: https://github.com/stryker-mutator/stryker-js/issues/4085
Hello. I have read the CONTRIBUTING.md, but please let me know if there are any mistakes.

### DIFF
--- a/packages/core/src/mutants/incremental-differ.ts
+++ b/packages/core/src/mutants/incremental-differ.ts
@@ -230,7 +230,12 @@ export class IncrementalDiffer {
       Object.entries(testFiles ?? {}).forEach(([fileName, oldTestFile]) => {
         const relativeFileName = toRelativeNormalizedFileName(fileName);
         const currentFileSource = currentRelativeFiles.get(relativeFileName);
-        if (currentFileSource !== undefined && oldTestFile.source !== undefined) {
+        if (currentFileSource === undefined && fileName !== '') {
+          // An empty file name means the test runner cannot report test file locations.
+          // If the current file is undefined while the test runner can report test file locations, it means it has been deleted
+          log.debug('Test file removed: %s', relativeFileName);
+          testStatisticsCollector.count(relativeFileName, 'removed', oldTestFile.tests.length);
+        } else if (currentFileSource !== undefined && oldTestFile.source !== undefined) {
           log.trace('Diffing %s', relativeFileName);
           const locatedTests = closeLocations(oldTestFile);
           const { results, removeCount } = performFileDiff(oldTestFile.source, currentFileSource, locatedTests);
--- a/packages/core/src/reporters/mutation-test-report-helper.ts
+++ b/packages/core/src/reporters/mutation-test-report-helper.ts
@@ -4,7 +4,7 @@ import { Location, Position, StrykerOptions, MutantTestCoverage, MutantResult, s
 import { Logger } from '@stryker-mutator/api/logging';
 import { commonTokens, tokens } from '@stryker-mutator/api/plugin';
 import { Reporter } from '@stryker-mutator/api/report';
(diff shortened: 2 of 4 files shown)


nicojs:
@mkizka, thanks for picking this one up!

I was thinking this should be fixed in the `incremental-differ.ts` file instead. The `IncrementalDiffer` is responsible for diffing the incremental report json file with the project's current status (discovered during the initial test run). 

It apparently keeps deleted test files in the report. You can fix that by applying this code change:

```diff
@@ -242,13 +242,9 @@ export class IncrementalDiffer {
             byKey.set(key, testInfo);
           });
         } else {
-          // No sources to compare, we should do our best with the info we do have
-          oldTestFile.tests.map((test) => {
-            const key = testToIdentifyingKey(test, relativeFileName);
-            const testInfo = { key, test, relativeFileName };
-            byId.set(test.id, testInfo);
-            byKey.set(key, testInfo);
-          });
+          // No sources to compare to, test file is removed.
+          log.debug('Test file removed: %s', relativeFileName);
+          testStatisticsCollector.count(relativeFileName, 'removed', oldTestFile.tests.length);
         }
       });
```

After that, some tests are failing; my guess is because of incorrect test data. Do you think you feel up to the task?

nicojs (changes requested):
See my comment.

mkizka:
Wow, this change looks better. I'll try to fix.

packages/core/src/mutants/incremental-differ.ts
```diff
@@ -230,7 +230,12 @@ export class IncrementalDiffer {
       Object.entries(testFiles ?? {}).forEach(([fileName, oldTestFile]) => {
         const relativeFileName = toRelativeNormalizedFileName(fileName);
         const currentFileSource = currentRelativeFiles.get(relativeFileName);
-        if (currentFileSource !== undefined && oldTestFile.source !== undefined) {
+        if (currentFileSource === undefined && fileName !== '') {
+          // An empty file name means the test runner cannot report test file locations.
+          // If the current file is undefined while the test runner can report test file locations, it means it has been deleted
+          log.debug('Test file removed: %s', relativeFileName);
+          testStatisticsCollector.count(relativeFileName, 'removed', oldTestFile.tests.length);
+        } else if (currentFileSource !== undefined && oldTestFile.source !== undefined) {
```
nicojs, line 238: I'm missing new unit tests for these:

image

nicojs (changes requested):
Wow, an awesome job so far! Tests green with small tweaks 👍

Could you also add a test for e test file that is removed?

mkizka:
Thanks for your review. I have added the test.
About the fileName, what is the situation that it is empty? Should I add a test for this too?
```js
if (currentFileSource === undefined && fileName !== '') {
```

nicojs:
> Thanks for your review. I have added the test. About the fileName, what is the situation that it is empty? Should I add a test for this too?
> 
> ```js
> if (currentFileSource === undefined && fileName !== '') {
> ```

This happens for example, when using the karma runner. It reports the tests without the file from which they originated. In that case, the JSON contains something like this:

```json
{
"testFiles": {
    "": {
      "tests": [
        { "id": "0", "name": "Add should be able to add two numbers" },
        { "id": "1", "name": "Add should be able 1 to a number" },
        { "id": "2", "name": "Add should be able negate a number" },
        { "id": "3", "name": "Add should be able to recognize a negative number" },
        { "id": "4", "name": "Add should be able to recognize that 0 is not a negative number" },
        { "id": "5", "name": "Circle should have a circumference of 2PI when the radius is 1" }
      ]
    }
  }
}
```

You're currently missing a test case for precisely that:

image

Since it is a big part of the feature set, would you mind adding a test for that as well?

mkizka:
I understand. I have added one more test.

nicojs (approved):

PR #4158, opened 2023-05-04
fix(reporters): add offset when reporting a single mutant

xandervedder:


packages/core/test/unit/reporters/mutation-test-report-helper.spec.ts
```diff
@@ -540,14 +540,34 @@ describe(MutationTestReportHelper.name, () => {
...
         };
         expect(actual).deep.include(expected);
       });
+
+      it('should not modify the original location property', () => {
+        // Arrange
+        const input = factory.mutantTestCoverage({
+          fileName: 'add.js',
+          id: '3',
+          location: factory.location(),
```
nicojs, line 557: You probably don't want to rely on the defaults for this test, since you assert them below. I always try to be explicit in the tests.

packages/core/src/reporters/mutation-test-report-helper.ts
```diff
@@ -97,6 +97,7 @@ export class MutationTestReportHelper {
   }
 
   private reportOne(result: MutantResult): MutantResult {
+    result.location = this.toLocation(result.location);
```
nicojs, line 100: I understand that this is more DRY than doing it in `reportMutantRunResult`, but the result is that the construction of `result` is now spread across 2 places. Also you're mutating an object on the heap, which can lead to undesired consequences (it's convoluted).

Can you move this to `reportMutantRunResult`? You can still do the location calculation once atop that function.

nicojs (changes requested):

packages/core/src/reporters/mutation-test-report-helper.ts
nicojs: Yes, that's probably better than the 2-staged way of doing now, do you agree?

xandervedder: Seems good, since we keep the creation/modification outside of the `reportOne` method (that should only report and not do anything else).

nicojs (approved):

PR #4187, opened 2023-05-13
Feat(config): Add 'always' option to cleanTempDir

adeelpm:
This 'always' option removes stryker-tmp folder irrespective of successful run or not


packages/api/schema/stryker-core.json
```diff
@@ -448,8 +448,12 @@
       "default": ".stryker-tmp"
     },
     "cleanTempDir": {
-      "description": "Choose whether or not to clean the temp dir (which is \".stryker-tmp\" inside the current working directory by default) after a successful run. The temp dir will never be removed when the run failed for some reason (for debugging purposes)",
-      "type": "boolean",
+      "description": "Choose whether or not to clean the temp dir (which is \".stryker-tmp\" inside the current working directory by default) after a run. The behavior depends on the value selected: if set to `true`, the temp dir will be removed only after a successful run; if set to `false`, the temp dir will be left in place after a run. If set to `always`, the temp dir will be removed after every run, regardless of whether it was successful or not.",
```
nicojs, line 451:
```suggestion
      "description": "Choose whether or not to clean the temp dir (which is \".stryker-tmp\" inside the current working directory by default).\n - `false`: Never delete the temp dir;\n - `true`: Delete the tmp dir after a successful run;\n - `'always'`: Always delete the temp dir, regardless of whether the run was successful.",
```

packages/core/src/stryker.ts
```diff
@@ -45,8 +45,10 @@ export class Stryker {
         return mutantResults;
       } catch (error) {
         const log = loggerProvider.resolve(commonTokens.getLogger)(Stryker.name);
-        log.debug('Not removing the temp dir because an error occurred');
-        mutantInstrumenterInjector.resolve(coreTokens.temporaryDirectory).removeDuringDisposal = false;
+        if (mutantInstrumenterInjector.resolve(coreTokens.temporaryDirectory).removeDuringDisposal !== 'always') {
```
nicojs, line 48:
```suggestion
        if (mutantInstrumenterInjector.resolve(commonTokens.options).cleanTempDir !== 'always') {
```

You should be able to use the `commonTokens` here to let the `StrykerOptions` be injected. You can use that to set the `removeDuringDisposal` to either `true` or `false`.

packages/core/src/utils/temporary-directory.ts
```diff
@@ -12,7 +12,7 @@ import { objectUtils } from './object-utils.js';
 export class TemporaryDirectory implements Disposable {
   private readonly temporaryDirectory: string;
   private isInitialized = false;
-  public removeDuringDisposal: boolean;
+  public removeDuringDisposal: boolean | string;
```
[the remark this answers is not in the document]
adeelpm: Hi, made all the changes as you suggested but if I make this line back to  `removeDuringDisposal: boolean;` I have to make the line 20 to `this.removeDuringDisposal = options.cleanTempDir as boolean;` as it throws "Type 'string | boolean' is not assignable to type 'boolean'" otherewise.

packages/core/src/utils/temporary-directory.ts
```diff
@@ -17,7 +17,7 @@ export class TemporaryDirectory implements Disposable {
   public static readonly inject = tokens(commonTokens.logger, commonTokens.options);
   constructor(private readonly log: Logger, options: StrykerOptions) {
     this.temporaryDirectory = path.resolve(options.tempDirName);
-    this.removeDuringDisposal = options.cleanTempDir;
+    this.removeDuringDisposal = typeof options.cleanTempDir === 'boolean' ? options.cleanTempDir : options.cleanTempDir == 'always' ? true : false;
```
nicojs, line 20:
```suggestion
    this.removeDuringDisposal = Boolean(options.cleanTempDir)
```
This should work here, right? Seems a lot simpler.

packages/core/test/unit/stryker.spec.ts
```diff
@@ -51,7 +53,9 @@ describe(Stryker.name, () => {
       .withArgs(commonTokens.getLogger)
       .returns(getLoggerStub)
       .withArgs(coreTokens.temporaryDirectory)
-      .returns(temporaryDirectoryMock);
+      .returns(temporaryDirectoryMock)
+      .withArgs(commonTokens.options)
+      .returns(strOptions);
```
nicojs, line 58:
```suggestion
      .returns(testInjector.options);
```

You can use the `testInjector.options` here. It is a helper object that contains options you can used. They will be cleaned after each test (reset to default)

packages/core/test/unit/stryker.spec.ts
```diff
@@ -197,4 +201,50 @@ describe(Stryker.name, () => {
...
+      cliOptions = {};
+      sut = new Stryker(cliOptions, () => injectorMock as typedInject.Injector<MutationTestContext>);
+    });
+
+    it('should not disable `removeDuringDisposal` on the temp dir when dry run rejects and cleanTempDir is set to `always`', async () => {
+      dryRunExecutorMock.execute.rejects(new Error('expected error for testing'));
+      await expect(sut.runMutationTest()).rejected;
+      expect(temporaryDirectoryMock.removeDuringDisposal).not.false;
+    });
+  });
```
nicojs, line 249: Why do you recreate an entire `describe` here? You should be able to just put this test in the previous `describe`:

```ts
  it('should not disable `removeDuringDisposal` on the temp dir when dry run rejects and cleanTempDir is set to `always`', async () => {
      testInjector.options.cleanTempDir= 'always';
      dryRunExecutorMock.execute.rejects(new Error('expected error for testing'));
      await expect(sut.runMutationTest()).rejected;
      expect(temporaryDirectoryMock.removeDuringDisposal).not.false;
    });
```

nicojs (changes requested):
Thanks for taking the time to make the requested changes! I have some new remarks on your new code 😅. If you need more help, please let me know.

packages/core/test/unit/stryker.spec.ts
adeelpm: Sorry for my blunder here, removed the describe section and I have a question why this line
`expect(temporaryDirectoryMock.removeDuringDisposal).not.false;`  when set to 
`expect(temporaryDirectoryMock.removeDuringDisposal).true;` throws `AssertionError: expected undefined to be true`

nicojs: Ah yes, this is because `temporaryDirectoryMock` is a mock, created on line 36. The property `removeDuringDisposal` is simply never set, so it doesn't get reflected here.

nicojs: I also added a unit test for the in `temporary-directory.spec.ts`: `"should remove the dir if cleanTempDir option is 'always'"`

nicojs (approved):

packages/core/test/unit/stryker.spec.ts
adeelpm: Thank for doing that.

Issue #4197, opened 2023-05-17
Allow Stryker to exit without any errors when no tests are found


emir7:
**Is your feature request related to a problem? Please describe.**
By default Stryker will exit with an error if no tests are found which lets you know that you might need to double check your Stryker configuration. However, there might be cases where you want to prevent Stryker from exiting with an error if no related tests are found since you are doing it intentionally. 

**Describe the solution you'd like**
Add a new configuration option (e.g. `--allowEmpty`) that lets the Strykers main process to exit without any errors in such cases and provide an additional log e.g. `No tests were found ` to let developer know that there are no related test given the current Stryker configuration.

PR #4198, opened 2023-05-17
Feat (config): Add --allowEmpty option

emir7:
A potential solution for [#4197](https://github.com/stryker-mutator/stryker-js/issues/4197)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -13,9 +13,9 @@ Keep in mind, that each option used on the command line will **completely replac
 ## Usage of Globbing Expressions on options
 
 The [`ignorePatterns`](#ignorepatterns-string), [`mutate`](#mutate-string) and other options referring to source files support globbing expressions using [node glob](https://github.com/isaacs/node-glob). See [the config file documentation](./config-file.md#glob-patterns) for more information. We suggest using [https://globster.xyz/](https://globster.xyz/) or referring to [`.gitignore`-style glob pattern](https://git-scm.com/docs/gitignore#_pattern_format).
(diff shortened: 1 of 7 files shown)


docs/configuration.md
```diff
@@ -13,9 +13,9 @@ Keep in mind, that each option used on the command line will **completely replac
 ## Usage of Globbing Expressions on options
```
nicojs, line 13: I see you've formatted this file with something. Please undo that, as this will pollute our history.

nicojs (changes requested):
This looks great. Just a small remark.

docs/configuration.md
emir7: not sure why the file was exactly formatted, but it should be fine now

nicojs: Thanks! It's not your fault though, it is normal to use a formatted in your editor. We should use formatter to verify it in the pipeline...

nicojs (approved):

emir7:
Seems like the tests are failing for some reason on Windows 😅 the error is a little bit odd and is related to the `vitest-runner`. It prints the following: 
`Error: EBUSY: resource busy or locked, rmdir 'D:\a\stryker-js\stryker-js\packages\vitest-runner\testResources\tmp\workDir-7081502'`
not 100% sure, but it looks like it might something to do with the fact that processes aren't fully closed in pipelines during the test run

nicojs:
Yeah, probably. I think it's a bug in Vitest itself (we're calling and awaiting `.close()`) We've just merged the Vitest runner and I'm seeing these errors for the first time... I'll restart the builds

PR #4246, opened 2023-05-31
feat(type-checking): disable type check by default

### LINKED ISSUE #4204: Change the default of `disableTypeChecks`
The default value for `disableTypeChecks` is `'{test,src,lib}/**/*.{js,ts,jsx,tsx,html,vue}'`. This does an assumption that your mutated and test code lives inside `test`. `src` or `lib`. Also the default extension does not include `.mts` or `.cts`. 

We should change it to be `true`.

**Additional context**
Many new StrykerJS use typescript type-checking via something like `ts-jest`. Using `true` here dramatically improves their experience. This is technically a breaking change, so we should go with v7.

nicojs:
Change the default of [`disableTypeChecks`](https://stryker-mutator.io/docs/stryker-js/configuration/#disabletypechecks-boolean--string) to `true`.

Closes #4204


SchroederSteffen:
Small hint: The docs aren't updated yet. https://github.com/stryker-mutator/stryker-js/blob/90ebf47767eafede567c8e39f82569078d5bcfbf/docs/configuration.md?plain=1#L200-L206

Issue #4266, opened 2023-06-05
Extend tap-runner plugin configuration to allow using "any" testing framework with TAP support


ericcornelissen:
**Is your feature request related to a problem? Please describe.**
Currently the [`tap-runner`](https://stryker-mutator.io/docs/stryker-js/tap-runner/) is limited to test frameworks that operate in a separate way (notably [`node-tap`](https://node-tap.org/) and the [Node.js built-in test runner](https://nodejs.org/api/test.html)). Other test frameworks (at least [AVA](https://github.com/avajs/ava)) that support the Test Anything Protocol (TAP) may not work due to the [way the plugin spawns process](https://github.com/stryker-mutator/stryker-js/blob/b159f73a9b8327c6cc1694a4c1b76b58d7701430/packages/tap-runner/src/tap-test-runner.ts#L128). Extending the `tap-runner` plugin to support more testing frameworks that support the TAP could provide a relatively lightweight way to have _some_ integration with more frameworks.

**Describe the solution you'd like**
I see two options:

1. Add an option to the `tap-runner` plugin to select the testing framework and implement a runner for each supported framework. The configuration could look something like:

   ```js
   { 
     testRunner: "tap",
     tap: {
       testFiles: [/* bla */],
       nodeArgs: [/* bla */],
       framework: "ava", // or "node-tap" or "node-native", probably defaults to "node-native"
     },
   }
   ```

   The main benefit of this I see is that it's simpler to understand, especially for users. The main drawback is that every framework has to be supported by manually adding a runner for it.

2. Extend the `tap-runner` plugin configuration to allow:
   - Specifying a "program" to run and corresponding "programArgs".
   - Specifying an alternative way to inject the `TapTestRunner.hookFile`.
   
   This could look something like:

   ```js
   { 
     testRunner: "tap",
     tap: {
       program: "./node_modules/.bin/ava",
       programArgs: ["--tap", "--node-arguments='-r %HOOKFILE%'"]
     },
   }
   ```

   If we go this route, I'd additionally suggest including `./node_modules/.bin` in the `PATH` so invoking any available commands is more convenient.

**Describe alternatives you've considered**
I described two possibilities above. A dedicated runner for AVA would also work for me, this has been discussed before in issues and discussions. I think supporting AVA and other frameworks with TAP support in this way would be beneficial regardless.

**Additional context**
See the discussion over at #3245 for more details and context.

nicojs:
[...]

PR #4283, opened 2023-06-09
feat(tap-runner): extend tap configuration

danny12321:
Extend the ability to configure the tap-runner to support frameworks that use their own runtime and support tap output. This feature request came from #4266 to support the Ava test framework.


packages/tap-runner/src/tap-helper.ts
```diff
@@ -58,3 +58,27 @@ export async function captureTapResult(tapProcess: ChildProcessWithoutNullStream
   }
   return tapResult;
 }
+
+export function parseArguments(args: string[], hookFile: string, testFile: string): string[] {
```
nicojs, line 62: It's not really parsing. How about `buildArguments`?

packages/tap-runner/src/tap-helper.ts
```diff
@@ -58,3 +58,27 @@ export async function captureTapResult(tapProcess: ChildProcessWithoutNullStream
...
+    if (index >= 0) {
+      newArgs[index] = newArgs[index].replace(arg.key, arg.replacement);
+    } else {
+      if (arg.placeInFront) {
+        newArgs.unshift(...arg.addition);
+      } else {
+        newArgs.push(...arg.addition);
+      }
+    }
+  });
```
nicojs, line 81: This seems pretty complicated. How about: 

```suggestion
const hooksFilePlaceholder = '{{hooksFile}}';
const testFilePlaceholder = '{{testFile}}';
if(!args.some(arg => arg.includes(hooksFilePlaceholder )){
   args = ['-r', hooksFilePlaceholder , ...args];
}
if(!args.some(arg => arg.includes(testFilePlaceholder ){
  args = [...args, testFilePlaceholder];
}
return args.map(arg => arg.replaceAll(testFilePlaceholder, testFile).replaceAll(hooksFilePlaceholder, hooksFile);
```

packages/tap-runner/test/unit/tap-helper.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+
+      // Assert
+      expect(result).to.eql(['-r', 'hook.js', 'custom-arg', 'test.js']);
+    });
+
+    it('should replace {{hookFile}} and {{testFile}} in the arguments', () => {
+      const result = parseArguments(['--file={{hookFile}}', '--test={{testFile}}'], 'hook.js', 'test.js');
+
+      // Assert
+      expect(result).to.eql(['--file=hook.js', '--test=test.js']);
```
nicojs, line 18:
```suggestion
      expect(result).deep.eql(['--file=hook.js', '--test=test.js']);
```

packages/tap-runner/test/unit/tap-helper.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { expect } from 'chai';
+
+import { parseArguments } from '../../src/tap-helper.js';
+
+describe('tap-helper', () => {
```
nicojs, line 5: Great to see some more unit tests 🤩.

Maybe also add tests for multiple instances of the placeholder.

packages/tap-runner/test/integration/tap-test-runner.it.spec.ts
```diff
@@ -208,4 +208,41 @@ describe('tap-runner integration', () => {
...
+        const run = await sut.dryRun(factory.dryRunOptions());
+
+        // Assert
+        assertions.expectCompleted(run);
+      }
+    });
+
+    function getAvaLocation() {
+      if (os.platform() === 'win32') {
+        return `${sandbox.tmpDir}/../../../node_modules/ava/entrypoints/cli.mjs`;
```
nicojs, line 242: This should actually also work for linux, right?

nicojs, line 242:
```suggestion
        return path.resolve(sandbox.tmpDir, '..', '..', '..', 'node_modules', 'ava', 'entrypoints', 'cli.mjs');
```

Maybe a small improvement?

packages/tap-runner/src/tap-helper.ts
danny12321: My thoughts was that it parses the template variables, but it indeed also builds the arguments if it is incomplete. I don't know a better name for now than `buildArguments`, but it isn't describing all the functionality in my opinion.

packages/tap-runner/test/integration/tap-test-runner.it.spec.ts
danny12321: Indeed. Will change this to only use the full path (like on windows).

packages/tap-runner/test/integration/tap-test-runner.it.spec.ts
```diff
@@ -208,4 +208,33 @@ describe('tap-runner integration', () => {
...
+      });
+      sut = testInjector.injector.injectFunction(createTapTestRunnerFactory('__stryker2__'));
+      await sut.init();
+    });
+
+    it('should be able to run', async function () {
+      // Test fails if debugger is used, because ava doesn't support debugging and tap output simultaneously.
+      // See: https://github.com/avajs/ava/blob/568fe40c987dd6c593dfbcf4144d1d1627955d46/lib/cli.js#L307
+      const { default: inspector } = await import('node:inspector');
+
```
nicojs, line 229: This is not needed. What we should do instead is disable `"autoAttachChildProcesses"` in launch.json. See https://rcrdo.com/2018/08/07/favorite-new-visual-studio-code-features-of-2018/#_node-debugging

PR #4290, opened 2023-06-11
set NODE_ENV to test in vitest runner

### LINKED ISSUE #4289: running with vitest not setting the `NODE_ENV` to `test`
**Summary**

vitest set NODE_ENV to test, and when running mutation it does not

Same behavior as jest
#3580

**Stryker environment**

```
npm ls | grep stryker
├── @stryker-mutator/core@7.0.2
├── @stryker-mutator/vitest-runner@7.0.2
```

```
vitest@0.32.0
```

**Your Environment**

| ---------------- | -------
| node             | 16.18.1
| npm              | 8.19.2
| Operating System | macOs

rluvaton:
fix #4289

I would like to add tests but I'm not really sure how, there are a lot of configuration needed, and coping from the jest test is not really fit


packages/vitest-runner/src/vitest-test-runner.ts
```diff
@@ -89,6 +89,7 @@ export class VitestTestRunner implements TestRunner {
   }
 
   private async run(testIds: string[] = []): Promise<DryRunResult> {
+    this.setEnv();
```
nicojs, line 92: I think this could be set in `.init()`, instead of `run`.

nicojs (changes requested):
Thanks for this PR! 🙏

I think it's OK to add a unit test for this. When you set the env in `init`, you can simply add a unit test to it in "test/unit/vitest-runner.spec.ts".

rluvaton:
I will try to update later this, week, thanks

rluvaton:
Hey @nicojs , there are a lot of missing things in the `createVitestMock` like `projects` and `state`, is there a reason to create these unit tests?

I tried to avoid unit tests and focus on component tests to reduce the number of changes I need to do in the tests when I do a refactor

rluvaton:
added the test and moved the set env to init

packages/vitest-runner/test/unit/vitest-runner.spec.ts
```diff
@@ -44,4 +44,12 @@ describe(VitestTestRunner.name, () => {
       await expect(sut.dispose()).not.rejected;
     });
   });
+
+  it('should set the NODE_ENV environment variable to test in init', async () => {
+    expect(process.env.NODE_ENV).to.not.equal('test');
```
nicojs, line 49: Maybe it's better to delete it?

```js
delete process.env.NODE_ENV;
```

rluvaton:
> > is there a reason to create these unit tests
> 
> Generally, everything should be tested as low in the testing pyramid as possible. If at all possible. In this case it is possible, so yes, probably a unit test.

I tend to disagree, component tests are much easier to write (as you come from the user point of view) and cover much more while also protecting you from changing the tests when doing refactor across units

(IMO)


Take a look at:
- https://github.com/testjavascript/nodejs-integration-tests-best-practices
- https://kentcdodds.com/blog/the-testing-trophy-and-testing-classifications

packages/vitest-runner/test/unit/vitest-runner.spec.ts
rluvaton: agree, done

nicojs (approved):

rluvaton:
Hope to contribute to this project again soon!

It was great working with you!

PR #4326, opened 2023-06-23
Feature: added tap runner force bail

### LINKED ISSUE #4276: Make bail behavior of tap-runner configurable
The way [bailing is implemented the tap-runner](https://github.com/stryker-mutator/stryker-js/blob/ff87edfc4e9b611fba041017a8ed77415c58b503/packages/tap-runner/src/tap-helper.ts#L26-L31) is not compatible with all ways in which tests can be run. The top-level [disableBail](https://stryker-mutator.io/docs/stryker-js/configuration/#disablebail-boolean) option can be used to avoid such problems, but just disabling the tap-runner's "hard" bail functionality would suffice (and keep the main functionality of bail in place).

Add a new option to the [tap-runner configuration](https://stryker-mutator.io/docs/stryker-js/tap-runner/#configuring) to be able to disable its "hard" bail functionality, e.g.:

```js
{
  "testRunner": "tap",
  "tap": {
    // etc...
    "hardBail": false
  }
}
```

As mentioned, the top-level [disableBail](https://stryker-mutator.io/docs/stryker-js/configuration/#disablebail-boolean) option can be used as a "workaround" but this is disables more than is absolutely necessary. No other alternatives were considered.

**Additional context**
See the discussion over at #3245 for more details and context.

danny12321:
Added optional `forceBail` option to the `tap-runner`, because test runners which run the tests in a child process that gets killed by Stryker don't output the test result. 

Fixes #4276


packages/tap-runner/test/integration/tap-test-runner.it.spec.ts
```diff
@@ -115,58 +114,29 @@ describe('tap-runner integration', () => {
...
-    it('should bail out current process when disableBail is false and os type is not windows', async function () {
-      if (os.platform() === 'win32') {
-        this.skip();
-      } else {
+    [
+      { disableBail: false, forceBail: false, shouldBail: false },
+      { disableBail: true, forceBail: false, shouldBail: false },
+      { disableBail: true, forceBail: true, shouldBail: false },
+      { disableBail: false, forceBail: true, shouldBail: os.platform() === 'win32' ? false : true },
+    ].forEach(({ disableBail, forceBail, shouldBail }) => {
```
nicojs, line 122: Probably better not to generate tests in this case, because you now need a `if` inside the test to figure out what to assert. 

Better would be to create and use helper functions inside designated `it`'s

packages/tap-runner/test/integration/tap-test-runner.it.spec.ts
```diff
@@ -115,58 +114,29 @@ describe('tap-runner integration', () => {
...
-
-      // Act
-      const run = await sut.mutantRun(factory.mutantRunOptions({ testFilter: testFiles, disableBail: true }));
-      const endTime = Date.now();
-
-      // Assert
-      assertions.expectKilled(run);
-      expect(endTime - startTime).gte(4000);
+        if (shouldBail) {
+          expect(timeDiff).lte(2000);
```
nicojs, line 135: Asserting test duration can be flaky. For example: when debugging the test you might be lingering in a breakpoint for 10 seconds.

Can we assert this some other way? For example using the failure message?

nicojs (changes requested):

nicojs (approved):

PR #4358, opened 2023-07-07
bug(tap-runnner): time spent not shown when tap result has no time spent

### LINKED ISSUE #4331: Tap runner should correctly report time spent when tap output lacks time
**Summary**

The tap-runner currently relies on test reporting from the `timeSpentMs`. However, time reporting in TAP output is optional.

<details>
<summary>With timing</summary>

```
nicojs@NicoJ01 ~/github/stryker-js/packages/tap-runner/testResources/example (master)$ node tests/math.spec.js
TAP version 13
# Subtest: Adding two numbers
    ok 1 - Adding 10 and 5 equal to 15
    ok 2 - Adding 100 and 200 equal to 300
    1..2
ok 1 - Adding two numbers # time=3.708ms

# Subtest: Slow adding two numbers
    ok 1 - Adding 10 and 5 equal to 15
    ok 2 - Adding 100 and 200 equal to 300
    1..2
ok 2 - Slow adding two numbers # time=0.604ms

1..2
# time=8.128ms
```

</details>

<details>
<summary>Without timing</summary>

```
nicojs@NicoJ01 ~/github/stryker-js/packages/tap-runner/testResources/ava (master)$ npx ava --tap tests/math.spec.js
TAP version 13

1..0
# tests 0
# pass 0
# fail 0
```

</details>

The code that grabs timing info from tap output is here: https://github.com/stryker-mutator/stryker-js/blob/master/packages/tap-runner/src/tap-test-runner.ts#L156

This has 2 consequences:

1. The timeout calculation doesn't work correctly, which can result in unexpected false positives regarding mutants that are reported as `timout` 
2. The progress reporter reports `NaN` as progress because it relies on the net time to report progress (and net time is always `0`). See https://github.com/ericcornelissen/shescape/pull/991#issuecomment-1605494916

To solve this issue, I think we should use our own timeout calculation based on elapsed time when the timing is not provided in TAP output. 

```ts
const now = () => new Date();
const before = now();
// Run file
const elapsedMs = now().getTime() - before.getTime();
```

danny12321:
Added custom time calculation for a test run which is used when the tap result does not include the time spent.

Fixes #4331

### DIFF
--- a/packages/tap-runner/src/tap-test-runner.ts
+++ b/packages/tap-runner/src/tap-test-runner.ts
@@ -136,8 +136,11 @@ export class TapTestRunner implements TestRunner {
     if (this.log.isDebugEnabled()) {
       this.log.debug(`Running: \`node ${args.map((arg) => `"${arg}"`).join(' ')}\` in ${process.cwd()}`);
     }
+    const now = () => new Date().getTime();
+    const before = now();
     const tapProcess = childProcess.spawn('node', args, { env });
     const result = await captureTapResult(tapProcess, !testOptions.disableBail && this.options.tap.forceBail);
+    const timeSpentMs = now() - before;
     const fileName = tempTapOutputFileName(tapProcess.pid);
     const fileContent = await fs.readFile(fileName, 'utf-8');
     await fs.rm(fileName);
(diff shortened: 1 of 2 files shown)


nicojs:
Looks 👍

PR #4407, opened 2023-09-08
feat(string-literals) change mutation of Symbols

### LINKED ISSUE #3870: Prevent string mutation when defining a Symbol
My project uses `Symbol` with a description as recommended by [`symbol-description`](https://eslint.org/docs/latest/rules/symbol-description) (I use [StandardJS](https://standardjs.com/)).
Yet this description is used mostly for debugging and can hardly be 'tested' (or there is no value to do so).
StrykerJS mutate the description and - as expected - the mutant survives.

I'd like StrykerJS to ignore the description of Symbol without having to ignore all static mutants.

I could disable the `symbol-description` rule but the goal of StandardJS is to avoid customization.

**Additional context**
I <3 StrykerJS

lex-funy:
This PR fixes the issue as described in https://github.com/stryker-mutator/stryker-js/issues/3870.

String literals and templates inside of Symbol declarations will no longer de mutated.

closes #3870


packages/instrumenter/src/mutators/string-literal-mutator.ts
```diff
@@ -8,17 +8,17 @@ export const stringLiteralMutator: NodeMutator = {
...
+    if (path.isStringLiteral() && isValidParentForStringLiteral(path)) {
       yield types.stringLiteral(path.node.value.length === 0 ? 'Stryker was here!' : '');
     }
   },
 };
 
-function isValidParent(child: NodePath<babel.types.StringLiteral>): boolean {
+function isValidParentForStringLiteral(child: NodePath<babel.types.StringLiteral>): boolean {
   const parent = child.parent;
   return !(
```
nicojs, line 23: This method now contains a lot of duplicate code. Can we call `isValidParentForTemplateLiteral` from this method instead? (so share the shared implementation). Or let them both call a `isValidParent` with the shared stuff in it?

nicojs (changes requested):
Thanks a lot so far! Could you take a look at my review comments?

nicojs (changes requested):
I've been thinking. I think it is perfectly valid to not take templated string literals into account.

For example: ``Symbol(`foo ${bar}`)``, I think it would be valid to mutate this one. Since the dev interpolated something into the string: I would also assume that to be important and tested (using `assert.eq(symbol.description, 'foo bar')`)

I would only ignore it in the simple case:

```
Symbol('foo')
Symbol("foo")
```

Not:

```
Symbol(`foo ${bar}`)
Symbol('foo' + bar)
```

Sorry for the back and forth. 😢

nicojs (approved):

PR #4465, opened 2023-10-05
chore: remove lodash.flatmap

hugo-vrijswijk:
Removes lodash.flatmap dependency. All environments support native flatMap nowadays, most lodash functions have a simple native implementation that saves a dependency.

Technically a breaking change as it removes the export from `@stryker-mutator/util`, but I assume nobody outside of this repo used it (why would you).


nicojs (approved):

nicojs:
thx! 🙋‍♀️

nicojs:
F.y.i. we don't support semver for `@stryker-mutator/util`

Issue #4386, opened 2023-08-03
Stryker copies binary file incorrectly : a pptx file (during initial test run)


edi9999:
**Summary**

Hello, it seems that the stryker command copies docx/pptx files incorrectly in my project.

It worked before (I can't tell when exactly, and which version I used before, but it worked 1 year ago I think).

**Stryker environment**

```
+-- @stryker-mutator/core@7.1.1
+-- @stryker-mutator/mocha-runner@7.1.1
```

**Your Environment**

| software         | version(s)
| ---------------- | -------
| node             | 18.17.0
| npm              | 9.6.7
| Operating System  | Ubuntu 22

**Reproduce the issue**

```
git clone https://github.com/open-xml-templating/docxtemplater.git
cd docxtemplater
npm install
npm run test:mutations
```

You can then view the difference between two files : 

```
$ ls examples/a16-row-id.pptx -lah
-rw-r--r-- 1 edgar edgar 39K Mar  7  2022 examples/a16-row-id.pptx
```

```
$ ls .stryker-tmp/sandbox*/examples/a16-row-id.pptx -lah
-rw-r--r-- 1 edgar edgar 64K Aug  3 20:57 .stryker-tmp/sandbox2472055/examples/a16-row-id.pptx
```

It seems that the copied file differ for some reasons.

edi9999:
I've just tested this with version 6.4.2 of stryker and it worked well with that version, so I think it has something to do with stryker 7.0

odinvanderlinden:
@edi9999 thanks for opening an issue. What is the problem you are facing with this bug?

edi9999:
The problem is that my software relies on some binary files during testing.

Those binary files are "pptx" documents, that need to be copied for the test run.

They are copied, but the contents of the file is incorrect : 


As you can see in this output : 

```
$ ls examples/a16-row-id.pptx -lah
-rw-r--r-- 1 edgar edgar 39K Mar  7  2022 examples/a16-row-id.pptx
$ ls .stryker-tmp/sandbox*/examples/a16-row-id.pptx -lah
-rw-r--r-- 1 edgar edgar 64K Aug  3 20:57 .stryker-tmp/sandbox2472055/examples/a16-row-id.pptx
```

The basefile is 39K big, and the copied file is 64K big, there is probably an encoding issue or something similar.

You can reproduce the issue with following script : 

```bash
git clone https://github.com/open-xml-templating/docxtemplater.git
cd docxtemplater
[...]

PR #4485, opened 2023-10-14
fix(core): disableTypeChecks true only forces js-like file match

randmon:
Solves issue #4386


nicojs:
@randmon thanks for this PR, can you fix the e2e test? Should be simple (`npm install` inside `e2e/test/use-binary-files/`)

nicojs (approved):
Great fix! Could you also update the docs here?

https://stryker-mutator.io/docs/stryker-js/configuration/#disabletypechecks-boolean--string

PR #4487, opened 2023-10-14
Feat/ignore mutants based on heuristics

### LINKED ISSUE #3229: Allow users to ignore mutants based on heuristics

We have had a couple of requests in the past to allow to ignore specific mutants:

#2966 #1174 #1470 #1464

Some of these issues have been solved by allowing users to [ignore using a comment](https://stryker-mutator.io/docs/stryker-js/disable-mutants#using-a--stryker-disable-comment), however, this isn't always practical.

For large codebases where you want to ignore the same pattern every time. For example: `console.log()`. 

Or you want to ignore mutants based on some other data, for example using the result of a git command, like the [git checker](https://github.com/lbtoma/stryker-git-checker) seems to be doing.

I would like to be able to ignore specific mutants based on heuristics (for lack of a better word). It would be run inside the instrumenter on each AST node visitation. It could then decide to ignore the current node with a reason. 

The API could look like this

```ts
type IgnoreResult = false | { ignore: true, reason: string };

interface Ignorer {
  init?(): Promise<void>;

  shouldIgnore(path: NodePath): IgnoreResult;  
}
```

We should document some examples for common use cases, like ignoring `console.log` statements. Issue #3230 needs to be solved to allow for locally defined plugins.

There are a number of issues linked where users thought of different approaches. However, I think this abstraction makes the most sense. It allows for the most freedom.

**Additional context**

All current plugin APIs are located inside the `@stryker-mutator/api` package. I want this to be no exception. However, the `NodePath` interface comes from `@babel/core`. I don't want that to be part of the dependencies of the API package. So we might want to mark it as a dev dependency 🤷‍♀️

odinvanderlinden:
Closes #3229


nicojs:
@odinvanderlinden Thanks for your work! As discussed, I've made some changes and am about to merge this 🎉

- I've moved the ignore bookkeeping to a separate class: `IgnorerBookkeeper`, and added some unit tests.
- I've rolled the change to the prettier config. I'd happily discuss prettier changes in a separate issue, but not inside a feature PR.
- I've merged the e2e test in 'ignore-project'. I remembered that this e2e project was designed to e2e test ignorers

PR #4542, opened 2023-11-01
feat(node): drop official support for node 16

nicojs:
BREAKING CHANGE: NodeJS 16 is no longer supported. Please use NodeJS 18 or higher. See https://nodejs.org/en/about/previous-releases


packages/core/test/unit/config/config-reader.spec.ts
```diff
@@ -242,7 +242,7 @@ describe(ConfigReader.name, () => {
 
       // Assert
       expect(error).instanceOf(ConfigError);
-      expect(((error as ConfigError).innerError as Error).message).eq('Unexpected token n in JSON at position 2');
+      expect(((error as ConfigError).innerError as Error).message).matches(/JSON at position 2/);
```
nicojs, line 245: Please improve

PR #4738, opened 2024-02-17
feat(typescript-checker): support prerelease TypeScript versions

davecardwell:
Stryker fails at launch when using `v5.4.0-beta` of TypeScript, displaying:
> `Error: @stryker-mutator/typescript-checker only supports typescript@3.6 or higher. Found typescript@5.4.0-beta`

This PR allows prerelease versions of TypeScript.


nicojs (approved):
I didn't know `includePrerelease` existed. Good find!

nicojs:
Released `8.2.4` with the fix 🚀

Issue #4917, opened 2024-07-31
Add mutation for date operation


boretti:
**Is your feature request related to a problem? Please describe.**
It would be interesting to have mutation produced for operation on Date.

Expression like :

```typescript
aDate.setDate(aDate.getDate()+1)
```
will produce interesting mutation based on the +1. It would also be interesting to have mutation changing the `setDate` or `getDate` operation (as updating month and not the day of month).

**Describe the solution you'd like**
I propose to add the following mutation (to `method-expression-mutator`) :

* `getDate` to/from `getMonth`
* `setDate` to/from `setMonth`
* `getHours` to/from `getMinutes`
* `setHours` to/from `setMinutes`

**Describe alternatives you've considered**

Maybe other method are interesting to be mutated, but I think, a first set to start with is enough

**Additional context**

I will also create a PR with the proposal and link it with this Feature Request.

nicojs:
This sounds like a great addition and is right in the spirit of what the method expression mutator is supposed to do. However, I don't see why we should stop at these four. For some use cases, people might be more interested in smaller or larger units of time.

This would be my suggestion.

| From                                                                                                                                            | To                                                                                                                                              |
| ----------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
| [Date.prototype.setDate()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate)                       | [Date.prototype.setTime()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setTime)                       |
| [Date.prototype.setFullYear()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setFullYear)               | [Date.prototype.setMonth()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setMonth)                     |
| [Date.prototype.setMonth()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setMonth)                     | [Date.prototype.setFullYear()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setFullYear)               |
[...]

PR #4918, opened 2024-07-31
feat(instrumenter): add expression mutator for Date (#4917)

boretti:
It would be interesting to have mutation produced for operation on Date.

Expression like :

```typescript
aDate.setDate(aDate.getDate()+1)
```
will produce interesting mutation based on the +1. It would also be interesting to have mutation changing the `setDate` or `getDate` operation (as updating month and not the day of month).

I propose to add the following mutation (to `method-expression-mutator`) :

* `getDate` to/from `getMonth`
* `setDate` to/from `setMonth`
* `getHours` to/from `getMinutes`
* `setHours` to/from `setMinutes`

See issue #4917


boretti:
Update based on proposal in the issue.

nicojs (commented):
The code looks great, But I first want to resolve the discussion in #4917 before merging

boretti:
Updated according exchange in the issue

nicojs (approved):

nicojs:
Thanks! ❤️

Issue #4936, opened 2024-08-06
Running `npm init stryker` use default npm registry instead of the one configured in npm config


Tamir-M:
**Summary**

When running `npm init stryker` in air-gapped network (i.e. without access to the default npm registry) the setup fail with connection problem, this is caused due to stryker not using the configured npm registry (e.g. in `.npmrc`)

**Stryker config**
 Not relevant

**Test runner config**

Not relevant

**Stryker environment**

- `create-stryker`: `8.3.0`

**Your Environment**

| software         | version(s)
| ---------------- | -------
| node             | 20.13.1
| npm              |  10.5.2
| Operating System | MacOS 14

nicojs:
Thanks, it seems like a great addition.

PR #4937, opened 2024-08-06
feat(core): use user configured npm registry

Tamir-M:
Fix:
- #4936

### DIFF
--- a/packages/core/src/initializer/index.ts
+++ b/packages/core/src/initializer/index.ts
@@ -1,5 +1,5 @@
 import { createInjector } from 'typed-inject';
-import { execaCommand } from 'execa';
+import { execaCommand, execaCommandSync } from 'execa';
 import { resolveFromCwd } from '@stryker-mutator/util';
 import { LogLevel } from '@stryker-mutator/api/core';
 
@@ -20,14 +20,15 @@ export function initializerFactory(): StrykerInitializer {
   LogConfigurator.configureMainProcess(LogLevel.Information);
   return provideLogger(createInjector())
     .provideValue(initializerTokens.out, console.log)
+    .provideValue(coreTokens.execa, execaCommand)
+    .provideValue(coreTokens.execaSync, execaCommandSync)
+    .provideValue(coreTokens.resolveFromCwd, resolveFromCwd)
     .provideFactory(initializerTokens.npmRegistry, getRegistry)
     .provideFactory(initializerTokens.restClientNpm, createNpmRegistryClient)
     .provideClass(initializerTokens.npmClient, NpmClient)
     .provideClass(initializerTokens.configWriter, StrykerConfigWriter)
     .provideClass(initializerTokens.gitignoreWriter, GitignoreWriter)
     .provideClass(initializerTokens.inquirer, StrykerInquirer)
-    .provideValue(coreTokens.execa, execaCommand)
-    .provideValue(coreTokens.resolveFromCwd, resolveFromCwd)
     .provideFactory(initializerTokens.customInitializers, createInitializers)
     .injectClass(StrykerInitializer);
 }
--- a/packages/core/src/initializer/index.ts
+++ b/packages/core/src/initializer/index.ts
@@ -1,6 +1,6 @@
 import { createInjector } from 'typed-inject';
 import { RestClient } from 'typed-rest-client';
-import { execaCommand } from 'execa';
+import { execaCommand, execaSync } from 'execa';
 import { resolveFromCwd } from '@stryker-mutator/util';
 import { LogLevel } from '@stryker-mutator/api/core';
 
@@ -16,13 +16,34 @@ import { StrykerInquirer } from './stryker-inquirer.js';
 import { createInitializers } from './custom-initializers/index.js';
 import { GitignoreWriter } from './gitignore-writer.js';
 
-const NPM_REGISTRY = 'https://registry.npmjs.com';
+let currentRegistry: string | undefined;
+
+function getRegistry(): string {
+  if (!currentRegistry) {
+    if (process.env.npm_config_registry) {
+      currentRegistry = process.env.npm_config_registry;
+    } else if (process.env.npm_command) {
+      // if running inside npm and not having the registry than it's the default one
+      currentRegistry = 'https://registry.npmjs.com';
+    } else {
+      // Using global as when trying to get the registry inside npm workspace it would fail
+      const registry = execaSync('npm', ['config', 'get', '--global', 'registry'], {
+        stdout: 'pipe',
+        timeout: 20000,
+      });
+
+      currentRegistry = registry.stdout;
+    }
+  }
+
+  return currentRegistry;
+}
 
 export function initializerFactory(): StrykerInitializer {
   LogConfigurator.configureMainProcess(LogLevel.Information);
   return provideLogger(createInjector())
     .provideValue(initializerTokens.out, console.log)
-    .provideValue(initializerTokens.restClientNpm, new RestClient('npm', NPM_REGISTRY))
+    .provideValue(initializerTokens.restClientNpm, new RestClient('npm', getRegistry()))
     .provideClass(initializerTokens.npmClient, NpmClient)
     .provideClass(initializerTokens.configWriter, StrykerConfigWriter)
     .provideClass(initializerTokens.gitignoreWriter, GitignoreWriter)
@@ -33,4 +54,4 @@ export function initializerFactory(): StrykerInitializer {
     .injectClass(StrykerInitializer);
 }
 
-export { initializerTokens };
+export { initializerTokens, getRegistry };
--- a/packages/core/src/initializer/index.ts
+++ b/packages/core/src/initializer/index.ts
@@ -1,12 +1,11 @@
 import { createInjector } from 'typed-inject';
-import { RestClient } from 'typed-rest-client';
-import { execaCommand, execaSync } from 'execa';
+import { execaCommand } from 'execa';
 import { resolveFromCwd } from '@stryker-mutator/util';
 import { LogLevel } from '@stryker-mutator/api/core';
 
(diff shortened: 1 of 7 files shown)


packages/core/src/initializer/index.ts
```diff
@@ -16,13 +16,34 @@ import { StrykerInquirer } from './stryker-inquirer.js';
...
+  }
+
+  return currentRegistry;
+}
 
 export function initializerFactory(): StrykerInitializer {
   LogConfigurator.configureMainProcess(LogLevel.Information);
   return provideLogger(createInjector())
     .provideValue(initializerTokens.out, console.log)
-    .provideValue(initializerTokens.restClientNpm, new RestClient('npm', NPM_REGISTRY))
```
nicojs, line 25: Could you instead use the dependency injection mechanism to provide a cached version of the npm registry? That way we get caching for free and it will be far easier to test.

Pseudo code:

```js
function createNpmRegistryClient(npmRegistry: string) {
  return new RestClient('npm', npmRegistry);
}
createNpmRegistryClient.inject = [initializerTokens.npmRegistry] as const;

// inside the chain
.provideFactory(initializerTokens.npmRegistry, getRegistry)
.provideFactory(initializerTokens.restClientNpm, createRestClient)
```

packages/core/src/initializer/npm-client.ts
```diff
@@ -65,7 +65,7 @@ export class NpmClient {
       const response = await this.innerNpmClient.get<NpmSearchResult>(path);
       return handleResult(path)(response);
     } catch (err) {
-      this.log.error(`Unable to reach 'https://registry.npmjs.com' (for query ${path}). Please check your internet connection.`, errorToString(err));
+      this.log.error(`Unable to reach '${getRegistry()}' (for query ${path}). Please check your internet connection.`, errorToString(err));
```
nicojs, line 68: Instead of using `getRegistry` here, you can simply let the npm registry location be injected, see previous comment about dependency injection

packages/core/src/initializer/index.ts
```diff
@@ -16,13 +16,34 @@ import { StrykerInquirer } from './stryker-inquirer.js';
 import { createInitializers } from './custom-initializers/index.js';
 import { GitignoreWriter } from './gitignore-writer.js';
 
-const NPM_REGISTRY = 'https://registry.npmjs.com';
+let currentRegistry: string | undefined;
+
+function getRegistry(): string {
```
nicojs, line 21: Can you add unit tests for this function? Also, move it to a separate file.

packages/core/src/initializer/index.ts
```diff
@@ -16,13 +16,34 @@ import { StrykerInquirer } from './stryker-inquirer.js';
...
+  if (!currentRegistry) {
+    if (process.env.npm_config_registry) {
+      currentRegistry = process.env.npm_config_registry;
+    } else if (process.env.npm_command) {
+      // if running inside npm and not having the registry than it's the default one
+      currentRegistry = 'https://registry.npmjs.com';
+    } else {
+      // Using global as when trying to get the registry inside npm workspace it would fail
+      const registry = execaSync('npm', ['config', 'get', '--global', 'registry'], {
```
nicojs, line 30: Please also add a `try ... catch` here, because a lot of stuff can go wrong here. If an error occurs we should log a warning (using the logger that you can inject) and use the default value,

packages/core/src/initializer/index.ts
```diff
@@ -16,13 +16,34 @@ import { StrykerInquirer } from './stryker-inquirer.js';
...
+        stdout: 'pipe',
+        timeout: 20000,
+      });
+
+      currentRegistry = registry.stdout;
```
nicojs, line 35: Should the result be `trim`med here? 
```suggestion
      currentRegistry = registry.stdout.trim();
```

nicojs (changes requested):
Thanks for the PR. I have some remarks.

packages/core/src/initializer/index.ts
Tamir-M: I am having some issues with the last case where I would need to use the `execaSync`, I have a few ways I can test this:

**Mocking:**
Would result in not actually testing the functionality of it, wouldn't do anything.

**Editing the `npm config`:**
This may cause a problem if I do `beforeEach` and `afterEach` to edit the `npm config` there are multiple things that could happen:
- If the command fails it can ruin the person's npm configuration.
- If the person aborts mid test it could ruin the person's npm configuration.

Seems like too be too many points of failure.
I have no idea how to test this or if I even should, as I gave a default value if it fails and warned them.
I would wait for your input.

Tamir-M:
@nicojs I've applied the changes.

packages/core/test/unit/initializer/npm-registry.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+      process.env.npm_command = oldNpmCommand;
+    });
+
+    it('should return default repository when run with npm command with no custom repository', () => {
+      process.env.npm_config_registry = '';
+
+      process.env.npm_command = 'value';
+
+      const registry = getRegistry(log4js.getLogger());
```
nicojs, line 29: Please use `testInjector.logger`. It contains a logger mock that is ready to be used. You can verify log messages on it using `sinon.asserts.calledOnceWithExactly(testInjector.logger.warn, '...');`

packages/core/test/unit/initializer/npm-registry.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(registry).to.equal(DEFAULT_REGISTRY);
+    });
+
+    it('should return default repository when run with npm command with custom repository', () => {
+      process.env.npm_config_registry = 'foo';
+
+      process.env.npm_command = 'value';
+
+      const registry = getRegistry(log4js.getLogger());
```
nicojs, line 39: Same here

packages/core/test/unit/initializer/npm-registry.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(registry).to.equal('foo');
+    });
+  });
```
nicojs, line 43: I'm missing tests for the happy flow with `execaSync` and the alternate flow (exception thrown).

packages/core/test/unit/initializer/stryker-initializer.spec.ts
```diff
@@ -55,7 +56,8 @@ describe(StrykerInitializer.name, () => {
     syncBuiltinESMExports();
     sut = testInjector.injector
       .provideValue(initializerTokens.out, out as unknown as typeof console.log)
-      .provideValue(initializerTokens.restClientNpm, npmRestClient)
+      .provideFactory(initializerTokens.npmRegistry, getRegistry)
+      .provideFactory(initializerTokens.restClientNpm, createNpmRegistryClient)
```
nicojs, line 60: Since this is a unit test for the `StrykerInitializer`, I wouldn't expect the "real" `getRegistry` to be used here. Instead, you can keep using `provideValue` here for the `npmRestClient` sinon mock here. 

Indeed, the tests are failing, see https://github.com/stryker-mutator/stryker-js/actions/runs/10341544430/job/28626286827?pr=4937

nicojs (changes requested):
Thanks for fixing my review comments. I have some final remarks regarding the tests.

packages/core/src/initializer/index.ts
nicojs: Ah yes, great question. Please add 2 more unit tests, 1 for the happy flow and one for the alternate flow using mocking of `execaSync`. You are right that we won't test the actual command `npm config get registry`, but I would expect that to work as documented. However, if you want to add a test for that, you are welcome to add an integration test (see `test/integration`).

Tamir-M:
@nicojs I've applied the changes you've asked for, would love to get them reviewed.

packages/core/test/unit/initializer/npm-registry.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+import { execaCommandSync } from 'execa';
+
+const DEFAULT_REGISTRY = 'https://registry.npmjs.com';
+
+describe('npm registry', () => {
+  describe('get registry', () => {
+    let oldNpmConfigRegistry: string | undefined;
+    let oldNpmCommand: string | undefined;
+
+    before(() => {
```
nicojs, line 14: This should be `beforeEach` instead of `before` (or change `afterEach` to be `after`)

packages/core/test/unit/initializer/npm-registry.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+      process.env.npm_config_registry = '';
+      process.env.npm_command = '';
+
+      const registry = getRegistry(testInjector.logger, execaCommandSyncMock as unknown as typeof execaCommandSync);
+
+      sinon.assert.callCount(execaCommandSyncMock, 1);
+      sinon.assert.calledWith(execaCommandSyncMock, 'npm config get --global registry', {
+        stdout: 'pipe',
+        timeout: 20000,
+      });
```
nicojs, line 59: Can be simplified to [`calledOnceWithExactly`](https://sinonjs.org/releases/latest/assertions/#sinonassertcalledoncewithexactlyspyorspycall-arg1-arg2-)

packages/core/test/unit/initializer/npm-registry.spec.ts
```diff
@@ -0,0 +1,87 @@
...
```
nicojs, line 78: Same here

nicojs (changes requested):
Thanks a lot for implementing the requested changes! I've got small remarks about the use of test hooks and assertions. Other than that this looks good.

Tamir-M:
Thank you for the remarks, I've applied them.

nicojs (approved):
LGTM

Issue #5015, opened 2024-09-17
Stryker `jest-runner` fails with jest prerelease versions above v27 when `coverageAnalysis` is set to "perTest"


rriski:
**Summary**

Stryker `jest-runner` crashes with `Cannot read properties of undefined (reading 'getEnv')` at this line https://github.com/stryker-mutator/stryker-js/blob/2faded7848d67ec306e1f65a00586c5cd3f6604d/packages/jest-runner/src/jest-plugins/jasmine2-setup-coverage-analysis.cts#L2 with these options: `"testRunner": "jest", `"coverageAnalysis": "perTest"` when running with a jest prerelease version (e.g. latest jest prerelease version at the time is [30.0.0-alpha.6](https://www.npmjs.com/package/jest/v/30.0.0-alpha.6).

The bug is triggered when `coverageAnalysis`is set to `perTest` as that invokes `withCoverageAnalysis` that in turn uses `semver` package to test the jest current version and checks if its >= 27. This check fails when using a prerelease version:

    > semver.satisfies('30.0.0-alpha.6', '>=27')
    false
    > semver.satisfies(semver.coerce('30.0.0-alpha.6'), '>=27')
    true

**Stryker config**

Relevant config options to trigger the bug are `"testRunner": "jest"` and `"coverageAnalysis": "perTest"` `withCoverageAnalysis` that in turn checks the jest version to set some configuration options. The version check doesn't coerce prerelease versions so the else branch is taken [here](https://github.com/stryker-mutator/stryker-js/blob/2faded7848d67ec306e1f65a00586c5cd3f6604d/packages/jest-runner/src/jest-plugins/with-coverage-analysis.ts#L26) for jest version 30.0.0-alpha.6.

```json
{
  "$schema": "./node_modules/@stryker-mutator/core/schema/stryker-schema.json",
  "testRunner": "jest",
  "reporters": [
    "progress",
    "clear-text",
    "html"
  ],
  "dryRunTimeoutMinutes": 20,
  "timeoutMS": 1200000,
  "incremental": true,
  "coverageAnalysis": "perTest",
  "logLevel": "trace",
  "checkerNodeArgs": ["--max-old-space-size=8192"],
  "testRunnerNodeArgs": ["--max-old-space-size=8192"],
  "plugins": ["@stryker-mutator/jest-runner"],
  "jest": {
    "projectType": "custom",
    "configFile": "jest.config.js",
    "enableFindRelatedTests": false
  },
[...]

PR #5016, opened 2024-09-17
fix: jest prerelease support (#5015)

rriski:
Previously, stryker failed `jasmine2-setup-coverage-analysis.cts`

    runTest ... node_modules/jest-runner/build/index.js:340:7), Cannot read
    properties of undefined (reading 'getEnv') TypeError: Cannot read properties
    of undefined (reading 'getEnv')

When using the latest jest prerelease version (30.0.0-alpha.6)

This commit fixes the error by using `semver.coerce`:

    > semver.satisfies('30.0.0-alpha.6', '>=27')
    false
    > semver.satisfies(semver.coerce('30.0.0-alpha.6'), '>=27')
    true


packages/jest-runner/src/jest-test-adapters/jest-test-adapter-factory.ts
```diff
@@ -15,7 +15,7 @@ export function jestTestAdapterFactory(
   options: StrykerOptions,
   injector: Injector<JestPluginContext>,
 ): JestGreaterThan25TestAdapter | JestLessThan25TestAdapter {
-  const version = jestWrapper.getVersion();
+  const version = semver.coerce(jestWrapper.getVersion());
   log.debug('Detected Jest version %s', version);
```
nicojs, line 19: Could we add a unit test for this? We already have a couple here: https://github.com/stryker-mutator/stryker-js/blob/renovate%2Ftypescript-5.x/packages/jest-runner/test/unit/jest-test-runner.spec.ts#L367

nicojs (changes requested):
Cool, I didn't know about `coerce`.

nicojs (approved):

nicojs:
I've added some unit tests and fixed the type errors.

Issue #5103, opened 2024-10-26
Script tags with type "module" are not being mutated


nicojs:
**Summary**

Script tags in html files with `type="module"` are not being mutated. See https://github.com/stryker-mutator/stryker-js/blob/feat%2Fmutation-testing-server/packages/instrumenter/src/parsers/html-parser.ts#L10

PR #5230, opened 2025-04-11
fix(instrumenter): Script tags with type "module" are not being mutated

bartekleon:
fixes: #5103


Andreas02-dev:
Looks good, thank you for the contribution!

PR #5243, opened 2025-04-25
fix(vitest-runner): Ensure no test failure screenshots from Vitest

### LINKED ISSUE #5242: Vitest tests with browser environment shouldn't screenshot on failure when mutating
**Summary**

Currently, the default configuration for testing in a browser environment using Vitest will take screenshots of the browser if the test fails. Since we mutate these tests and expect them to fail, we don't need to take screenshots of the failures.

Andreas02-dev:
When using browser environments in Vitest, browser screenshots on test failure are now enabled by default. This is unnecessary overhead when using StrykerJS, so this should be disabled instead.

Closes #5242

### DIFF
--- a/packages/vitest-runner/test/integration/browser-mode.it.spec.ts
+++ b/packages/vitest-runner/test/integration/browser-mode.it.spec.ts
@@ -175,9 +175,6 @@ describe('VitestRunner in browser mode', () => {
 
     // See issue https://github.com/stryker-mutator/stryker-js/issues/5242
     it("shouldn't take a screenshot on test failure", async () => {
-      // Arrange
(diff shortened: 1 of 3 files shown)


packages/vitest-runner/test/integration/browser-mode.it.spec.ts
```diff
@@ -171,5 +172,27 @@ describe('VitestRunner in browser mode', () => {
...
+          mutantActivation: 'static',
+          testFilter: [test3],
+        }),
+      );
+
+      // Assert
+      assertions.expectKilled(runResult);
+      expect(runResult.killedBy).deep.eq([test3]);
+      expect(runResult.failureMessage).contains('42 - 2 = undefined');
+      expect(fs.existsSync(screenshotPath)).to.be.false;
```
nicojs, line 195: can we use `fs.glob` instead?

PR #5460, opened 2025-08-02
chore: switch to `plugin-proposal-explicit-resource-management`

ericcornelissen:
Replace [`@babel/plugin-proposal-explicit-resource-management`](https://www.npmjs.com/package/@babel/plugin-proposal-explicit-resource-management) by [`@babel/plugin-transform-explicit-resource-management`](https://www.npmjs.com/package/@babel/plugin-transform-explicit-resource-management) as suggested by the deprecation warning of the former[^1]. This is a surface level replacement of the dependency (no related source code as adopted following this replacement nor the fact that the proposal has graduated). E.g., I noticed [`new-ts-features.ts`](https://github.com/stryker-mutator/stryker-js/blob/7270f111ff36d6b18669302f5702fd42f664d53e/packages/instrumenter/testResources/parser/new-ts-features.ts), where `using` should arguably no longer be considered a "new feature".

[^1]: "This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use `@babel/plugin-transform-explicit-resource-management` instead."

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -299,6 +299,15 @@
         "semver": "bin/semver.js"
       }
     },
+    "node_modules/@babel/helper-globals": {
+      "version": "7.28.0",
+      "resolved": "https://registry.npmjs.org/@babel/helper-globals/-/helper-globals-7.28.0.tgz",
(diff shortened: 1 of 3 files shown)


nicojs:
Thanks @ericcornelissen 💐

PR #5480, opened 2025-08-30
chore(vscode): enable vscode plugin

nicojs:


nicojs:
I found a way to enable the plugin. Might not be ideal, but it works.

PR #5498, opened 2025-09-26
feat(server): validate and filter fileRanges against mutate config or default patterns

### LINKED ISSUE #5496: Mutation server ignores mutate property in config
The current implementation of the StrykerJS mutation server overwrites the mutate property from the user’s Stryker configuration. This happens in both the discover and mutationTest methods. As a result, files are selected for mutation that:

- Cannot be parsed or mutated by StrykerJS (e.g., .css files).
- Are explicitly excluded by the user via the mutate property in their Stryker config.

This behavior leads to errors and failed mutation test runs, especially for users of the pre-released VSCode extension:

Image

This results in errors such as “No parser registered for .css” and premature abortion of the mutation test run.

**Expected Behavior**

- The StrykerJS server should validate whether the received fileRanges match the glob patterns from the project's Stryker configuration. If the user has specified a mutate property: filter against that. If no mutate property is provided: use StrykerJS’ default glob patterns to match. Any fileRanges not matching these patterns must be ignored.
- The client (e.g., the VSCode extension) must not pass folders as fileRanges if we use the prescribed solution. This prevents the server from trying to process invalid or ignored files within those folders. (to be discussed, this depends on where we implement the previously described matching logic)

The server should be responsible for filtering the fileRanges/handling them gracefully, since the mutation server protocol is intended to be framework-agnostic and should be able to send any file in the params. Otherwise, the client would need to parse Stryker configuration files (or any other supported framework's config).

jaspervdveen:
Resolves #5496

### DIFF
--- a/packages/core/src/fs/project-reader.ts
+++ b/packages/core/src/fs/project-reader.ts
@@ -119,25 +119,20 @@ export class ProjectReader {
       defaultOptions.mutate,
     );
 
-    // If targetMutatePatterns are provided, first filter the inputFileNames by those patterns.
-    // Otherwise, start with all inputFileNames as not mutated.
-    const mutateInputFileMap = new Map<string, FileDescription>();
+    const targetedInputFileNames = new Set<string>();
     if (targetMutatePatterns) {
       for (const pattern of targetMutatePatterns) {
         const files = this.filterMutatePattern(inputFileNames, pattern);
-        for (const [fileName, file] of files) {
-          mutateInputFileMap.set(
-            fileName,
-            this.mergeFileDescriptions(file, mutateInputFileMap.get(fileName)),
-          );
+        for (const [fileName, _] of files) {
+          targetedInputFileNames.add(fileName);
         }
       }
-    } else {
-      inputFileNames.forEach((fileName) =>
-        mutateInputFileMap.set(fileName, { mutate: false }),
-      );
     }
 
+    const mutateInputFileMap = new Map<string, FileDescription>();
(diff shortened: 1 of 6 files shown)


nicojs:
I've implemented the intersection of the mutate property with the target mutated files. Still todo:

- [ ] Unit test newly added utility methods (`filter` and `map`).
- [ ] Add more unit tests for ranges including columns

packages/core/src/fs/project-reader.ts
```diff
@@ -166,6 +211,52 @@ export class ProjectReader {
...
+            return {
+              start: { line: startLine, column: startColumn },
+              end: { line: endLine, column: endColumn },
+            };
+          }),
+        )
+        .filter(notEmpty);
+      return { mutate: intersectedRanges };
+    } else if (first.mutate === true && second.mutate === true) {
+      return { mutate: true };
```
nicojs, line 249: Unused branched

nicojs (changes requested):
One small issue remaining

packages/core/src/fs/project-reader.ts
```diff
@@ -166,6 +211,52 @@ export class ProjectReader {
...
+    } else if (first.mutate === true) {
```
nicojs, line 250:
```suggestion
    } else if (first.mutate === true) {
```

Unused branch

Andreas02-dev (changes requested):

Andreas02-dev (approved):
LGTM

PR #5510, opened 2025-10-04
fix(init): stryker init process should complete

### LINKED ISSUE #5504: `npm init stryker` hangs forever
**Summary**

Running `npm init stryker` hangs forever.

Image

Reproduce:

```bash
mkdir tmp
cd tmp
npm init --yes
npm init stryker
# ... fill in the questionnaire however you like
```

florisg-infosupport:
Resolves #5504

The process never completed due to a class that was not properly disposed of.
That caused users to previously have to manually exit the proces with ctrl+c.


nicojs:
This is awesome, thanks!

PR #5520, opened 2025-10-17
feat(server): support stdio transport channel

### LINKED ISSUE #5518: Support new MSP protocol (dual transport + static port)
The MSP spec now requires servers to support both stdio and socket transports, with a fixed port/address for socket.
Stryker-JS’s MSP server should be updated to match.

Tasks:
- Add serve <stdio|socket> CLI.
- stdio: read/write JSON-RPC over stdin/stdout, log only to stderr.
- socket: start on --port (required) and optional --address (default localhost).
- Exit with error if port is taken.
- should still support `runServer` command in the old fashion way, so there are no breaking changes.

Reference:
[https://github.com/stryker-mutator/editor-plugins/commit/1b82ac707b6c4756c5535b4a7d139d09b15858cc](https://github.com/stryker-mutator/editor-plugins/commit/1b82ac707b6c4756c5535b4a7d139d09b15858cc)

nicojs:
Support `stdio` process communication, adhering to mutation server protocol v0.4.'s [transport requirements](https://github.com/stryker-mutator/editor-plugins/tree/main/packages/mutation-server-protocol#transport-requirements).

- Add new `serve` command (see below).
- Add `stdio` as channel.
- Support configured port and address for `socket` channel, as per protocol
- Delete temp dir after a run (was a bug)
- Log to stderr when in server-mode (as per protocol)
- the old `runServer` command is deprecated, use `serve` instead

```
Usage: stryker serve [options] <channel> [-- <args>...]

Start the Stryker server

Arguments:
  channel             The channel to use for communicating with the Stryker server (choices: "stdio", "socket")
  -- <args>           Override Stryker options here, (see run command)

Options:
  --port <number>     The port to listen on (only for socket channel)
  --address <string>  The host to listen on (only for socket channel) (default: "localhost")
```

Closes #5518

### DIFF
--- a/packages/core/test/integration/stryker-server.it.spec.ts
+++ b/packages/core/test/integration/stryker-server.it.spec.ts
@@ -98,7 +98,7 @@ describe(StrykerServer.name, () => {
         configFilePath: 'stryker.conf.js',
       };
       const expectedResult: ConfigureResult = {
-        version: '1',
+        version: '0.4.0',
       };
 
       const result = await client.configure(configureParam);


packages/core/src/stryker-cli.ts
```diff
@@ -45,43 +31,126 @@ function parseCleanDirOption(val: string) {
...
+          const command = this.#provideStrykerOptions(
+            new Command().addArgument(configFileArgument),
+          ).parse(remainingArguments, { from: 'user' });
+          const runOptions = command.opts();
+          const [configFile] = command.args;
+          const strykerOptions = this.#readStrykerOptions(
+            runOptions,
+            configFile,
+          );
+          if (channel === 'socket') {
```
jaspervdveen: I noticed that the serve command's --port option is optional, but according to the mutation server protocol specification, the port should be required when using the socket channel. I'm fine with it either way, just checking if it's intentional.

packages/core/test/unit/stryker-cli.spec.ts
```diff
@@ -189,32 +192,142 @@ describe(StrykerCli.name, () => {
...
+      expect(helpText).matchSnapshot();
+    });
+
+    function actRunServer(...args: string[]): void {
+      act(['runServer', ...args]);
+    }
+  });
+
+  describe('serve', () => {
+    let consoleLogStub: sinon.SinonStubbedMember<typeof console.log>;
```
jaspervdveen: Could you add a test case that it should accept a config file as last argument?

jaspervdveen (commented):
Nice! I like this implementation and it looks good to me, though I would love to do some sanity checks with the editor plugin's implementation as well before we merge this.

packages/core/test/unit/stryker-cli.spec.ts
jaspervdveen: I removed this functionality since we don't need it for the server.

PR #5579, opened 2025-11-03
chore: improve list and duration formatting using Intl.ListFormat and Intl.NumberFormat

hugo-vrijswijk:
Uses [Intl.ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat) to format lists in error messages and logs, enhancing readability by using proper conjunctions. Also employs [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) for consistent number formatting in duration outputs.

### DIFF
--- a/packages/core/src/config/config-reader.ts
+++ b/packages/core/src/config/config-reader.ts
@@ -159,7 +159,7 @@ export class ConfigReader {
         if (namedExports.length === 0) {
           return "In fact, it didn't export anything.";
         } else {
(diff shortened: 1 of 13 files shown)


hugo-vrijswijk:
> Was this done with AI?

No, all artisanal and handcrafted code 👨‍🎨

PR #5713, opened 2026-01-04
feat(isolated scope): add --testFiles option to limit which tests are executing

mopi1402:
Add --testFiles option to limit which tests are executed with `testFiles` you can limit which test files are executed during mutation testing. When specified, only tests from these files will be run, allowing you to verify that a module's dedicated unit tests can kill all its mutants independently. Glob patterns are supported.

Here is the issue with all the explanations:
https://github.com/stryker-mutator/stryker-js/issues/5712

image


packages/vitest-runner/src/vitest-wrapper.ts
```diff
@@ -1,6 +1,25 @@
...
+    const require = createRequire(path.join(process.cwd(), 'package.json'));
+    const vitestPath = require.resolve('vitest/node');
+    const { createVitest } = await import(vitestPath);
+    return createVitest(...args);
+  } catch {
+    // Fall back to bundled vitest
+    return createVitestOriginal(...args);
+  }
+}
+
```
nicojs: Why is this `createVitestFromProject` needed for this PR? Please separate this out in a separate PR so we can discus it there.

mopi1402:
@nicojs , I've refactored the code, taken the feedback into account, and added end-to-end (E2E) tests for each of the runner tests.

Sorry, that's a lot of files (I'm more used to 10 to 20 max, but there is a lot of tests to cover).

The Karma implementation seems a bit more complex, so I've left it out.

Feel free to let me know if I need to make any further changes.

I've prepared the pull request for #5714, but I'm waiting for it to be finished because I wrote an E2E test for fastcheck/vitest with the --testFiles option.

PR #5726, opened 2026-01-13
Performance statistics in mutation report

SaketGupte:
Related to issue: [5635](https://github.com/stryker-mutator/stryker-js/issues/5635)

Add timer to stryker.ts to allow calculating of time taken for each phase.

Side-effect:
This would cause reportAll, wrapUp, and logDone to be called from Stryker.run() instead of in mutation-test-report-helper.ts


jaspervdveen (commented):
@SaketGupte Thanks for adding the performance statistics, I think it's a nice addition! I do have some concerns about the current implementation approach though.

Concerns with the current approach:

- `Stryker.run()` is starting to accumulate performance measurement logic, while its main responsibility should remain orchestration of the process flow.
- We already have a `Timer` service that’s injected across the pipeline and used by individual executors. For example, `DryRunExecutor` already marks performance boundaries via `this.timer.mark(INITIAL_TEST_RUN_MARKER)`. The current approach seems to duplicate that responsibility.

Suggested direction:

- Reuse the existing `Timer` pattern instead of centralizing all performance logic in `Stryker.run()`.
- Let each executor continue to mark its own performance boundaries using the injected `Timer`.
- Move the performance aggregation and calculation logic to a helper method in `MutationTestExecutor`, or introduce a dedicated performance statistics service.
- Replace magic strings for timer marks with constants to make the code easier to reason about and less error-prone.

I think this would keep responsibilities clearer and better align with the existing architecture.

PR #5745, opened 2026-01-22
fix(vitest): resolve dual instance problem with Vitest plugins

### LINKED ISSUE #5714: Dual Vitest Instance Causes Plugins (e.g., @fast-check/vitest) to Fail
**Summary**

Dual Vitest Instance Causes Plugins (e.g., @fast-check/vitest) to Fail

**Stryker config**

```json
{
  "testRunner": "vitest",
  "reporters": ["progress", "clear-text", "html"],
  "coverageAnalysis": "perTest"
}
```

**Test runner config**

```typescript
// vitest.config.ts
import { defineConfig } from 'vitest/config';

export default defineConfig({
  test: {
    // ...
  },
});
```

**Stryker environment**

```
+-- @stryker-mutator/core@9.4.0
+-- @stryker-mutator/vitest-runner@9.4.0
+-- @stryker-mutator/api@9.4.0
```

```
+-- vitest@4.0.13
+-- @fast-check/vitest@0.1.4
```

**Test runner environment**

```shell
vitest run
```

```typescript
// vitest.config.ts (standard configuration)
```

**Your Environment**

| ---------------- | -------
| node             | v20.0.0 (or higher)
| npm              | 10.x.x
| Operating System | Mac OS

**Add stryker.log**

When running Stryker with plugins like `@fast-check/vitest`, tests fail to kill mutants even when they should.

Specifically, `fast-check` relies on the global `vitest` instance. Stryker loads its own bundled `vitest` instance, creating a "dual instance" scenario. The user's tests interacts with the project's Vitest (imported in their code), while Stryker orchestrates the run using its own internal Vitest instance.

As a result, `fast-check` property executions are not correctly intercepted or reported back to Stryker's runner context, leading to surviving mutants or silent test failures.

Verification confirmed that ensuring Stryker loads the *project's* local `vitest` instance resolves the issue.

mopi1402:
## Problem

Vitest plugins like `@fast-check/vitest` register themselves globally on a Vitest instance. When Stryker uses its bundled Vitest instance while the project uses its own local instance, this creates a "dual instance" problem where:
- Plugins register on the project's Vitest instance
- Stryker orchestrates tests using its bundled Vitest instance
- Result: Plugins fail to work, tests are not discovered, and Stryker crashes with "No tests were executed"

## Solution

Load the project's local Vitest installation instead of Stryker's bundled one. This ensures:
- Single Vitest instance shared between plugins and Stryker
- Plugins work correctly during mutation testing
- Falls back to bundled Vitest if project doesn't have it installed

## Testing

Added e2e test `vitest-fastcheck` that:
- Uses `@fast-check/vitest` plugin to reproduce the dual instance problem
- Validates that property-based tests work correctly
- Achieves 100% mutation score (7/7 mutants killed)

Fixes #5714

### DIFF
--- a/packages/vitest-runner/src/vitest-wrapper.ts
+++ b/packages/vitest-runner/src/vitest-wrapper.ts
@@ -1,5 +1,5 @@
 import { createRequire } from 'module';
-import { pathToFileURL } from "node:url";
+import { pathToFileURL } from 'node:url';
 import path from 'path';
 import { createVitest as createVitestOriginal } from 'vitest/node';
 
@@ -22,7 +22,9 @@ async function createVitest(...args: Parameters<typeof createVitestOriginal>) {
     const require = createRequire(path.join(process.cwd(), 'package.json'));
     const vitestPath = require.resolve('vitest/node');
 
-    const { createVitest: createVitestProject } = await import(pathToFileURL(vitestPath).href);
+    const { createVitest: createVitestProject } = await import(
+      pathToFileURL(vitestPath).href
+    );
     return createVitestProject(...args);
   } catch {
     // Fallback to Stryker's bundled Vitest if project doesn't have it


Andreas02-dev (approved):
Thanks for the fix @mopi1402, looks good! I have simplified the Vitest project import a bit by using node's `pathToFileURL` function

Issue #5729, opened 2026-01-15
Error when using vitest-runner with Angular contentChild signal


vinylhero:
**Summary**

I'm experiencing an issue when I run Stryker with the vitest-runner when the mutator tries to modify a `contentChild` signal. This doesn't seem to be an issue with the `@ContentChild` decorator.

By telling stryker to avoid mutating the line below I can see that the run passes:

```ts
  // Stryker disable all
  readonly detail = contentChild(StepDetailDirective, { read: TemplateRef });
  // Stryker restore all
```

**Stryker config**

Below the "targetPath" is dynamic depending on the library under test (using NX).

```json
{    // Common settings
    packageManager: 'npm',
    reporters: ['progress', 'json'],
    testRunner: 'vitest',
    timeoutMS: 60000,
    coverageAnalysis: 'perTest',
    cleanTempDir: 'always',
    concurrency: 4,
    disableTypeChecks: false,
    ignoreStatic: true,
    ignorers: ['angular'],
    checkers: ['typescript'],
    
    // Incremental mode (speeds up subsequent runs)
    incremental: true,
    incrementalFile: `stryker-reports/${targetPath}-incremental.json`,

    // Dynamic settings based on the library path
    jsonReporter: {
      fileName: `stryker-reports/${targetPath}-mutations.json`,
    },
    vitest: {
      configFile: `${targetPath}/vite.config.mts`,
      related: false
    },
    tsconfigFile: `${targetPath}/tsconfig.stryker.json`,
    mutate: [
       ${targetPath}/**/*.ts`,
      `!${targetPath}/**/*.spec.ts`,
      `!${targetPath}/**/*.stories.ts`,
      `!${targetPath}/**/*.module.ts`,
      `!${targetPath}/**/index.ts`,
      `!${targetPath}/**/test-setup.ts`,
      `!${targetPath}/e2e-tests/**/*.ts`,
      `!${targetPath}/docs/**/*.ts`,
      `!${targetPath}/.storybook/**/*.ts`,
    ]
}
```

**Test runner config**

Vitest Config (vite.config.mts)

```json
{
  root: __dirname,
  cacheDir: '../../../../node_modules/.vite/libs/.../ui',
  plugins: [angular(), nxViteTsPaths(), nxCopyAssetsPlugin(['*.md'])],
  ssr: {
    noExternal: ['...'],
  },
  test: {
    name: getNxProjectName(),
    watch: false,
    globals: true,
    environment: 'jsdom',
[...]

PR #5767, opened 2026-01-30
feat(instrumenter): ignore mutations of the signal query options object

Andreas02-dev:
closes: #5729

### DIFF
--- a/packages/instrumenter/src/frameworks/angular-ignorer.ts
+++ b/packages/instrumenter/src/frameworks/angular-ignorer.ts
@@ -10,10 +10,10 @@ const ANGULAR_SIGNAL_QUERY_FUNCTIONS = Object.freeze([
 ]);
 
 const INPUT_MODEL_OUTPUT_CONFIG_MSG =
-  'Angular signal based input, model and output functions configuration object cannot be mutated as that causes issues with the ivy compiler.';
+  'Angular signal based input, model and output functions configuration object cannot be mutated as that causes issues with the Angular compiler.';
 
 const SIGNAL_QUERY_OPTIONS_MSG =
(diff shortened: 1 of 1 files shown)


packages/instrumenter/src/frameworks/angular-ignorer.ts
```diff
@@ -2,16 +2,38 @@ import { Ignorer, NodePath } from '@stryker-mutator/api/ignore';
...
- */
+const ANGULAR_SIGNAL_QUERY_FUNCTIONS = Object.freeze([
+  'contentChild',
+  'contentChildren',
+  'viewChild',
+  'viewChildren',
+]);
+
+const INPUT_MODEL_OUTPUT_CONFIG_MSG =
+  'Angular signal based input, model and output functions configuration object cannot be mutated as that causes issues with the ivy compiler.';
```
nicojs:
```suggestion
  'Angular signal-based input, model, and output functions configuration object cannot be mutated as that causes issues with the Angular compiler.';
```

packages/instrumenter/src/frameworks/angular-ignorer.ts
```diff
@@ -2,16 +2,38 @@ import { Ignorer, NodePath } from '@stryker-mutator/api/ignore';
...
+    }
+
+    return undefined;
+  }
+
+  #isClassFieldLike(path: NodePath): boolean {
+    return (
+      path.isClassProperty() === true ||
+      path.isClassPrivateProperty() === true ||
+      path.isClassAccessorProperty() === true
```
nicojs:
```suggestion
    return (
      path.isClassProperty() ||
      path.isClassPrivateProperty() ||
      path.isClassAccessorProperty() 
```

These functions return booleans, right?

PR #5986, opened 2026-04-24
feat(core): save partial incremental report on unexpected exit

### LINKED ISSUE #4886: Incremental file saved incrementally during execution rather than whole thing at the end
If a mutation testing run is interrupted prematurely the results up to that point are lost.

It would be nice to have an option to update the incremental file after X number of mutations (maybe not after every mutation passes/fails). That way you could still make progress towards completing a large increment even if you need to interrupt the process.

I have attempted to write a node function which will mutate a file at a time but I am having trouble getting it  to correctly update the files at the end of each run for some reason. Also, there is the overhead of startup for each file.

jaspervdveen:
Closes #4886

When running in incremental mode, pressing CTRL+C mid-run (or any other unexpected exit) loses all progress of the run. The next run has to redo the work from the exited run. This PR fixes that by writing a partial report to the incremental file before the process exits, so the next incremental run can pick up where the interrupted run left off.

This adds an async exit handler to `UnexpectedExitHandler` (via a new `registerAsyncHandler` method) and uses it in `MutationTestReportHelper` to write a partial report to the incremental file before the process exits. The save only happens when incremental mode is on, at least one mutant has been tested, and the normal report hasn't already been written. If a second signal arrives while the save is in progress, the process force-exits immediately.

The `writeIncrementalReport` method was extracted to avoid duplicating the file-write logic between the normal completion path and the interrupt path.

For testing, a custom signal reporter plugin triggers an interrupt on the first `onMutantTested` callback. The E2E test verifies that the process was actually interrupted and that the partial report was saved.


e2e/test/incremental-interrupt/verify/verify.js
Andreas02-dev, line 1: Would it be possible to expand the E2E test to 'resume' afterwards, and (if possible) verify that it completes with all tests without re-doing the work/rerunning the tests it already ran before the exit?

e2e/test/incremental-interrupt/stryker.conf.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "../../node_modules/@stryker-mutator/core/schema/stryker-schema.json",
+  "testRunner": "mocha",
+  "concurrency": 1,
+  "coverageAnalysis": "perTest",
+  "reporters": ["clear-text", "signal"],
+  "plugins": ["@stryker-mutator/mocha-runner", "./signal-reporter.js"],
```
Andreas02-dev, line 7: Does this mean that a user would have to enable this plugin manually? And if yes, is this something we should discuss adding to a stryker init?

Andreas02-dev (commented):
Looks good! Just added a comment about an improvement to the E2E test, and a question about the enabling of the feature, and if/how we should enable this by default

e2e/test/incremental-interrupt/verify/verify.js
jaspervdveen: The scope of this test is specifically to verify that the partial incremental report is saved when the process is unexpectedly interrupted. Verifying that a subsequent run correctly reads and uses the saved report would test the incremental report reader, which is already covered separately.

e2e/test/incremental-interrupt/stryker.conf.json
jaspervdveen: No, the signal-reporter is a test helper that only exists in this E2E test to programmatically trigger a SIGINT mid-run (to simulate an unexpected exit). End users never need to configure it. The actual feature works automatically when incremental: true is set, nothing extra required.

Andreas02-dev (approved):

e2e/test/incremental-interrupt/verify/verify.js
```diff
@@ -0,0 +1,85 @@
...
+
+const rootResolve = path.resolve.bind(
+  path,
+  fileURLToPath(new URL('..', import.meta.url)),
+);
+
+const incrementalFile = rootResolve('reports', 'stryker-incremental.json');
+
+describe('incremental interrupt', () => {
+  afterEach(async () => {
```
nicojs:
```suggestion
  beforeEach(async () => {
```

It's a bit safer. If a test run gets killed (`kill -9`), the next run still works.

e2e/test/incremental-interrupt/verify/verify.js
```diff
@@ -0,0 +1,85 @@
...
+    const report = JSON.parse(reportContent);
+
+    // Verify it contains at least some mutant results.
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+    const totalMutants = Object.values(report.files).reduce(
+      (sum, file) => sum + file.mutants.length,
+      0,
+    );
+    expect(totalMutants).to.be.greaterThan(0);
+  });
```
nicojs, line 165: Should we maybe also have a test that runs Stryker in a loop. Eventually all mutants should be tested. You might want a smaller example project for that .

packages/core/src/unexpected-exit-handler.ts
```diff
@@ -15,10 +19,33 @@ export class UnexpectedExitHandler implements Disposable {
     process.on('exit', this.handleExit);
     signals.forEach((signal) => process.on(signal, this.processSignal));
   }
+
   private readonly processSignal = (_signal: string, signalNumber: number) => {
-    // Just call 'exit' with correct exitCode.
-    // See https://nodejs.org/api/process.html#process_signal_events, we should exit with 128 + signal number
-    this.process.exit(128 + signalNumber);
```
nicojs, line 21: I'm pretty sure the code is necessary. I remember that Signal handling doesn't work on native Windows for the `ctrl+c` use case. The ONLY thing that works is _synchronous exit handlers_, which is why the `process.exit` approach was taken. We should also be able to verify this in the e2e test.

nicojs (changes requested):
Please verify that `CTRL+C` actually results in a SIGINT on native Windows. See my remark. I'm not so sure, I seem to remember it only triggered the `process.on('exit')` handler, which is why the synchronous code path existed.

EDIT:

That async handler isn’t going to work. I thought I remembered that SIGINT doesn’t work on Windows, but it’s subtly  different: it’s no longer possible to cancel terminating the process from the SIGINT handler.

EDIT:

I stand corrected:

afbeelding

Maybe this was the case? But clearly no longer.

packages/core/src/reporters/mutation-test-report-helper.ts
```diff
@@ -158,18 +179,28 @@ export class MutationTestReportHelper {
...
-        'utf-8',
-      );
+      await this.writeIncrementalReport(report);
     }
+    this.reportCompleted = true;
     this.determineExitCode(metrics);
   }
 
+  private async writeIncrementalReport(
+    resultsOrReport: readonly MutantResult[] | schema.MutationTestResult,
```
nicojs: Why this convoluted method signature? I would expect the mapping to `schema.MutationTestResult` to be handled in the exit handler.

packages/core/src/unexpected-exit-handler.ts
nicojs: Nope. I stand corrected

nicojs:
This implementation adds an async path to handlers. I think it would be better to convert all code to async handlers, including the other exit handlers. That way, the code can be simplified a lot.

It was synchronous because I was under the impression that it was the only way to handle exit logic in Windows. That is not (or no longer) the case.

@jaspervdveen, what do you think?

e2e/test/incremental-interrupt/verify/verify.js
jaspervdveen: Done!

e2e/test/incremental-interrupt/verify/verify.js
jaspervdveen: In the end, I decided to add an E2E test which validates that all mutants are eventually tested when Stryker runs in a loop with interrupts. I have added an assertion which checks that it reuses the mutant results from the previous iteration. 👍🏼

Andreas02-dev (approved):

PR #5993, opened 2026-04-28
fix(core): rebase incremental report paths across project roots

jvanes-awaze:
- This PR restores the incremental report path-rebasing improvement as a focused follow-up, split out from #5866.
- It addresses a real incremental-mode usability gap: reuse currently depends on report file keys matching the caller's current working directory.
- When an incremental report was produced from a different `projectRoot`, Stryker now rebases stored `files` and `testFiles` entries to the current working directory before diffing.
- This preserves existing behavior when `projectRoot` is unchanged or absent.
- Scope is intentionally narrow: only incremental report key rebasing plus focused docs/tests.
- Coverage includes a unit test proving report file names are remapped correctly when `projectRoot` differs from `cwd`.
- Net effect: incremental reuse stays stable across invocation contexts without changing default mutation behavior.

## Summary

This PR extracts the incremental report path-rebasing behavior into its own focused change.

Today, incremental reuse depends on the stored report keys lining up with the current invocation context. When an incremental report was generated from a different `projectRoot`, that alignment can break even when the underlying files are still the same, causing otherwise reusable history to be missed.

This PR fixes that by rebasing the incremental report's `files` and `testFiles` keys from the report's stored `projectRoot` to the current working directory before Stryker performs its diffing and reuse logic.

## Why this matters

Incremental mode is supposed to help users keep reuse stable between runs. In practice, invocation context can vary:

- running from a different working directory
- generating the prior report in another checkout location
- using tooling that changes the effective project root between runs

Without rebasing, those cases can make equivalent paths look different and reduce incremental reuse for no meaningful behavioral reason.

This PR closes that gap by normalizing the stored report paths to the current invocation context before reuse analysis starts.

### 1) Incremental report key rebasing
- `ProjectReader` now checks the incremental report's `projectRoot`.
- When `projectRoot` differs from `process.cwd()`, report keys are rebased to the current working directory.
- Rebasing is applied to:
  - `files`
  - `testFiles`

### 2) No-op when rebasing is unnecessary
- If `projectRoot` is absent, behavior stays unchanged.
- If `projectRoot === process.cwd()`, behavior stays unchanged.

### 3) Documentation update
- `docs/incremental.md` now explains that the incremental file stores `projectRoot` and that entries are rebased when the invocation context differs.

### 4) Focused unit coverage
- Added a unit test proving that report file names are remapped correctly when the report was produced from another project root.

### Benefits
- **More stable incremental reuse** across different invocation contexts.
- **Less surprising behavior** when equivalent reports are reused from another checkout/root.
- **Small, isolated scope** with no change to default non-incremental behavior.

### Costs / trade-offs
- **Slightly more preprocessing** when loading an incremental report.
- **Behavior change in incremental mode** for cross-root reuse scenarios, which is why this is split into its own PR.

## Compatibility and risk

- Backward compatibility: preserved for unchanged-root scenarios.
- Risk level: low to moderate; behavior only changes when loading incremental reports produced from a different `projectRoot`.
- Scope is intentionally limited to the incremental report loading path.

## Test coverage

Added focused coverage for:

- rebasing incremental `files` keys when `projectRoot` differs from `cwd`
- rebasing incremental `testFiles` keys when `projectRoot` differs from `cwd`
- preserving existing behavior for normal incremental reads

## Commits (high level)

- rebase incremental report keys to current working directory during report loading
- document the cross-root incremental behavior
- add focused unit coverage for rebased report paths

### DIFF
--- a/docs/incremental.md
+++ b/docs/incremental.md
@@ -21,6 +21,8 @@ _Setting `"incremental": true` in your stryker.config.json file is also supporte
 
 StrykerJS stores the previous result in a "reports/stryker-incremental.json" file (determined by the [--incrementalFile](./configuration.md#incrementalfile-string) option). The next time StrykerJS runs, it will read this JSON file and try to reuse as much of it as possible.
 
+The incremental file also stores the `projectRoot` of the run that produced it. When that root differs from your current working directory, StrykerJS rebases incremental file entries to the current directory before diffing, so reuse still works when invocation context changes.
+
 Reuse is possible when:
 
 - A mutant was "Killed"; the culprit test still exists, and it didn't change.
--- a/packages/core/src/fs/project-reader.ts
+++ b/packages/core/src/fs/project-reader.ts
@@ -15,6 +15,7 @@ import {
   ERROR_CODES,
   I,
   isErrnoException,
+  normalizeFileName,
   notEmpty,
 } from '@stryker-mutator/util';
 import type { MutationTestResult } from 'mutation-testing-report-schema/api';
@@ -418,7 +419,7 @@ export class ProjectReader {
     try {
       // TODO: Validate against the schema or stryker version?
       const contents = await this.fs.readFile(this.incrementalFile, 'utf-8');
-      const result: MutationTestResult = JSON.parse(contents);
+      const result = this.rebaseIncrementalReport(JSON.parse(contents));
       return {
         ...result,
         files: Object.fromEntries(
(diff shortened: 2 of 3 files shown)


d-es-ign:
Future changes will come from this account of mine as @jvanes-awaze will be deactivated in the near future, so include @d-es-ign for future questions

Andreas02-dev:
@d-es-ign Thanks for the PR!

When running this patch to verify the behavior, I noticed that the configuration itself won't be rebased when the invocation context changes.

Thinking about it more, there's also other configuration items that would then need to be rebased in order to keep the behavior consistent across invocation context changes, such as the ignorePatterns, mutate, and any other calls to process.cwd() (possibly even in plugins). To implement this correctly, the scope for this PR would therefore increase by a *lot*, and currently I don't think the feature is worth the refactor and added complexity to StrykerJS that a relative projectRoot would cause.

Without the rebasing of the projectRoot throughout Stryker and plugins (the current state of this PR), changing invocation context would cause files to stop being ignored, or other files to get included when this is not intended. Looking technically, this is expected, however I do not think that this is what a user would intend to happen when attempting to run Stryker incremental one or more directories above the repo root.

My apologies for the long wait and not seeing this issue earlier, but I think it's best to not go through with this feature at this time.

PR #6008, opened 2026-05-22
refactor: modernize codebase with new ECMAScript features

jaspervdveen:
Replace legacy patterns with modern ES2023-ES2025 equivalents now that the target has been upgraded to ES2025.


jaspervdveen:
Awaits typescript 6

PR #6009, opened 2026-05-22
chore(deps): upgrade to TypeScript 6

jaspervdveen:
This updates Stryker to work properly with TypeScript 6 and Node.js 22 by addressing TypeScript 6's breaking changes and the deprecation of ts-node.

TypeScript 6 changed how the rootDir compiler option works (now defaults to the current directory instead of the source directory), and ts-node is completely broken with TypeScript 6.

So I replaced ts-node with tsx in our tests, a modern drop-in replacement that handles TypeScript execution properly on Node 22 and works with TypeScript 6.


e2e/test/cucumber-ts/verify/verify.js.snap
```diff
@@ -4,19 +4,19 @@ exports[`After running stryker on a cucumber-ts project should report expected s
...
-  "mutationScore": 45.89041095890411,
-  "mutationScoreBasedOnCoveredCode": 62.616822429906534,
+  "killed": 65,
+  "mutationScore": 45.20547945205479,
+  "mutationScoreBasedOnCoveredCode": 61.6822429906542,
   "noCoverage": 39,
   "pending": 0,
   "runtimeErrors": 17,
-  "survived": 40,
+  "survived": 41,
```
Andreas02-dev, line 13: Do you know why one mutant now survives?

Andreas02-dev (approved):
Looks good, just one question since there's a mutant that now survives instead of getting killed, and it doesn't show up as compileTimeError like in the other snapshot

PR #6010, opened 2026-05-22
feat: mutant filtering

### LINKED ISSUE #4463: Feature: Mutant filters

We want to add a mutator that removes statements or expressions. I.e. #3916

```diff
-resource.dispose();
+ // 👽
```

However, this would result in many additional mutants without real value since another mutant already covered the statement. For example, we wouldn't want to remove this statement:

```diff
- const result = foo + bar;
+ const result = foo - bar; // 👽
```

We would like to be able to filter mutants using a `MutantFilter`, like is implemented in Stryker.NET: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/MutantFilters/IgnoreBlockMutantFilter.cs#L10

Note: we shouldn't mark filtered-out mutants as 'ignored'; they should not exist in your report.

A valid approach might be to add an optional `filter` method on the mutator API, that, when implemented, indicated whether or not the mutant is valid.

```diff
export interface NodeMutator {
  mutate(path: NodePath): Iterable<types.Node>;
+ filter?(mutableInScope: Mutable[]): boolean;
  readonly name: string;
}
```

Keep like it is now, with a lot of noise when implementing the statement mutator.

**Additional context**

0

Andreas02-dev:
closes: #4463


packages/instrumenter/.mocharc.cjs
```diff
@@ -1,9 +1,10 @@
 module.exports = {
-  "require": ["dist/test/setup.js"],
-  "spec": [
-    "dist/test/unit/**/*.js",
-    "dist/test/integration/**/*.js"
+  require: ['dist/test/setup.js'],
+  spec: [
+    'dist/test/unit/**/*.js',
+    'dist/test/integration/**/*.js',
+    'dist/test/perf/**/*.js',
```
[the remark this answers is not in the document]
Andreas02-dev: Not necessary, we assume that CI has the weakest machines and base our thresholds on that, with different thresholds for Windows runners and Linux runners, since Windows is a lot slower.

packages/instrumenter/src/transformers/babel-transformer.ts
```diff
@@ -90,13 +102,50 @@ export const transformBabel: AstTransformer<ScriptFormat> = (
...
+      }
+    }
+    if (!filteringMutators) {
+      return;
+    }
+    const mutantsInScope = mutantCollector.mutants.slice(subtreeStart);
+    for (const mutator of filteringMutators) {
+      if (!mutator.filter!(mutantsInScope)) {
+        const mutantsToFilterOut = mutantsInScope.filter(
+          (mutant) => mutant.mutatorName === mutator.name,
```
jaspervdveen:
```suggestion
          (mutant) => mutant.mutatorName === mutator.name && mutant.original === path.node,
```
I think only filtering by name is not enough, for example in case of a filter that says "only mutate if I'm the only one of my kind in scope":

```ts
filter(mutantsInScope) {
  return mutantsInScope.filter(m => m.mutatorName === this.name).length <= 1;
}
```
With a + b + c which translates to
BinaryExpression(BinaryExpression(a, +, b), +, c)
1. Inner a + b exits first, 1 arithmetic mutant in scope, filter passes, mutant placed in AST via replace.
2. Outer (a+b) + c exits, 2 arithmetic mutants in scope, filter returns false, code removes all arithmetic mutants from collector, including the inner one
Now stryMutAct("0") exists in the output code but mutant 0 is gone from the report.

packages/instrumenter/src/transformers/mutant-collector.ts
jaspervdveen: When remove() shrinks the internal array, the next collect() reuses an already-taken ID because it uses this._mutants.length (which is now smaller).

```ts
const total = a + b + c;
const doubled = x * 2;
```

During traversal
Step | Action | Array | New ID
-- | -- | -- | --
1 | collect on outer + | [m0] | "0" (length was 0)
2 | collect on inner + | [m0, m1] | "1" (length was 1)
3 | exit inner + -> filter passes, placed | [m0, m1] | stryMutAct("1") now in AST
4 | exit outer + -> filter rejects, removed | [m1] | m0 gone, length = 1
5 | collect on * | [m1, m2] | "1" (length was 1 again)

Now the generated code has two is_active__("1") calls at different locations. When Stryker activates mutant "1" to test the * operator, the + mutation also fires, so that's a problem.

So I think you need to use a simple counter here instead of array length.

packages/instrumenter/src/transformers/babel-transformer.ts
```diff
@@ -90,13 +102,50 @@ export const transformBabel: AstTransformer<ScriptFormat> = (
...
+        const mutator = mutatorsByName.get(mutant.mutatorName);
+        if (mutator?.filter) {
+          (filteringMutators ??= new Set()).add(mutator);
+        }
+      }
+    }
+    if (!filteringMutators) {
+      return;
+    }
+    const mutantsInScope = mutantCollector.mutants.slice(subtreeStart);
```
jaspervdveen:
```suggestion
    const mutantsInScope = mutantCollector.mutants.slice(subtreeStart).filter(m => !m.ignoreReason);
```
e.g. // Stryker disable
I think ignore mutants should be filtered out before being passed as in scope.

packages/instrumenter/src/transformers/babel-transformer.ts
```diff
@@ -90,13 +102,50 @@ export const transformBabel: AstTransformer<ScriptFormat> = (
...
+    for (const mutator of filteringMutators) {
+      if (!mutator.filter!(mutantsInScope)) {
+        const mutantsToFilterOut = mutantsInScope.filter(
```
jaspervdveen, line 142: This filter can now include mutants with ignoreReason set. These get removed from the collector entirely, so they disappear from the report (user never sees the "Ignored" status). I would fix this by filtering out ignored mutants.

packages/instrumenter/testResources/instrumenter/specific-no-mutants.ts.out.snap
```diff
@@ -1,72 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`instrumenter integration with mutation ranges should not make any mutations in a file not found in the specific mutants 1`] = `
```
jaspervdveen, line 3: I don't see changes in specific-no-mutants.ts, so how come this code has been removed in the snapshot?

packages/instrumenter/testResources/instrumenter/string-mutations.ts
```diff
@@ -1,7 +1,5 @@
-require('foo'); // no mutation
```
jaspervdveen, line 1: Is this removal on purpose?

packages/instrumenter/testResources/instrumenter/string-mutations.ts.out.snap
```diff
@@ -1,8 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`instrumenter integration should be able to instrument string literals in different places 1`] = `
-"require('foo'); // no mutation
-import 'foo'; // no mutation
```
jaspervdveen, line 5: Is this removal on purpose?

packages/instrumenter/testResources/instrumenter/ts-declarations.ts
```diff
@@ -1,6 +1,3 @@
 export declare const globalNamespace = "globalNamespace";
 declare function foo(): 'foo';
-declare module 'express' {
-}
-
-React.useState<false>();
```
jaspervdveen, line 6: Is this removal on purpose?

packages/instrumenter/src/transformers/mutant-collector.ts
jaspervdveen: Since the mutant collector is shared across all files this issue causes a lot of ID collisions.

jaspervdveen (changes requested):
Really cool feature! Took me a bit to wrap my head around the bottom-up exit-phase filtering, but once it clicks, the design is elegant.

I found a few issues that need fixing but other than that, I think it's ready.

packages/instrumenter/src/transformers/mutant-collector.ts
Andreas02-dev: Good find, I'll keep track of the ID and increment it.

packages/instrumenter/src/transformers/babel-transformer.ts
Andreas02-dev: Thanks for the deep look, makes sense, I'll change it.

packages/instrumenter/src/transformers/babel-transformer.ts
Andreas02-dev: Agreed

packages/instrumenter/src/transformers/babel-transformer.ts
Andreas02-dev: Good find, I'll add a check

packages/instrumenter/testResources/instrumenter/specific-no-mutants.ts.out.snap
Andreas02-dev: This snapshot doesn't exist anymore, my assumption is that the snapshot state changed and it never picked up this snapshot is unused until I deleted the file and it re-generated it without the snapshot

packages/instrumenter/testResources/instrumenter/string-mutations.ts
Andreas02-dev: I believe it was since require here did something unintentional with the empty-expression-mutator, but I'll revert this and introduce it there if it's still require-d.

packages/instrumenter/testResources/instrumenter/string-mutations.ts.out.snap
Andreas02-dev: I believe it was since require here did something unintentional with the empty-expression-mutator, but I'll revert this and introduce it there if it's still require-d.

packages/instrumenter/testResources/instrumenter/ts-declarations.ts
Andreas02-dev: Double-checked, this shouldn't have been removed, thanks!

jaspervdveen (approved):
LGTM

PR #6012, opened 2026-05-22
feat: empty-expression-mutator

### LINKED ISSUE #5765: Empty expression mutator
```ts
// OK to mutate to empty statement
foo();
console.log();
foo(bar);
throw new Error();

// This is an empty statement:
;

// Not needed
console.log('test');
function add(a, b){
  if(a < 0) {
    return;
  }
  const result = a + b;
  return result;
}
```

Andreas02-dev:
closes: #5765


e2e/test/svelte-jest/verify/verify.js.snap
```diff
@@ -3,7 +3,7 @@
 exports[`Verify stryker has ran correctly should report correct score 1`] = `
 Object {
   "compileErrors": 0,
-  "ignored": 0,
+  "ignored": 2,
```
jaspervdveen: As far as I can see in `App.svelte` the arguments are not mutated in the `sum` and `isOldEnough` function calls so I don't really understand why the expressions are ignored. Is this correct?

jaspervdveen: I see the same thing in other Svelte E2E tests as well.

jaspervdveen: Never mind, I see they are explicitly ignored in the Svelte transformer because Svelte template expressions don't support statements. 👍🏼

packages/instrumenter/src/transformers/svelte-transformer.ts
```diff
@@ -19,10 +19,17 @@ export const transformSvelte: AstTransformer<AstFormat.Svelte> = (
   [root.moduleScript, ...root.additionalScripts]
     .filter(notEmpty)
     .forEach((script) => {
+      // Template expressions are expression-only contexts. Excluding the
+      // statement-producing CallExpression mutator prevents invalid Svelte
+      // output (for example `if (...) {}` inside `{ ... }`).
+      const excludedMutations = script.isExpression
```
jaspervdveen: Is `void 0` or `undefined` an option here as a replacement for call expressions in Svelte templates?

jaspervdveen (commented):
Very nice work @Andreas02-dev! 💯 
I just have one question about ignoring the call expression mutations in Svelte templates, but other than that it looks good to me!

packages/instrumenter/src/transformers/svelte-transformer.ts
Andreas02-dev: Yes, I deliberately omitted that from this PR so it could be finished earlier and we could improve on it later, but after your comment I have decided to add it now.

nicojs (approved):
LGTM

PR #6021, opened 2026-06-03
feat(mocha-runner): support Mocha v12

lo1tuma:
Mocha 12 migrated its CLI to ES modules, so the runner's synchronous `require('mocha/lib/cli/options')` no longer resolves. `lib-wrapper` now falls back to a dynamic `import()` of `lib/cli/options.mjs` when the require fails, leaving the existing path intact for Mocha < 12, and the peer dependency range is widened to `>= 7.2 < 13`.

Verified by running the integration suite against `mocha@12.0.0-beta-9.4` (note: a prerelease won't satisfy the range under strict semver, but the upcoming `12.0.0` stable will).

### DIFF
--- a/packages/mocha-runner/package.json
+++ b/packages/mocha-runner/package.json
@@ -54,10 +54,11 @@
   },
   "devDependencies": {
     "@stryker-mutator/test-helpers": "workspace:*",
-    "@types/node": "24.13.2"
+    "@types/node": "24.13.2",
+    "mocha": "11.7.5"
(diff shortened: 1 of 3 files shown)


nicojs:
Thanks, @lo1tuma. It seems a new Mocha release is imminent. Happy to merge when this is green.

lo1tuma:
Thanks for the reply @nicojs. I’ve rebased this PR.

PR #6024, opened 2026-06-05
fix(instrumenter): parse svelte template expressions as TS when file uses lang="ts"

### LINKED ISSUE #6003: BABEL_PARSE_ERROR for TypeScript in Svelte expressions
Svelte 5 supports TypeScript syntax in template expressions such as the following (this is just a minimal example, a real app probably uses e.g. `?` or `!` for optional properties).

```html
<script lang="ts"></script>

{"test" as string}
```

If you use such TS syntax in an expression `npx stryker run` fails with:

```
SyntaxError: .../Foo.svelte: Missing semicolon. (1:6)

> 1 | "test" as string
    |       ^
(omitted stacktrace)
{
  code: 'BABEL_PARSE_ERROR',
  reasonCode: 'MissingSemicolon',
  loc: Position { line: 1, column: 12, index: 12 },
  pos: 12,
  syntaxPlugin: undefined
}
```

[Reproduction repository](https://github.com/not-my-profile/repro-stryker-svelte)

**Stryker environment**
```
├── @stryker-mutator/core@9.6.1
├── @stryker-mutator/vitest-runner@9.6.1
└── vitest@4.1.6
```

jaspervdveen:
Template expressions like `{user!.name}` in Svelte files with `lang="ts"` were parsed as JavaScript, causing a Babel parse error on TypeScript-specific syntax.

Fix: derive the expression format from whether any script tag in the file has `lang="ts"`, matching how Svelte itself treats TypeScript as file-wide.

Adds a unit test and an e2e test (svelte-ts-vitest) reproducing the issue.


Andreas02-dev (approved):
LGTM!

PR #6099, opened 2026-07-03
feat(typescript-checker): add experimental support for TypeScript@7

### LINKED ISSUE #6070: Experimental support for TS7
## Problem

On a small TypeScript project, `@stryker-mutator/typescript-checker` spent most of the run in the JS `tsc` watch-builder checker processes. The same 238-mutant / 5-file run went from 54s to 15s when the checker was backed by tsgo (native TypeScript), with about 40s of the original run spent in the 5 TypeScript checker processes.

## Approach

The prototype keeps a resident tsgo process via `@typescript/native-preview/unstable/sync` instead of starting or maintaining JS `tsc` checker workers. It gives tsgo an in-memory FS overlay, applies one mutant at a time, calls incremental `updateSnapshot({ fileChanges })` (about 2-5ms in the measured project), and reads structured semantic diagnostics.

TypeScript diagnostics with category `Error` map directly to `CheckStatus.CompileError`. The checker groups one mutant per group so every whole-program diagnostic after the overlay update is attributable to that mutant.

## Correctness

For a checker, speed is only useful if compile-error classification matches `tsc`. tsgo is still a preview and may miss a diagnostic that `tsc` reports, or add one `tsc` does not. Either mismatch can silently skew mutation scores.

The reference package ships a parity gate:

```sh
npx stryker-tsgo-parity
```

It runs the same Stryker config twice, once with `checkers: ["typescript"]` and once with `checkers: ["tsgo"]`, then compares `CompileError` classification by mutant location/mutator/replacement instead of report IDs. On the measured real project, both runs produced the same killed/survived/CompileError counts, including 54 matching `CompileError` mutants.

## Caveat and ask

`@typescript/native-preview` is pre-GA and the API is under `unstable/*`, so I would gate this behind an opt-in flag or experimental checker until tsgo stabilizes.

The reference implementation is CC0/public domain, specifically so the Stryker team can copy or absorb it wholesale without taking a dependency or carrying a license obligation:

- npm: https://www.npmjs.com/package/stryker-tsgo-checker
- source: https://github.com/audreyt/stryker-tsgo-checker

nicojs:
Add experimental support for TypeScript 7 (a.k.a. tsgo or TypeScript native preview) to @stryker-mutator/typescript-checker.

Enable it in stryker.config.json:

```json
{
  "typescriptChecker": {
    "experimentalNativePreview": true
  }
}
```

It currently _assumes typescript v7 is installed as `@typescript/native`_, you can accomplish this with:

```bash
npm install -D @typescript/native@npm:typescript@latest
```

Closes #6070

## Performance improvements

Preliminary results show more than a 2x performance improvement.

## Limitations

1. This implementation uses the `typescript/unstable/*` api, which might be unstable.
2. Project references are not supported, since we currently seem to be missing a way to load tsgo in `--build` mode.
3. Grouping mutants into smart groups that can be checked together is currently not supported. We're missing the `program.getAllDependencies` api.
4. Unfortunately, typescript@6 still needs to be installed under the "normal" typescript name. This is needed for some utility functions we currently have no replacement for in the experimental TS7 api (i.e. `ts.parseConfigFileTextToJson`)


packages/typescript-checker/src/ts-native/native-diagnostic.ts
```diff
@@ -0,0 +1,71 @@
...
+import path from 'path';
+
+import type { Diagnostic, DiagnosticCategory } from 'typescript7/unstable/sync';
+import { PositionConverter } from './position-converter.js';
+import { toPosixFileName } from '../tsconfig-helpers.js';
+/**
+ * A located version of a diagnostic reported by the native TypeScript compiler (TypeScript 7).
+ * Locations are resolved eagerly, because diagnostic positions refer to in-memory (mutated)
+ * content that is reset after each check.
+ */
```
Andreas02-dev, line 14: For my understanding, does this mean that this will have to change when build mode (--build) is used?

packages/typescript-checker/test/integration/ts-v7/single-project.it.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+    ]);
+
+    // Assert
+    const expectedResult: Record<string, CheckResult> = {
+      id42: { status: CheckStatus.Passed },
+    };
+    expect(result).deep.eq(expectedResult);
+  });
+
+  it('should be allow mutations in unrelated files', async () => {
```
Andreas02-dev:
```suggestion
  it('should allow mutations in unrelated files', async () => {
```

packages/typescript-checker/test/integration/ts-v7/single-project.it.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+      'id45',
+    );
+    const expectedResult: Record<string, CheckResult> = {
+      id45: { status: CheckStatus.Passed },
+    };
+    const actual = await sut.check([mutant]);
+    expect(actual).deep.eq(expectedResult);
+  });
+
+  it('should be able invalidate 2 mutants that do result in a compile error', async () => {
```
Andreas02-dev:
```suggestion
  it('should be able to invalidate 2 mutants that do result in a compile error', async () => {
```

packages/typescript-checker/test/integration/ts-v7/single-project.it.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+      'mutId2',
+    );
+    const actual = await sut.check([mutant, mutant2]);
+    assertions.expectCompileError(actual.mutId);
+    assertions.expectCompileError(actual.mutId2);
+    expect(actual.mutId.reason).has.string('todo.ts(15,9): error TS2322');
+    expect(actual.mutId2.reason).has.string('counter.ts(7,5): error TS2322');
+  });
+
+  it('should group mutants outside the project together and check project mutants one-by-one', async () => {
```
Andreas02-dev: Doesn't this test need an additional mutant outside of the project to verify the expected result? Right now, there's only one mutant outside of the project so the test doesn't validate that if there's more than one, it should be grouped together.

Andreas02-dev (changes requested):
Looks good, just have one question for my understanding and a test that doesn't validate the expected behavior correctly.

What do you think of using `tsgo-*` instead of `native-*` for the names? I think that makes more sense, so for example: `tsgo-file-system` instead of `native-file-system`, etc.

packages/typescript-checker/src/ts-native/native-diagnostic.ts
nicojs: No, the code needs to be revised when we start applying more than 1 mutant in the source code at once. If you add a mutant on line 3, it might shift all other mutants in the same file by 1 line.  We will either start correcting for that or remove the line/column altogether.

f.y.i. this comment was generated by Claude

Andreas02-dev: Thanks, that makes sense.

packages/typescript-checker/test/integration/ts-v7/single-project.it.spec.ts
nicojs: True. Will add an additional test.

nicojs:
@Andreas02-dev  I've addressed all your remarks. Feel free to merge and release a version next Friday.

PR #6146, opened 2026-07-26
fix(vitest-runner): never report a mutant run that executed zero tests as survived

### LINKED ISSUE #6073: vitest-runner: non-deterministic mutant verdicts under coverageAnalysis=perTest (9.6.1, vitest 4.1.9)
Follow-up to #5928 (closed) at @Andreas02-dev's request, you reproduced the non-determinism on the [tsgit](https://github.com/scolladon/tsgit) repo and asked for a dedicated issue with a minimal reproduction. Here it is.

## Summary

Under `coverageAnalysis: "perTest"`, the vitest-runner produces **non-deterministic mutant verdicts** across identical runs. The same mutant flips `Timeout` ↔ `Survived` and `Killed` ↔ `Survived`, and the global status counts swing run-to-run. No test uses `describe.concurrent` / `it.concurrent` anywhere in the suite (Stryker sets vitest concurrency to 1 regardless).

This matches your own two-run observation on tsgit: `1 survived / 0 timeout` vs `0 survived / 1 timeout` on the same mutant.

## Minimal reproduction

https://github.com/scolladon/stryker-pertest-flip-repro

```bash
npm install
npm run flip        # runs `stryker run` 6x and tabulates each mutant's status per run
```

`run-many.mjs` prints, for each run, the global status counts plus the status of every mutant in `src/hot.ts`. On a correct runner the table is **identical every run**; here it changes between runs and prints `>>> FLIP DETECTED`.

How it is built:
- `src/hot.ts` is imported by **many** unit-test files, so its statements carry heavy per-test coverage — where `perTest` pairing mis-maps most often.
- `drain()` uses a `while (i < limit && active)` guard. The `LogicalOperator` mutant `&&` → `||` loops forever (`active` is always `true`), so a correctly **paired** covering test **times out** (a kill). When `perTest` mis-pairs, that covering test is never run against the mutant and it is reported **`Survived`** — the `Timeout` ↔ `Survived` flip on a single mutant.

## Observed over 6 identical runs

Global counts: `Survived` 5→8, `Timeout` 4→5, `Killed` 45→47. Example per-mutant flips in `src/hot.ts`:

| mutant | most runs | flipped run |
| --- | --- | --- |
| `AssignmentOperator` l25:5 (loop body) | `Timeout` | `Survived` |
| `BlockStatement` l24:31 (loop body) | `Timeout` | `Survived` |
| `ConditionalExpression` l24:10 (loop guard) | `Killed`/`Timeout` | `Survived` |
| `ConditionalExpression` l6:7 (`classify` guard) | `Killed` | `Survived` |
| `EqualityOperator` l9:19 (`classify` guard) | `Killed` | `Survived` |

A hand-applied check confirms these are genuine false survivors: applying the `classify` `&&` → `||` mutant by hand and running its covering tests directly (`npx vitest run test/unit/classify-*.test.ts`) **fails** (`expected 'small' to be 'out-of-range'`), proving a killer test exists for a mutant Stryker reports as `Survived`.

## Environment

| software | version |
| --- | --- |
| @stryker-mutator/core | 9.6.1 |
| @stryker-mutator/vitest-runner | 9.6.1 |
| @stryker-mutator/typescript-checker | 9.6.1 |
| vitest | 4.1.9 |
| @vitest/coverage-v8 | 4.1.9 |
| typescript | 6.0.3 |
| node | v22.22.3 |
| OS | macOS (Darwin 25.5.0) |

Stryker config: `coverageAnalysis: "perTest"`, `concurrency: "100%"`, vitest `dir: "test/unit"`, typescript checker enabled, `timeoutMS: 60000`, `timeoutFactor: 2`, `ignoreStatic: true`.

scolladon:
Fixes #6073

Under load, a filtered mutant run can complete with its test tasks collected but never executed. `ctx.state` then holds result-less tasks, `.filter((test) => test.result)` empties the result list, and `toMutantRunResult` reports **survived** with `hitCount 0` — even though no test ever ran against the mutant. This is the source of the non-deterministic `Survived` ↔ `Killed`/`Timeout` verdict flips in #6073 (full instrumented evidence in [this comment](https://github.com/stryker-mutator/stryker-js/issues/6073#issuecomment-5084446638)): every unstable verdict is a zero-test run, and verdicts with at least one executed test are stable.

The run loss itself is timing-dependent (clustered after bail-aborted runs). I could not reproduce it outside real Stryker orchestration (~1,400 instrumented pure-vitest runs mimicking the runner exactly are clean), but the runner should not trust an empty run regardless of the trigger: a run that executed none of its selected tests proves nothing about the mutant.

### Change

In `mutantRun`, when `testFilter` is non-empty and the run completed with zero executed tests:

1. retry the run (bounded, 3 attempts) — re-dispatching succeeds in practice;
2. if it still comes back empty, return an **error** run result stating the invariant, so core reschedules/reports it instead of recording a phantom survivor.

Unfiltered runs (no `testFilter`) are unaffected.

### Verification

Against [stryker-pertest-flip-repro](https://github.com/scolladon/stryker-pertest-flip-repro) (vitest 4.1.9 and 4.1.10):

- unpatched: `>>> FLIP DETECTED` on every 5-run batch; survivors swing 5–11;
- patched: 5/5 byte-identical reports (`47 Killed / 5 Survived / 5 Timeout`); 16 lost runs absorbed by retries (max 3 for one mutant); genuine survivors unchanged.

Open question: if you'd rather surface exhaustion as a different status than an error result (e.g. treat it like a timeout so core retries in a fresh process), happy to adjust.

### DIFF
--- a/packages/vitest-runner/src/vitest-test-runner.ts
+++ b/packages/vitest-runner/src/vitest-test-runner.ts
@@ -50,6 +50,12 @@ type StrykerNamespace = '__stryker__' | '__stryker2__';
 const STRYKER_SETUP = fileURLToPath(
   new URL('./stryker-setup.js', import.meta.url),
 );
+/**
+ * Maximum number of times a mutant run is re-issued when vitest completed it
+ * without executing any of the tests selected by the test filter. See
+ * https://github.com/stryker-mutator/stryker-js/issues/6073.
+ */
+const MAX_EMPTY_FILTERED_RUN_RETRIES = 3;
 
 interface RunFilter {
   /**
@@ -207,13 +213,49 @@ export class VitestTestRunner implements TestRunner {
     this.ctx!.provide('hitLimit', options.hitLimit);
     this.ctx!.provide('mutantActivation', options.mutantActivation);
     this.ctx!.provide('activeMutant', options.activeMutant.id);
-    const dryRunResult = await this.run({
+    const runFilter: RunFilter = {
       testIds: options.testFilter,
       relatedFiles: [options.sandboxFileName],
-    });
+    };
+    let dryRunResult = await this.run(runFilter);
+    for (
+      let retry = 1;
+      this.isEmptyFilteredRun(options.testFilter, dryRunResult) &&
+      retry <= MAX_EMPTY_FILTERED_RUN_RETRIES;
+      retry++
+    ) {
+      this.log.debug(
+        'Mutant run for mutant %s completed without executing any of its %s filtered tests, retrying (%s/%s).',
+        options.activeMutant.id,
+        options.testFilter!.length,
+        retry,
+        MAX_EMPTY_FILTERED_RUN_RETRIES,
+      );
+      dryRunResult = await this.run(runFilter);
+    }
+    if (this.isEmptyFilteredRun(options.testFilter, dryRunResult)) {
+      // A run that executed none of its selected tests proves nothing about
+      // the mutant; reporting it based on "no test failed" would produce a
+      // false survivor (see https://github.com/stryker-mutator/stryker-js/issues/6073).
+      return toMutantRunResult({
+        status: DryRunStatus.Error,
+        errorMessage: `Vitest completed the mutant run without executing any of the ${options.testFilter!.length} tests selected for mutant ${options.activeMutant.id} (after ${MAX_EMPTY_FILTERED_RUN_RETRIES} retries).`,
+      });
+    }
     const hitCount = this.readHitCount();
     const timeOut = determineHitLimitReached(hitCount, options.hitLimit);
     return toMutantRunResult(timeOut ?? dryRunResult);
+  }
+
+  private isEmptyFilteredRun(
+    testFilter: string[] | undefined,
+    result: DryRunResult,
+  ): boolean {
+    return (
+      (testFilter?.length ?? 0) > 0 &&
+      result.status === DryRunStatus.Complete &&
+      result.tests.length === 0
+    );
   }
 
   private async run({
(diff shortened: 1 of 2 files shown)


scolladon:
Follow-up: I've now pinned the exact mechanism with probes inside vitest's main chunk, and it changes the framing slightly — the lost runs are **bail's own cancellation racing the result flush**. Stryker passes `bail: 1`; on the first failing test the worker fires `rpc().onCancel("test-failure")` immediately, while that failing test's result is still in the throttled task-update batch. The main process handles the cancel → `pool.cancel()` force-stops the worker before the batch is applied, and `state.cancelFiles` registers bare result-less tasks for the remaining files — so the run resolves "complete" with zero results, and the runner scores it survived. **The flipping mutants were actually being killed** — bail destroyed the evidence of its own trigger.

Three confirmations:

- probes at `executeTests` / `cancelCurrentRun` / the worker→main `onCancel` RPC: 13/13 zero-test verdicts show the in-window `onCancel("test-failure")` → silent-cancel trace (files parked via `state.cancelFiles`, run resolves cleanly);
- this PR's retry guard eliminates the flip (5/5 byte-identical reports on the repro);
- `disableBail: true` also eliminates it (5/5) — so users hitting this today have a config-level workaround, at the cost of running all covering tests per mutant.

Given the root cause, you may prefer a different fix than the retry — e.g. not enabling vitest's bail and implementing bail runner-side (cancel after reading results), or detecting the interrupted run via the reporter lifecycle. The retry guard stands as defense-in-depth either way (an empty filtered run should never be scored survived regardless of the cause). Happy to rework in whichever direction you prefer.

PR #6188, opened 2026-08-25
feat(vitest-runner): report test locations

### LINKED ISSUE #6186: vitest-runner: report test locations so incremental mode gets "Full" test reporting
The [incremental docs](https://stryker-mutator.io/docs/stryker-js/incremental/) grid rates the Vitest runner as **"Tests per file without location"**, one level below Jest's **"Full"**. The consequence is on the same page: Stryker assumes every test inside a test file changed when that file changed, so one edit to one test re-runs all mutants that any test in that file killed.

Vitest can supply the missing data. It has done so since v1.4 through the `includeTaskLocation` option, and the runner's peer range is already `vitest >=2.0.0`. The gap is on the Stryker side: `@stryker-mutator/vitest-runner` neither enables the option nor maps the resulting `location` onto `TestResult.startPosition`.

**Additional context**
I patched this locally on a real project. One edited test went from **29 of 31 mutants re-run** to **2 of 31**.

**Proposed change**

1. Add `includeTaskLocation: true` to the `createVitest` overrides.
2. Map the location in `convertTestToTestResult`. Vitest's `task.location` is 1-based on    `line`, the same convention as Jest, so the same `- 1` applies:

```ts
const baseTestResult: BaseTestResult = {
  id: normalizeTestId(toRawTestId(test)),
  name: collectTestName(test),
  timeSpentMs: test.result?.duration ?? 0,
  fileName: test.file?.filepath && path.resolve(test.file.filepath),
  startPosition: test.location
    ? { line: test.location.line - 1, column: test.location.column }
    : undefined,
};
```

christopher-buss:
Enable vitest's `includeTaskLocation` and translate the reported task location into the `startPosition` that the test runner API already carries, and that the jest-runner and cucumber-runner report today.

Incremental mode uses the test locations to diff test files. Without them, Stryker has to assume that every test in a changed test file changed, so it reruns every mutant covered by that file. With them, only the mutants covered by the tests that actually changed are rerun. This moves the vitest runner from "tests per file without location" to "full" test reporting.

Stryker stores 0-based lines and columns internally. Vitest builds the location from a V8 stack trace, so both its line and its column are 1-based, and both are decremented. The jest-runner decrements the line only, because jest takes its columns from babel, where they are already 0-based.

Closes #6186.

AI Attribution:
Assisted via Claude Code but reviewed and filed by myself.

### DIFF
--- a/docs/incremental.md
+++ b/docs/incremental.md
@@ -70,7 +70,7 @@ Running in incremental mode, Stryker will do its best to produce an accurate mut
 | 🔵 Karma           | ⚠ Test names only                 |
 | 🥒 CucumberJS      | ✅ Full                           |
 | 📼 Tap             | ⚠ Tests per file without location |
-| ⚡ Vitest          | ⚠ Tests per file without location |
+| ⚡ Vitest          | ✅ Full                           |
 | ▶ Command          | ❌ Nothing                        |
 
 You can use this table to understand why StrykerJS decides not to rerun a specific mutant even though you've changed tests covering that mutant.
--- a/docs/vitest-runner.md
+++ b/docs/vitest-runner.md
@@ -62,6 +62,7 @@ The following options will be set by Stryker and cannot be overridden:
 {
   threads: true,
   coverage: { enabled: false },
+  includeTaskLocation: true,
   singleThread: true,
   watch: false,
   bail: options.disableBail ? 0 : 1,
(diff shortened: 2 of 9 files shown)


Andreas02-dev (approved):
Great improvement, thank you for your contribution @christopher-buss!

PR #6205, opened 2026-08-31
fix(mocha-runner): load Mocha cli internals regardless of file extension

lo1tuma:
Mocha 12 turned its package into an ES module and renamed the cli internals the runner reaches into to `.cjs`, so the extensionless `require` calls no longer resolve and importing the plugin throws before it can register itself.

- `lib/cli/run-helpers`, `lib/cli/options` and `lib/cli/collect-files` now go through one lookup that probes `.js`, `.cjs` and `.mjs`. It replaces the `options.mjs` fallback from #6021, which only ever matched the withdrawn Mocha 12 betas.
- `e2e/test/mocha-next-version` mirrors `mocha-old-version` (Mocha 7.2) against Mocha 12, so the newest supported major is covered too.
- The `">= 7.2 < 13"` peer range already covers Mocha 12, so it stays as is.


e2e/test/mocha-next-version/src/add.js
```diff
@@ -0,0 +1,6 @@
+module.exports.add = function(num1, num2) {
+  return num1 + num2;
+};
+
+
+
```
ericcornelissen, line 6: minor but I don't see the point of these stray trialing newlines

```suggestion
```

e2e/test/mocha-next-version/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "scripts": {
+    "clean:reports": "rimraf \"reports\"",
+    "verify": "mocha --no-config --no-package --timeout 0 verify/verify.js",
+    "test:unit": "mocha",
+    "test": "npm run clean:reports && stryker run && npm run verify"
+  },
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "mocha": "12.0.0-rc.6",
```
ericcornelissen, line 16: should this be just v12 now that it has been released?

```suggestion
    "mocha": "12.0.0",
```

ericcornelissen (approved):
I tried these changes against https://github.com/ericcornelissen/eslint-plugin-top/pull/2044 locally and it worked as expected.

I went over the changes and they looked good, left some minor comments below. I'm not sure if iterating over file extensions is much better than the previous `try`-`catch` based resolution in `resolveLoadOptions` but both seem valid to me.

Note that I'm just an outsider to the project trying to help this along since I would like to upgrade to Mocha v12 and keep using Stryker, great work on this @lo1tuma 🙂

e2e/test/mocha-next-version/src/add.js
lo1tuma: Removed. They came along when I copied the fixture from mocha-old-version.

e2e/test/mocha-next-version/package.json
lo1tuma: Bumped to 12.0.0. That also settles the peer range question: the final release satisfies ">= 7.2 < 13" without the prerelease exception, so I dropped that note.

PR #6220, opened 2026-09-09
fix(vitest-runner): match Vitest 5 full test names

seandearnaley:
Vitest 5 matches full test names with ` > ` between suites and tests. The runner constructed both reported test IDs and sandbox coverage IDs using spaces, so a successful dry run was followed by filtered mutant runs that executed no covered tests. Supply one version-aware separator to both paths while preserving older Vitest behavior.

Adds an integration test that round-trips a dry-run test ID into a killing mutant run, plus an automatically discovered Vitest 5 E2E fixture covering nested suites and regex punctuation. The fixture checks killed mutants and the corresponding reported test IDs.

### DIFF
--- a/e2e/test/vitest-5/package.json
+++ b/e2e/test/vitest-5/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "vitest-5",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "description": "Vitest 5 per-test mutation filtering with nested suite names",
+  "scripts": {
+    "test": "stryker run && npm run verify",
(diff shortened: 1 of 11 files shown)


cplieger:
Confirming the separator is the whole defect. I patched only the two joins in the installed `@stryker-mutator/vitest-runner@10.0.0` dist to `' > '` and re-ran Stryker on the same file under vitest 5.0.0: 33 killed / 3 survived / 91.67%, identical to the vitest 4.1.11 baseline down to which three mutants survive (unpatched vitest 5: 0 killed, 0.00 tests per mutant). Both copies have to move together, which this PR does through `testNameSeparator`. Happy to run this branch against more repos if useful.

